PR #3, opened 2015-11-26
Preliminary MSIL-driven Roslyn diagnostics analyzer support (for symbol actions only).

michaelcfanning:
@nguerrera @BillyONeal @lgolding @tikiwanMS @bobschumaker @davidknise


src/BinSkim.Driver/RoslynAnalyzer/RoslynSymbolVisitor.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            Visit(symbol.GlobalNamespace);
+            base.VisitModule(symbol);
+        }
+
+        public override void VisitNamespace(INamespaceSymbol symbol)
+        {
+            foreach (var member in symbol.GetMembers()) { Visit(member); }
+
+            base.VisitNamespace(symbol);
```
nguerrera: Nit: inconsistent spacing between foreach and base call vs. the others.

src/BinSkim.Driver/AnalyzeCommand.cs
```diff
@@ -261,58 +270,114 @@ private HashSet<IBinarySkimmer> CreateSkimmers(IMessageLogger<BinaryAnalyzerCont
...
+        private static void DisposePortableExecutableContextData(BinaryAnalyzerContext context)
+        {
+            context.DisposePortableExecutableData();
+        }
+
+        private void AnalyzeManagedAssembly(string assemblyFilePath, IEnumerable<string> roslynAnalyzerFilePaths, BinaryAnalyzerContext context)
+        {
+            ILDiagnosticsAnalyzer roslynAnalyzer = new ILDiagnosticsAnalyzer();
+
+            foreach (string analyzerFilePath in roslynAnalyzerFilePaths)
```
michaelcfanning: Note that we reload all analyzers for each analysis target, which entails reflecting over the analyzer set again and again. This could be made more efficient.

src/BinSkim.Driver/AnalyzeCommand.cs
```diff
@@ -261,58 +270,114 @@ private HashSet<IBinarySkimmer> CreateSkimmers(IMessageLogger<BinaryAnalyzerCont
...
+                        LogExceptionLoadingRoslynAnalyzer(analyzerFilePath, context, ex);
+                        throw new ExitApplicationException<FailureReason>(DriverResources.UnexpectedApplicationExit, ex)
+                        {
+                            FailureReason = FailureReason.ExceptionCreatingLogFile
+                        };
+                    }
+                );
             }
+
+            roslynAnalyzer.Analyze(assemblyFilePath, diagnostic =>
```
michaelcfanning: Note that this code is not tested for exceptional cases. I will file an issue on this (and on the previous issue related to reloading analyzers over and over)

src/BinSkim.Driver/AnalyzeCommand.cs
nguerrera: Since we use a fresh ILDiagnosticsAnalyzer and only call Analyze once, my earlier comment about bleeding/duplicated registrations is mostly academic.

Maybe just comment that we don't currently support more than one Load/Analyze pair in ILDiagnosticsAnalyzer, but better to either multiple analyzes per instance correctly or change the ILDiagnosticsAnalyzer interface to prevent it.

src/BinSkim.Driver/RoslynAnalyzer/RoslynSymbolVisitor.cs
michaelcfanning: fixed

src/BinSkim.Driver/AnalyzeCommand.cs
```diff
@@ -261,58 +272,122 @@ private HashSet<IBinarySkimmer> CreateSkimmers(IMessageLogger<BinaryAnalyzerCont
...
-            if (context.PE != null) { context.PE.Dispose(); }
-            // TODO: handle pdb but we don't want to fault the PDB in a 2nd time           
+            context.DisposePortableExecutableData();
+        }
+
+        private void AnalyzeManagedAssembly(string assemblyFilePath, IEnumerable<string> roslynAnalyzerFilePaths, BinaryAnalyzerContext context)
+        {
+            ILDiagnosticsAnalyzer roslynAnalyzer = new ILDiagnosticsAnalyzer();
+
+            if (_globalRoslynAnalysisContext == null)
```
michaelcfanning: ok, now we retain a global analysis context object that is initialized once only. we push this onto new ILDiagnosticsAnalyzer instances rather than reloading from analyzer paths. Did I get all this right? I will defer the end compilation action to a future change: because this callback can report diagnostics, I will need to plumb through a mechanism to log through typical channels.

michaelcfanning:
@nguerrera  added compilation end callbacks on both global analysis context and compilation start context. updated brain-dead unit tests to analyze multiple files and to fire messages for relevant compilation end callbacks (to verify we observe the expected set). cleaned up the ILDiagnosticsAnalyzer use pattern. we now have two Create factory methods that provide for a single point of initialization. You can load individual analyzers but only against a context object (rather than an ILDiagnosticsAnalyzer instance).

michaelcfanning:
also, going ahead with the merge on this (but I hope you will take a good look at it) as I'd like to move on to some work updating the output format (which is likely to leak into many places).

src/BinSkim.Driver/RoslynAnalyzer/ILDiagnosticsAnalyzer.cs
```diff
@@ -0,0 +1,110 @@
...
+            // binary. We cannot currently retrieve IL from method bodies.
+            var reference = MetadataReference.CreateFromFile(targetPath);
+            var compilation = CSharpCompilation.Create("_", references: new[] { reference });
+            var target = compilation.GetAssemblyOrModuleSymbol(reference);
+
+            // For each analysis target, we create a compilation start context, which may result
+            // in symbol action registration. We need to capture and throw these registrations 
+            // away for each binary we inspect. 
+            var compilationStartAnalysisContext = new RoslynCompilationStartAnalysisContext(compilation, _options, cancellationToken);
+            _perCompilationSymbolsActions = compilationStartAnalysisContext.SymbolActions;
```
nguerrera: I would pass these down to `Analyze(ISymbol, ...)` and avoid the field that we have to null out. It will get captured with the right lifetime in the `symbol => Analyze(...)` closure.

src/BinSkim.Driver/RoslynAnalyzer/RoslynCompilationStartAnalysisContext.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace Microsoft.CodeAnalysis.IL
+{
+    /// <summary>
+    /// Basic analysis context provided to Roslyn analyzers. We use a singleton instance of this class to 
+    /// capture all symbol action registration for analyzers. These actions will subsequently be invoked
+    /// as we visit the IL of all analysis targets.
+    /// </summary>
+    internal sealed class RoslynCompilationStartAnalysisContext : CompilationStartAnalysisContext
+    {
+        public ActionMap<SymbolAnalysisContext, SymbolKind> SymbolActions { get; set; }
```
nguerrera: Why is this publicly settable?

src/BinSkim.Driver/RoslynAnalyzer/RoslynCompilationStartAnalysisContext.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public Action<CompilationAnalysisContext> CompilationEndActions { get; set; }
```
nguerrera: Ditto. Why public setter? I would expect these to be ReadOnly with RegisterXxx the only way to add new actions.

nguerrera:
LGTM with a few minor comments on changes since the last iteration I reviewed.

PR #34, opened 2016-03-25
Fix failing BinSkim.Driver.FunctionalTests

nguerrera:
They were not being run from BuildAndTest.cmd, but they were running and failing in VS. This fixes the test and enables it on the command line as well.

The baseline had machine, configuration, and runtime-specific strings in it. The fix is to strip them from the baseline on generation and from the actual output during the test run.

Also fix up .gitignore to ignore these files and the .vs/ directory and remove a baseline file that has no actual counterpart.

@michaelcfanning

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,9 +2,11 @@
 *.user 
 *.userosscache 
 *.sln.docstates 
-  
+.vs/
+
 # Build results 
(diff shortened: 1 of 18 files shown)


nguerrera:
Reviewed offline.

PR #51, opened 2016-07-29
Sign securely

michaelcfanning:
@BillyONeal , @tikiwanMS just fyi. 'sign securely' rule now only fires in the event that binaries are signed exclusively with cryptographically weak algorithms. We previously fired false positives for binaries signed with multiple certs (sha1 + sha256, for example) if the first one retrieved was a weak algo.


BillyONeal:
Does vccorlib pass now? (It would be nice to have an example that fails)


Billy3

michaelcfanning:
Previous test was a false positive. :P I opened an issue to create a new test case. Probably will be a generated test case with an untrusted publisher. This will still provoke the right code path (should fire 'untrusted publisher', not 'untrusted algo'

PR #68, opened 2017-01-25
Exlcuding checks for boot binaries

pabrookes:
Some mitigations are not relevant to boot level binaries

### DIFF
--- a/src/BinSkim.Rules/InitializeStackProtection.cs
+++ b/src/BinSkim.Rules/InitializeStackProtection.cs
@@ -6,7 +6,6 @@ using System.Linq;
 using System.Reflection.PortableExecutable;
 
 using Microsoft.CodeAnalysis.BinaryParsers.ProgramDatabase;
-using Microsoft.CodeAnalysis.BinaryParsers.PortableExecutable;
 using Microsoft.CodeAnalysis.Sarif.Driver;
 using Microsoft.CodeAnalysis.IL.Sdk;
 using Microsoft.CodeAnalysis.Sarif;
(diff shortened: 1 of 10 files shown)


src/BinSkim.Rules/InitializeStackProtection.cs
```diff
@@ -57,6 +58,12 @@ protected override IEnumerable<string> FormatIds
 
         public override AnalysisApplicability CanAnalyze(BinaryAnalyzerContext context, out string reasonForNotAnalyzing)
         {
+            PE portableExecutable = context.PE;
+            AnalysisApplicability result = AnalysisApplicability.NotApplicableToSpecifiedTarget;
+
+            reasonForNotAnalyzing = MetadataConditions.ImageIsBootBinary;
```
michaelcfanning: Can this check be moved into the common helper StackProtectionUtilities,CommonCanAnalyze?

src/BinaryParsers/PortableExecutable/PE.cs
```diff
@@ -522,6 +523,35 @@ public bool IsXBox
             }
         }
 
+        public bool IsBoot
+        {
+            get
+            {
+                if(null != _isBoot)
```
michaelcfanning: Style nit, C# doesn't permit unexpected assignment in conditionals, so it isn't typical to order args for a null check in this way. if (_isBoot != null) would be more typical

michaelcfanning:
+1 on this change, please go ahead and merge after verifying whether we can move IsBoot check into stack protection CommonCanAnalyze (and fixing one minor style nit). :)

PR #72, opened 2017-03-16
BinScope Backwards Compatibility

Evmaus-MS:
For rules which had an equivalent check in BinScope, this adds an element to the Result's PropertyBag which denotes the older result name.

This should make it easier for teams to do an in-place replacement of one tool with the other, even if they have automation which takes a dependency on the old naming scheme.


michaelcfanning:
Nice change. Merging.

PR #73, opened 2017-03-22
Update README (formatting and new version on nuget.org).

scovetta:
No content changes, but I fixed the formatting within the README file and bumped the version link to reflect v1.3.9, which is now available on nuget.org.


michaelcfanning:
+1 Thanks for this!

PR #84, opened 2017-05-19
Add response file support. Add additional _Tlgwrite exception for BA2014

michaelcfanning:
@BillyONeal @pabrookes


michaelcfanning:
@Evmaus-MS

PR #104, opened 2017-12-20
Update SARIF dependencies to 1.7.1

michaelcfanning:
@Evmaus-MS


Evmaus-MS:
Should this bump the minor version?

PR #123, opened 2018-01-19
Spectre rule refinements

michaelcfanning:
Fixed unit test. Updated strings per team review. Removed some checks (and opened TODOs) that we should reintroduce more comprehensively across the tool. Removed compiler version check (which needs to be moved to BA2006).

BuildAndTest.Cmd still fails due to a SARIF SDK bug (which we're fixing right now).

@paddymcd-MSFT @Evmaus-MS


src/BinSkim.Rules/EnableSpectreMitigations.cs
```diff
@@ -135,196 +132,197 @@ public override void Analyze(BinaryAnalyzerContext context)
...
+                            version >= omVer)
                         {
                             supportsQspectre = compilerVersionEntry.Value.QSpectre;
-                            supportsd2guardspecload = compilerVersionEntry.Value.d2specguard;
+                            supportsd2guardspecload = compilerVersionEntry.Value.D2SpecGuard;
                         }
                     }
                 }
 
                 if (!supportsd2guardspecload && !supportsQspectre)
```
Evmaus-MS: If we want BA2006 to fire here, we will need to update the version that it requires--I believe the check requires VS2013 right now.  We'd need to bump to 2015.3-(QSpectre patch).

src/BinSkim.Rules/EnableSpectreMitigations.cs
```diff
@@ -107,8 +104,8 @@ public override void Analyze(BinaryAnalyzerContext context)
                 return;
             }
 
-            TruncatedCompilandRecordList badModuleList = new TruncatedCompilandRecordList();
-            TruncatedCompilandRecordList masmModuleList = new TruncatedCompilandRecordList();
+            TruncatedCompilandRecordList mitigationNotEnabledModules = new TruncatedCompilandRecordList();
```
Evmaus-MS: Question--if we're likely to need to expand on the types of states here (e.x. compiled with MASM, etc.), would it make more sense to instead use a structure that maps a state to the appropriate TruncatedCompilandRecordList, rather than continuing to add new lists?

E.x. something like:
Dictionary<State, TruncatedCompilandRecordList>;
Mapping "ExplicitlyDisabled" -> ..., "NotEnabled" -> ..., "MASM" -> ..., etc.

Evmaus-MS:
By and large, looks good.  Had a couple of questions in the comments, but they don't block anything.

src/BinSkim.Rules/EnableSpectreMitigations.cs
```diff
@@ -40,13 +41,9 @@ protected override IEnumerable<string> FormatIds
             {
                 return new string[] {
                     nameof(RuleResources.BA2024_Error),
-                    nameof(RuleResources.BA2024_Error_BuildWithSpectreMitigation_BadCompilerVersion),
-                    nameof(RuleResources.BA2024_Error_BuiildWithSpectreMitigation_UnrecognizedCompiler),
-                    nameof(RuleResources.BA2024_Error_BuildWithSpectreMitigation_SpectreMitigationMissing),
-                    nameof(RuleResources.BA2024_Error_BuildWithSpectreMitigation_SpectreMitigationDisabled),
```
paddymcd-MSFT: As per discussions - I do not feel we should not be disabling checks for MASM, version and other things in this rule just because they are not supported in the rest of the tool's checks.

I know there are feelings about this being too aggressive etc, but given the severity of Spectre we should make this an opt-in check and provide all the information.

If users can specify modules they assert to have checked and been clean we have user performable actions for every error and warning.

michaelcfanning: ok. will address this in a future change.

src/BinSkim.Rules/EnableSpectreMitigations.cs
Evmaus-MS: Agree--however, the design of BinSkim doesn't, at the moment, allow folks to opt in/out of rules.  Per Michael, this was deliberate--it separates the reporting layer (BinSkim) from a separate policy layer (e.x. tools that can consume the results and act on them).

@michaelcfanning I do agree with Paddy that if we aren't going to bump the version in BA2006, we should report this here.

PR #125, opened 2018-01-22
Complete full set of Spectre checks possible with current PDB records

paddymcd-MSFT:
Add support for /Od checking and updated CompilerCommandLine to be able to cope with aliasing and overriding switched to return an actual state based on precedence.

I'm pretty sure that this is the full set of checks that are possible with the existing PDB records.  I will have to work with the compiler team on if we can surface additional information.

This is independent of the refactoring changes @michaelcfanning proposed in his PR.

### DIFF
--- a/src/BinSkim.Rules/EnableSpectreMitigations.cs
+++ b/src/BinSkim.Rules/EnableSpectreMitigations.cs
@@ -44,6 +44,7 @@ namespace Microsoft.CodeAnalysis.IL.Rules
                     nameof(RuleResources.BA2024_Error_BuiildWithSpectreMitigation_UnrecognizedCompiler),
                     nameof(RuleResources.BA2024_Error_BuildWithSpectreMitigation_SpectreMitigationMissing),
                     nameof(RuleResources.BA2024_Error_BuildWithSpectreMitigation_SpectreMitigationDisabled),
+                    nameof(RuleResources.BA2024_Error_BuiltWithSpectreMitigation_OdUsedAndUnsupported),
                     nameof(RuleResources.BA2024_Pass),
                     nameof(RuleResources.BA2024_Pass_WithMASM),
                     nameof(RuleResources.BA2024_Warning_BuildWithSpectreMitigation_MASMDetected),
@@ -202,6 +203,8 @@ namespace Microsoft.CodeAnalysis.IL.Rules
                 // Get the appropriate compiler Version against which to check this compiland
                 bool supportsQspectre = false;
                 bool supportsd2guardspecload = false;
+                bool supportsOd = false;
+                bool onbyDefault = false;
 
                 // check that we are greater than or equal to the first fully supported release: 15.6 first
                 Version omVer = omDetails.CompilerVersion;
@@ -209,6 +212,8 @@ namespace Microsoft.CodeAnalysis.IL.Rules
                 {
                     supportsQspectre = minimumCompilers[VS2017_15_6_PREV4].QSpectre;
                     supportsd2guardspecload = minimumCompilers[VS2017_15_6_PREV4].d2specguard;
+                    supportsOd = minimumCompilers[VS2017_15_6_PREV4].debugCodeMitigated;
+                    onbyDefault = minimumCompilers[VS2017_15_6_PREV4].EnabledByDefault;
                 }
                 else
                 {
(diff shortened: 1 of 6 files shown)


src/BinSkim.Rules/EnableSpectreMitigations.cs
```diff
@@ -238,50 +245,73 @@ public override void Analyze(BinaryAnalyzerContext context)
...
+                    }
 
-                // if either QSpectre or d2guardspecload are enabled AND neither is explicitly disabled then we are protected
-                //      (use of both is confusing so issue an error in this scenario even though they are effectively the same switch)
-                if ((QSpectreState == SwitchState.SwitchEnabled || d2guardspecloadState == SwitchState.SwitchEnabled) && 
-                    (QSpectreState != SwitchState.SwitchDisabled && d2guardspecloadState != SwitchState.SwitchDisabled))
-                {
-                    effectiveState = SwitchState.SwitchEnabled;
-                }
+                    if (QSpectreState == SwitchState.SwitchNotFound && d2guardspecloadState == SwitchState.SwitchNotFound)
```
Evmaus-MS: Nit/Style:  Either have both continues on the inside of the blocks in this if/else, or have one after the if/else is done.

src/BinSkim.Rules/EnableSpectreMitigations.cs
```diff
@@ -223,6 +228,8 @@ public override void Analyze(BinaryAnalyzerContext context)
                         {
```
Evmaus-MS: Correct me if I'm wrong--this block assumes the minimum compiler version list is always sorted in ascending order, correct?  We should document that here or there.

src/BinSkim.Rules/EnableSpectreMitigations.cs
```diff
@@ -238,50 +245,73 @@ public override void Analyze(BinaryAnalyzerContext context)
...
+                if (supportsOd == false)
                 {
-                    // built with tools that support the Spectre mitigations but these have not been enabled.
-                    badModuleList.Add(
-                        om.CreateCompilandRecordWithSuffix(
-                            string.Format(CultureInfo.InvariantCulture,
-                            RuleResources.BA2024_Error_BuildWithSpectreMitigation_SpectreMitigationMissing)));
+                    string[] OptimizerSwitches = { "/Od", "/O1", "/O2", "/Ox", "/Og" };
+
+                    bool debugEnabled = false;
```
Evmaus-MS: Per the documentation (https://docs.microsoft.com/en-us/cpp/build/reference/od-disable-debug ), /Od is enabled by default--I haven't touched the PDB command line stuff, so I have a clarifying question here:  will the flag show up in the command line parameters in the PDB if it isn't passed at all?

paddymcd-MSFT: No - it will not show up in the command line.

The logic encapsulated by the new functionality in the CompilerCommandLine.GetSwitchStateFactoringOverrides (really not sure if this name conforms with the codebase conventions) allows you to pass a default state which will be set if none of the overriding switches are present.  

So if you are aware of what the default state is you can pass that in, or check for SwitchState.SwitchNotFound

src/BinSkim.Rules/EnableSpectreMitigations.cs
paddymcd-MSFT: Yes, it makes that assumption, which at present that is correct.  I gave up on a sorted set earlier as I couldn't get it to work and wanted to concentrate on the actual check.  In order to make this robust we will need to either sort the list appropriately such that this assumption is correct, or change the logic to cope with that not being the case.

src/BinSkim.Rules/EnableSpectreMitigations.cs
paddymcd-MSFT: good catch - will correct

src/BinSkim.Rules/EnableSpectreMitigations.cs
paddymcd-MSFT: Added a quick TODO to the code such that I can deal with this next iteration

Evmaus-MS: Makes sense.  Can we note this in a comment on the structure itself (i.e. "This is assumed to be sorted elsewhere in this check"?), so that any future changes can take that into account?

src/BinSkim.Rules/EnableSpectreMitigations.cs
Evmaus-MS: Ahh, OK, so the SwitchState.Enabled parameter in the following takes care of that.  Makes sense, thanks!

src/BinSkim.Rules/EnableSpectreMitigations.cs
paddymcd-MSFT: Makes sense - I'll coordinate with @michaelcfanning to make sure this doesn't get missed in the merging of his sanitization with the full experimental rule

PR #126, opened 2018-01-23
Platform specific checks

Evmaus-MS:
This has one remaining TODO for usability (waiting on https://github.com/Microsoft/sarif-sdk/pull/756, then changing the supported platform implementation slightly), but the overall work is done to guard all of the platform specific functionality on non-Windows platforms.

Tests were updated for non-Windows platforms as well, so it should be possible to build/test on Linux now and have all of the tests pass.


src/BinSkim.Driver.FunctionalTests/BaselineTests.cs
```diff
@@ -26,14 +27,14 @@ public BuiltInRuleFunctionalTests(ITestOutputHelper output)
             _testOutputHelper = output;
         }
 
-        private static string TestDirectory = GetTestDirectory(@"BinSkim.Driver.FunctionalTests\BaselineTestsData");
+        private static string TestDirectory = GetTestDirectory(@"BinSkim.Driver.FunctionalTests"+Path.DirectorySeparatorChar+"BaselineTestsData");
```
michaelcfanning: NIT, are there spacing issues in this line?

src/BinSkim.Driver.FunctionalTests/BaselineTests.cs
```diff
@@ -85,8 +86,13 @@ private void RunRules(StringBuilder sb, string inputFileName)
         {
             string fileName = Path.GetFileName(inputFileName);
             string actualDirectory = Path.Combine(Path.GetDirectoryName(inputFileName), "Actual");
-            string expectedDirectory = Path.Combine(Path.GetDirectoryName(inputFileName), "Expected");
-
+            string expectedDirectory;
+            if(PlatformSpecificHelpers.RunningOnWindows()){
```
michaelcfanning: More spacing issues. I will stop making this comment. We should consider running CodeFormatter against our code to get it into a uniform condition and scripting execution of this tool so that it's easy to maintain consistency.

src/BinSkim.Driver.FunctionalTests/BaselineTestsData/NonWindowsExpected/Corrupted_Native_x86_VS2013_Default.exe.sarif
```diff
@@ -0,0 +1 @@
+{"$schema":"http://json.schemastore.org/sarif-1.0.0","version":"1.0.0","runs":[{"id":"22691344-89fa-4fdf-b03c-db1a387980c5","tool":{"name":"BinSkim","fullName":"BinSkim 1.5.0.0","version":"1.5.0.0","semanticVersion":"1.5.0","sarifLoggerVersion":"1.7.1.0","language":"en-US","properties":{"Comments":"A security and correctness analyzer for portable executable and MSIL formats.","CompanyName":"Microsoft","ProductName":"BinSkim Portable Executable Analyzer"}},"results":[],"configurationNotifications":[{"id":"ERR997.ExceptionLoadingAnalysisTarget","physicalLocation":{"uri":"file://Z:/src/BinSkim.Driver.FunctionalTests/BaselineTestsData/Corrupted_Native_x86_VS2013_Default.exe"},"message":"Could not load analysis target 'Corrupted_Native_x86_VS2013_Default.exe'. (Image is either too small or contains an invalid byte offset or count.)","level":"error"}],"invocation":{"startTime":"2018-01-17T00:34:44.418Z","endTime":"2018-01-17T00:34:44.647Z"}}]}
```
michaelcfanning: All tests files will need to be refreshed once we restore default pretty printing. We can ignore issues in this change now, if you are sure you haven't altered analysis results. You can use a private build of SARIF SDK with the pretty-print option restored if it's helpful in the short-term

src/BinSkim.Rules.FunctionalTests/ILDiagnosticsAnalyzerTests.cs
```diff
@@ -78,6 +78,7 @@ public void RunSymbolNameAndRoslynCallbackReportingsAnalyzers()
                 "Symbol encountered in MSIL 'global::<Module>'",
                 "Symbol encountered in MSIL 'global::Microsoft.CodeAnalysis.BinaryParsers.BinaryParsersResources'",
                 "Symbol encountered in MSIL 'global::Microsoft.CodeAnalysis.BinaryParsers.Packer'",
+                "Symbol encountered in MSIL 'global::Microsoft.CodeAnalysis.BinaryParsers.PlatformSpecificHelpers'",
```
michaelcfanning: this test file will be deleted soon, you might just want to delete it in your branch so that it doesn't continue to produce costs for you.

src/BinSkim.Rules/DoNotShipVulnerableBinaries.cs
```diff
@@ -80,13 +80,22 @@ private static StringToVersionMap BuildDefaultVulnerableBinariesMap()
 
         public override void Analyze(BinaryAnalyzerContext context)
         {
+            // Throw a platform unsupported exception--FileVersionInfo does not behave "correctly" on Linux.
+            BinaryParsers.PlatformSpecificHelpers.ThrowIfNotOnWindows();
```
michaelcfanning: This is a good and clean change. Moving forward, I wonder if it'd be helpful to annotate rules with some metadata that indicates a dependency on PDBs, allowing us to move this handling somewhere in a more central place. Something to keep in mind.

src/BinSkim.Rules/DoNotShipVulnerableBinaries.cs
```diff
@@ -80,13 +80,22 @@ private static StringToVersionMap BuildDefaultVulnerableBinariesMap()
...
+            
             string fileName = Path.GetFileName(context.PE.FileName);
 
             Version minimumVersion;
             if (context.Policy.GetProperty(VulnerableBinaries).TryGetValue(fileName, out minimumVersion))
             {
                 FileVersionInfo fvi = FileVersionInfo.GetVersionInfo(Path.GetFullPath(context.PE.FileName));
                 string rawVersion = fvi.FileVersion;
+                // Version is sometimes null on non-Windows platforms.
```
michaelcfanning: You can replace this code by making the previous line look like so:

string rawVersion = fvi.FileVersion ?? String.Empty;

src/BinSkim.Rules/NativeInterop.cs
```diff
@@ -218,10 +218,15 @@ internal struct CRYPTOAPI_BLOB
         internal const string szOID_CERT_STRONG_KEY_OS_1 = "1.3.6.1.4.1.311.72.2.1";
 
         [DllImport("winterop.dll", EntryPoint = "HashPublicKeyInfo", CharSet = CharSet.Unicode, ExactSpelling = true, PreserveSig = false)]
-        internal static extern void HashPublicKeyInfo(IntPtr certContext, byte[] publicKeyInfoHashed, ref uint sizePublicKeyInfoHashed);
+        private static extern void HashPublicKeyInfo(IntPtr certContext, byte[] publicKeyInfoHashed, ref uint sizePublicKeyInfoHashed);
+
+        internal static void HashPublicKeyInfoWrapper(IntPtr certContext, byte[] publicKeyInfoHashed, ref uint sizePublicKeyInfoHashed)
```
michaelcfanning: Do these wrappers exist with this implementation because the interop helpers are present in code but currently unused? It's hard for me to tell what's going on here using the GitHub diff UI. :) I will look at this more closely after enlisting in your branch.

src/BinaryParsers/PlatformSpecificHelpers.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            return RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+        }
+
+        public static void ThrowIfNotOnWindows()
+        {
+            if(!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                throw new PlatformNotSupportedException(
+                    string.Format(BinaryParsersResources.PlatformUnsupportedFormat, RuntimeInformation.OSDescription, OSPlatform.Windows));
```
michaelcfanning: NIT, note that our source is currently inconsistent in whether we use the C# aliases or formal .NET primitive names when invoking statics on the types. I think CodeFormatter can help clean this up. StyleCop may be able to as well, another option.

src/BinaryParsers/ProgramDatabase/Pdb.cs
```diff
@@ -92,34 +92,44 @@ private void Init(string peOrPdbPath, string symbolPath)
         {
             try
             {
-                DiaSource diaSource = new DiaSource();
-                Environment.SetEnvironmentVariable("_NT_SYMBOL_PATH", "");
+                PlatformSpecificHelpers.ThrowIfNotOnWindows();
+                WindowsNativeLoadPdbUsingDia(peOrPdbPath, symbolPath);
+            }
+            catch (PlatformNotSupportedException ex)
```
michaelcfanning: I wonder whether we wouldn't want to let this specific exception through as it is. It is one thing for PDB loading to be entirely unsupported. Generally, a consumer should simply turn of broadstroke functionality as a result. A PDB load issue should indicate that we had a chance to load the PDB but it failed for some reason (e.g., the PDB could not be located). This distinction is helpful (you have a chance to succeed but it went wrong somehow vs. you have no chance at all). Interested in your thoughts.

src/BinSkim.Rules/NativeInterop.cs
Evmaus-MS: We don't use these at the moment, but I think I must have just missed converting the wrappers.  Updated.

src/BinSkim.Rules.FunctionalTests/ILDiagnosticsAnalyzerTests.cs
Evmaus-MS: I've added an ignore tag on my other branch.  :)

src/BinSkim.Driver.FunctionalTests/BaselineTestsData/NonWindowsExpected/Corrupted_Native_x86_VS2013_Default.exe.sarif
Evmaus-MS: I think this is probably fine for now.  These are all of the non-Windows ones--this change doesn't modify the expected results on Windows (and that test still works correctly), I'll need to do more extensive verification that the results are all correct/expected after the next set of changes.

src/BinSkim.Driver.FunctionalTests/BaselineTests.cs
Evmaus-MS: Agree that we should find a tool for this.  I went through the changes and tried to correct them whenever I found spacing issues.

PR #132, opened 2018-01-25
Enable reporting of MASM modules

michaelcfanning:
@Evmaus-MS @paddymcd-MSFT 

There is a deserialization issue with the new option, will get this fixed in the next commit. Wanted to show the basics of this change. We need a test binary that links a MASM modules

### DIFF
--- a/src/BinSkim.Rules/EnableSpectreMitigations.cs
+++ b/src/BinSkim.Rules/EnableSpectreMitigations.cs
@@ -52,8 +52,9 @@ namespace Microsoft.CodeAnalysis.IL.Rules
         {
             return new List<IOption>
             {
-                MinimumToolVersions,
-                AllowedLibraries
+                Reporting,
+                AllowedLibraries,
+                //MinimumToolVersions
             }.ToImmutableArray();
         }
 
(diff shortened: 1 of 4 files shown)


src/BinSkim.Rules/EnableSpectreMitigations.cs
```diff
@@ -402,13 +416,21 @@ private static StringToVersionMap BuildAllowedLibraries()
         }
     }
 
+    [Flags]
+    internal enum ReportingOptions
+    {
+        None = 0x0,
+        Default = 0x1,
+        MasmModules = 0x2
```
paddymcd-MSFT: 1) Default should be a selection of what the actual meaningful flags are, not a flag on its own (in this instance None or MasmModules)
2) The Default for this rule should be MasmModules

paddymcd-MSFT:
The crt's from Microsoft (msvcrt.lib etc) include asm object files for things like eh handlers, so you actually have to put effort in to get a binary that does **not** include asm objects

src/BinSkim.Rules/EnableSpectreMitigations.cs
Evmaus-MS: Agree with (1).

Disagree with (2).  As we've discussed, the default behavior should be the one that applies to most users, and should be immediately actionable.  Until we have a complete story around how to mark MASM modules as validated, it should not be the default.

src/BinSkim.Rules/RuleResources.resx
```diff
@@ -364,6 +364,10 @@ Modules triggering this check were:
   <data name="BA2024_Error" xml:space="preserve">
     <value>'{0}' was compiled with one or more modules that do not properly enable code generation mitigations for speculative execution side-channel attack (Spectre) vulnerabilities. Spectre attacks can compromise hardware-based isolation, allowing non-privileged users to retrieve potentially sensitive data from the CPU cache. To resolve the issue, provide the /Qspectre switch on the compiler command-line (or /d2guardspecload in cases where your compiler supports this switch and it is not possible to update to a toolset that supports /Qspectre). The following modules are out of policy:
 {1}</value>
+  </data>
+  <data name="BA2024_Error_MasmModulesDetected" xml:space="preserve">
+    <value>The following MASM modules were detected. The MASM compiler does not currently mitigate against speculative execution attacks:
```
Evmaus-MS: Do we have any guidance we can link to or include on what folks should *do* when MASM modules are detected?

src/BinSkim.Rules/EnableSpectreMitigations.cs
michaelcfanning: yes, as we've discussed, the full reporting should be opt-in. can discuss further offline.

src/BinSkim.Rules/RuleResources.resx
michaelcfanning: this is the entire point around making this reporting opt-in. the only action possible is to suppress (which is configuration work) or to drop the MASM use (which entails additional costs which could be high, or it may not be feasible to replace).

paddymcd-MSFT: The action on asm files is that once you have vetted them or trust that the provider has vetted them that you add them to a blessed list that should be fed to the tool - am currently investigating if we have access to the MD5 hash for the file in the current PDB interface to add that as a failsafe should the asm file change

michaelcfanning: it is good to have capability. we won't be able to service Binskim currently if it introduces these costs by default for teams that are onboarding. open to discussion moving forward as we collect data and discuss policy changes that might be driven by this technical problem.

PR #136, opened 2018-01-28
Add check for /Od disabling mitigations.

paddymcd-MSFT:
Add check for /Od disabling mitigations.
Add test case surce and generation script for test cases based on compiler version (no binaries as yet - need to enumerate through appropriate compiler versions).

### DIFF
--- a/src/BinSkim.Rules/RuleResources.resx
+++ b/src/BinSkim.Rules/RuleResources.resx
@@ -369,6 +369,10 @@ Modules triggering this check were:
     <value>The following MASM modules were detected. The MASM compiler does not currently mitigate against speculative execution attacks:
 {0}</value>
   </data>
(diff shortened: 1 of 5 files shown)


src/BinSkim.Rules/RuleResources.resx
```diff
@@ -367,6 +367,10 @@ Modules triggering this check were:
   </data>
   <data name="BA2024_Error_MasmModulesDetected" xml:space="preserve">
     <value>The following MASM modules were detected. The MASM compiler does not currently mitigate against speculative execution attacks:
+{0}</value>
+  </data>
+  <data name="BA2024_Error_SpectreMitigationDisabledByOd" xml:space="preserve">
+    <value>The following modules were compiled /Od which disables /Qspectre:
```
michaelcfanning: This string should explicitly note that /Of disables optimizations, for tier 1 support/other folks who don't understand the implications of the switch

PR #148, opened 2018-02-03
ELF Compiler Detection, GCC Specific Check

Evmaus-MS:
This adds a GCC specific ELF check (that -DFortifySource=2 should be used with -O2 to add runtime checking for dangerous functions).


Evmaus-MS:
Merging this into the working branch; it can just be reviewed with the rest of the code.

PR #149, opened 2018-02-05
.NET Core Port, Windows-Specific Rule handling, and ELF Rules

Evmaus-MS:
This PR encompasses a couple of changes (some of which were reviewed separately, but they should be re-reviewed as a whole):
-.NET Core port of BinSkim
-Adds Windows Specific rule handling (for rules that require interop with PDB parsing/Windows crypto APIs)
-Refactors to allow loading of multiple binary types
-Adds ELF parsing, ELF Compiler Detection, and ELF rules.


src/BinSkim.Driver.FunctionalTests/BaselineTestsData/NonWindowsExpected/gcc.shared_library.so.sarif
```diff
@@ -0,0 +1,610 @@
...
+        {
+          "id": "WRN998.NotSupportedPlatform",
+          "physicalLocation": {},
+          "message": "Rule 'BuildWithSecureTools' was disabled as it cannot run on the current platform 'Windows'.  It can only run on 'Linux'.",
+          "level": "note"
+        },
+        {
+          "id": "WRN998.NotSupportedPlatform",
+          "physicalLocation": {},
+          "message": "Rule 'DoNotDisableStackProtectionForFunctions' was disabled as it cannot run on the current platform 'Windows'.  It can only run on 'Linux'.",
```
Evmaus-MS: OS comments are backwards due to https://github.com/Microsoft/sarif-sdk/pull/767

PR #150, opened 2018-02-07
New Spectre Config Model

Evmaus-MS:
This replaces the existing configuration model with a slightly more flexible one, largely as we've discussed elsewhere, and provides new unit tests for the various helper methods included in the change.

The XML config for MitigatedCompilers looks like this with this change:

`<Properties Key="MitigatedCompilers" Type="PropertiesDictionary">
      <Properties Key="Arm" Type="PropertiesDictionary" />
      <Properties Key="X86" Type="PropertiesDictionary">
        <Property Key="19.0.*.* - 19.0.*.*" Value="QSpectreAvailable" />
        <Property Key="19.10.*.* - 19.10.*.*" Value="QSpectreAvailable" />
        <Property Key="19.12.*.* - 19.12.*.*" Value="QSpectreAvailable, D2GuardSpecLoadAvailable" />
        <Property Key="19.12.25830.2-19.12.*.*" Value="D2GuardSpecLoadAvailable" />
        <Property Key="19.13.*.* - 19.13.*.*" Value="QSpectreAvailable, D2GuardSpecLoadAvailable" />
        <Property Key="19.13.26029.0 - 19.13.26029.*" Value="D2GuardSpecLoadAvailable" />
      </Properties>`

Where * is replaced with int.MaxValue.

The check is also changed slightly to handle the new data structure & lookup (but it shouldn't be functionally different).

### DIFF
--- a/src/BinSkim.Rules/BuildWithSecureTools.cs
+++ b/src/BinSkim.Rules/BuildWithSecureTools.cs
@@ -171,13 +171,10 @@ namespace Microsoft.CodeAnalysis.IL.Rules
                         // Now we'll retrieve relevant compiler mitigation details to
                         // ensure this object module's build and revision meet
                         // expectations.
-                        MitigatedVersion mitigationData;
+                        CompilerMitigations newMitigationData = 
+                            EnableSpectreMitigations.GetAvailableMitigations(context, machineType, actualVersion);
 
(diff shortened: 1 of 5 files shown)


michaelcfanning:
Well done. I checked out your branch and made a couple of small renaming changes. Now merging.

PR #155, opened 2018-02-19
Dev/paddymcd msft/is stripped

michaelcfanning:
@paddymcd-MSFT @Evmaus-MS 

This changes builds on Paddy's binary parsers update to identify stripped PDBs. Changes include:

1) Updated all docs to reflect that BA2024 reads PDBs
2) Generated test case with stripped PDB and added to driver functionals
3) Created a PDB reading skimmer base class that captures all PDB validation
4) added binary parser unit tests that validates new properties and exception capturing behaviors


src/BinaryParsers/BinaryParsers.csproj
```diff
@@ -2,12 +2,12 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory).., build.netcore.props))\build.netcore.props" />
   <PropertyGroup>
-    <RootNamespace>Microsoft.CodeAnalysis.BinaryParsers</RootNamespace> 
+    <RootNamespace>Microsoft.CodeAnalysis.BinaryParsers</RootNamespace>
+    <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
```
Evmaus-MS: Is there a reason we're now building both of these here?  This is a library, it should be consumable as .net standard from both .net 461 and .net core.

(Also--we have macros for these if we need to build both for some reason.)

Issue #153, opened 2018-02-14
BA2006 with AdvancedMitigationsEnforced Improvements


Evmaus-MS:
BA2006 with AdvancedMitigationsEnforced=Spectre currently only checks for & returns the minimum compiler supporting D2GuardSpecLoad/QSpectre.

This is the incorrect behavior--we should check if the flag(s) are not supported, and if they are not, we should return the minimal *upgrade* required to get the support, rather than the minimal compiler that supports it (i.e. if someone is on a version of VS2017 that does not have support, we should not suggest a downgrade to VS2015 as a fix once the servicing release is available for VS2015 update 3).

Evmaus-MS:
Working branch is https://github.com/Microsoft/binskim/tree/dev/evmaus-ms/BA2006_SpectreReportingFix

Still need to ensure this covers all cases and write tests.

Evmaus-MS:
This was closed a while ago.

PR #156, opened 2018-02-20
Dev/evmaus ms/ba2006 spectre reporting fix

Evmaus-MS:
Fix for https://github.com/Microsoft/binskim/issues/153

The issue this is addressing is that if we're flagging for not using a compiler version that supports Spectre mitigations, the "minimum compiler version" they should upgrade to is the smallest possible upgrade from their current version--and should never be a downgrade. The previous implementation didn't address either of these.


michaelcfanning:
@paddymcd-MSFT would appreciate your feedback on this

src/BinSkim.Rules/PERules/BA2024.EnableSpectreMitigations.cs
```diff
@@ -94,6 +94,15 @@ public IEnumerable<IOption> GetOptions()
 
         private static Dictionary<MachineFamily, CompilerVersionToMitigation[]> _compilerData = null;
 
+        // Used only for testing.
+        internal static Dictionary<MachineFamily, CompilerVersionToMitigation[]> _compilerDataCache
```
michaelcfanning: >_compilerDataCache [](start = 81, length = 18)

why is this a property and not a field? did you do it to make this debuggable on property set?

suggest just marking _compilerData internal and be done with it :)

PR #157, opened 2018-02-21
Fix for a bug in high entropy va check, with new test cases.

Evmaus-MS:
Fixes a FN case where we compile as AnyCPU and do not have 'prefers 32 bit' checked, when HighEntropyVA is disabled.

Per experimentation, a binary compiled with this configuration will be loaded below the 4GB boundary, but in a 64 bit address space.


src/BinSkim.Rules/PERules/BA2015.EnableHighEntropyVirtualAddresses.cs
```diff
@@ -60,8 +60,16 @@ public override AnalysisApplicability CanAnalyzePE(PEBinary target, Sarif.Proper
             if (portableExecutable.IsKernelMode) { return result; }
 
             reasonForNotAnalyzing = MetadataConditions.ImageIsNot64BitBinary;
```
michaelcfanning: >ImageIsNot64BitBinary [](start = 55, length = 21)

New condition?

'ImageDoesNotRunAs64BitProcess'

src/BinSkim.Rules/PERules/BA2015.EnableHighEntropyVirtualAddresses.cs
```diff
@@ -60,8 +60,16 @@ public override AnalysisApplicability CanAnalyzePE(PEBinary target, Sarif.Proper
...
-            if (portableExecutable.PEHeaders.PEHeader.Magic != PEMagic.PE32Plus) { return result; }
-
+            if (portableExecutable.PEHeaders.PEHeader.Magic != PEMagic.PE32Plus
```
michaelcfanning: >portableExecutable.PEHeaders.PEHeader.Magic != PEMagic.PE32Plus [](start = 16, length = 63)

complex condition, let's comment what this logic is accomplishing, flagging an AnyCpu binary that is configured such that it might run as 64-bit

src/ReleaseHistory.md
```diff
@@ -59,5 +59,8 @@
 * Sign all BinSkim binaries
 
 ## **v1.5.0** [NuGet Package](https://www.nuget.org/packages/Microsoft.CodeAnalysis.BinSkim/1.5.0
+* Cross platform (Windows/Linux) support.
 * Provide preliminary BA2024.EnableSpectreMitigations analysis
 * Identify and fire configuration errors when located PDBs are stripped
+* New rules for ELF Binaries
+* False negative removed:  Correctly flags an AnyCPU binary with HighEntropyVA and Prefer32Bit disabled
```
michaelcfanning: >False negative removed [](start = 2, length = 22)

Cite rule id and friendly name here. BA2015.EnableHighEntropyVA.

Also we should consider a convention that describes new analysis.

michaelcfanning:
>       "tool": {

1. We need a multitool sort command and we need to apply it to our tests.
2. Suggest undoing changes except for test file adds.

---
Refers to: src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/clang.default_compilation.sarif:6 in 5f4a9be. [](commit_id = 5f4a9bea17cf1e03cc3a43ce3b2615d249522bc6, deletion_comment = False)

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/Managed_AnyCPU_VS2017_NoPrefer32Bit.exe.sarif
```diff
@@ -0,0 +1,887 @@
...
+          },
+          "message": "Analyzing 'Managed_AnyCPU_VS2017_NoPrefer32Bit.exe'...",
+          "level": "note",
+        }
+      ],
+      "rules": {
+        "BA3001": {
+          "id": "BA3001",
+          "name": "EnablePIEOnExecutables",
+          "shortDescription": "A Position Independent Executable (PIE) relocates all of its sections at load time, including the code section, if ASLR is enabled in the Linux kernel (instead of just the stack/heap).",
```
michaelcfanning: >PIE [](start = 66, length = 3)

Can we change casing of this to Pie? And ELF to Elf in any rule ids/names before we ship

src/BinSkim.Rules/PERules/BA2015.EnableHighEntropyVirtualAddresses.cs
Evmaus-MS: #resolved

Evmaus-MS:
Addressed feedback.  Not rolling back results files as renaming some of the checks changed the results.

PR #172, opened 2018-07-18
Fixing build to copy correct dependencies

Evmaus-MS:
This fixes the build so that the native dependencies are properly copied.

### DIFF
--- a/src/BinSkim.Driver/BinSkim.Driver.csproj
+++ b/src/BinSkim.Driver/BinSkim.Driver.csproj
@@ -64,8 +64,8 @@
 
   <ItemGroup>
     <!-- May need to be changed to x86 on a 32 bit machine. -->
-    <MsDiaLibs Include="..\..\refs\x64\*" Condition="'$(Configuration)' == 'Debug' And '$(TargetFramework)'=='$(NetRuntimeVersion)' And '$(RuntimeIdentifier)'=='win7-x86'" />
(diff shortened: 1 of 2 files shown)


src/BinSkim.Driver/BinSkim.Driver.csproj
```diff
@@ -64,8 +64,8 @@
 
   <ItemGroup>
     <!-- May need to be changed to x86 on a 32 bit machine. -->
-    <MsDiaLibs Include="..\..\refs\x64\*" Condition="'$(Configuration)' == 'Debug' And '$(TargetFramework)'=='$(NetRuntimeVersion)' And '$(RuntimeIdentifier)'=='win7-x86'" />
```
michaelcfanning: >'$(Configuration)' == 'Debug' And '$(T [](start = 53, length = 38)

ouch! :)

src/BinSkim.Driver/BinSkim.cs
```diff
@@ -10,6 +10,7 @@
 using System.Linq;
 using System.Runtime.InteropServices;
 using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.Driver.Sdk;
```
michaelcfanning: >CodeAnalysis.Sarif.Driver [](start = 16, length = 25)

is this change actually required? just curious

Evmaus-MS: Yeah, the namespace for EntryPointUtilities in the driver was changed at some point.

PR #174, opened 2018-07-30
COM Fixes

Evmaus-MS:
This fixes the issues we were having with unregistered COM stuff and .net core by manually creating the COM objects.

This means we can ship entirely in .net core, so I've removed the .net framework from the build--this allows us to simplify our shipping story (and the number of things we have to test for compatibility/etc.) dramatically.


Evmaus-MS:
Had a couple of iterations of build tweaking to get it building everywhere, but they should be done now--sorry for the slight churn on a PR.

BuildAndTest.cmd
```diff
@@ -37,11 +37,11 @@ echo         public const string Version = AssemblyVersion + Prerelease;
 echo     }                                                                           >> %VERSION_CONSTANTS%
 echo  }                                                                              >> %VERSION_CONSTANTS%
 
-%~dp0.nuget\NuGet.exe restore %~dp0src\BinSkim.sln -ConfigFile "%NuGetConfigFile%" -OutputDirectory "%NuGetPackageDir%"
```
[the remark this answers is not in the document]
Evmaus-MS: Not yet, we still use it for packing via a .nuspec.  I haven't looked into what's required to move us over to packing via 'dotnet pack' (which targets a .csproj file)--our packing needs are a little nonstandard (we're using it for full tool distribution instead of library distribution), so that may be slightly more challenging to port to the .csproj packing model.

src/BinaryParsers/PEBinary/ProgramDatabase/MSDiaComWrapper.cs
```diff
@@ -0,0 +1,52 @@
...
+    public interface IClassFactory
+    {
+        [PreserveSig]
+        int CreateInstance(IntPtr pUnkOuter, ref Guid riid, out IntPtr ppvObject);
+        int LockServer(int fLock);
+    }
+
+    public class MSDiaComWrapper
+    {
+        [DllImport("msdia140.dll", CharSet = System.Runtime.InteropServices.CharSet.Ansi, ExactSpelling = true, BestFitMapping = false)]
```
michaelcfanning: >ExactSpelling = true [](start = 90, length = 20)

not strictly required, is this? also not a critical problem...

src/BinaryParsers/PEBinary/ProgramDatabase/MSDiaComWrapper.cs
```diff
@@ -0,0 +1,52 @@
...
+            IntPtr pClassFactory = IntPtr.Zero;
+            int hr = DllGetClassObject(clsidOfServer, new Guid("00000001-0000-0000-C000-000000000046"), out pClassFactory);
+            IClassFactory classFactory = Marshal.GetObjectForIUnknown(pClassFactory) as IClassFactory;
+            classFactory.CreateInstance(IntPtr.Zero, ref riid, out pvObject);
+            Marshal.Release(pClassFactory);
+            Marshal.ReleaseComObject(classFactory);
+        }
+
+
+        private const string IDiaDataSourceRIId = "79F1BB5F-B66E-48E5-B6A9-1545C323CA3D";
```
michaelcfanning: >IDiaDataSourceRIId [](start = 29, length = 18)

casing nit! IDiaDataSourceRiid would be more acceptable, a 'riid' is a sort of discrete native originating term

src/BinaryParsers/PEBinary/ProgramDatabase/MSDiaComWrapper.cs
```diff
@@ -0,0 +1,52 @@
...
+        private const string DiaSourceCLSID = "E6756135-1E65-4D17-8576-610761398C3C";
```
michaelcfanning: >DiaSourceCLSID [](start = 29, length = 14)

More NITS! s/be DiaSourceClsid (casing)

src/BinaryParsers/PEBinary/ProgramDatabase/MSDiaComWrapper.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public static IDiaDataSource GetDiaSource()
+        {
+            object objectForIUnknown = null;
+            IDiaDataSource diaSourceInstance = null;
```
michaelcfanning: >= null [](start = 45, length = 6)

unnecessary init, this is a dead store

src/BinaryParsers/PEBinary/ProgramDatabase/MSDiaComWrapper.cs
```diff
@@ -0,0 +1,52 @@
...
+using Dia2Lib;
+
+namespace Microsoft.CodeAnalysis.BinaryParsers.ProgramDatabase
+{
+    [ComImport, InterfaceType(ComInterfaceType.InterfaceIsIUnknown), Guid("00000001-0000-0000-C000-000000000046")]
+    public interface IClassFactory
+    {
+        [PreserveSig]
+        int CreateInstance(IntPtr pUnkOuter, ref Guid riid, out IntPtr ppvObject);
+        int LockServer(int fLock);
```
michaelcfanning: >LockServer [](start = 12, length = 10)

are we using this? If not, delete for now

michaelcfanning:
suggestion: add a simple instantiation unit test? maybe not super necessary, since lots of other code fails if this mechanism breaks. so, feel free to bounce if you have to move on to other things

PR #183, opened 2018-08-04
Make package conform to new NuGet ship requirements.

michaelcfanning:
@Evmaus-MS


michaelcfanning:
FYI, have updated the BinSkim package to meet NuGet.org ship requirements.

PR #193, opened 2018-08-10
Sarif v2

michaelcfanning:
@Evmaus-MS @pabrookes


michaelcfanning:
Note that I will add v1 format compatibility in a follow on change. Also, we need a couple of SDK fixes (such as removing use of ruleMessageStringId'.

src/build.netcore.props
```diff
@@ -5,6 +5,6 @@
   <PropertyGroup>
     <NetCoreVersion>netcoreapp2.0</NetCoreVersion>
     <NetStandardVersion>netstandard2.0</NetStandardVersion>
-	  <SarifSDKVersion>1.7.5</SarifSDKVersion>
+	  <SarifSDKVersion>2.0.0-csd.1</SarifSDKVersion>
```
Evmaus-MS: Nit: indentation (was already here)

Evmaus-MS:
>             ILDiagnosticsAnalyzer roslynAnalyzer = ILDiagnosticsAnalyzer.Create(_globalRoslynAnalysisContext);

Do we want to just delete this?  This is the unused IL analysis code.  We can always tag the 1.5.0 release on master as the last release with it if we need to keep track of it.

---
Refers to: src/BinSkim.Driver/AnalyzeCommand.cs:110 in 1ddfd0b. [](commit_id = 1ddfd0bedaf2973ce76462a680aeff2950ae7404, deletion_comment = False)

Evmaus-MS:
>     internal static class RoslynExtensionMethods

Same comment here about delete vs. keep.

---
Refers to: src/BinSkim.Driver/RoslynAnalyzer/RoslynExtensionMethods.cs:10 in 1ddfd0b. [](commit_id = 1ddfd0bedaf2973ce76462a680aeff2950ae7404, deletion_comment = False)

PR #245, opened 2019-09-11
Results caching

michaelcfanning:
Update to SARIF 2.1.6, which provides results caching by file hash. Passing --hashes on BinSkim command-line will now perform a single analysis of any file (by hash) that it encounters during scan. This greatly reduces analysis time scanning large build drops (where dependencies are copied repeatedly to different output dirs.).

In a future change, we will redesign the tool to crawl a PDB a single time (and call back to all PDB walking rules) rather than having each rule crawl the data.

This change also includes another improvement: when passing the symbols server path (via --sympath) we would previously ignore a PDB if it were alongside the binary (i.e., you could probe local paths or the symbol server exclusively). Now we test for a local copy first, then fall back to the symbol server API, if configured.


michaelcfanning:
@rtaket

PR #258, opened 2020-02-01
Formatting clean-up changes after last work.

michaelcfanning:
@chinarosesz 

Little bit of tail from your recent and fantastic reformatting infrastructure work.

### DIFF
--- a/src/.editorconfig
+++ b/src/.editorconfig
@@ -41,18 +41,18 @@ dotnet_style_parentheses_in_relational_binary_operators = always_for_clarity:sil
 dotnet_style_require_accessibility_modifiers = for_non_interface_members:error
 
 # Expression-level preferences
-dotnet_style_coalesce_expression = true:silent
-dotnet_style_collection_initializer = true:silent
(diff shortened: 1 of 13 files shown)


chinarosesz:
> #### Core EditorConfig Options ####

Since we faced an issue with editorconfig automatically remove unnecessary removing cast, should we consider this rule here?

    # IDE0004: Remove Unnecessary Cast
dotnet_diagnostic.IDE0004.severity = silent


---
Refers to: src/.editorconfig:7 in cfbafa4. [](commit_id = cfbafa42b75caa83b76322fb05f96418c94f1304, deletion_comment = False)

michaelcfanning:
> #### Core EditorConfig Options ####

yes indeed, thanks for chasing this down.

---
In reply to: [580966506](https://github.com/microsoft/binskim/pull/258#issuecomment-580966506) [](ancestors = 580966506)

---
Refers to: src/.editorconfig:7 in cfbafa4. [](commit_id = cfbafa42b75caa83b76322fb05f96418c94f1304, deletion_comment = False)

PR #259, opened 2020-02-01
Zap all underscore prefixes from member names.

michaelcfanning:
@chinarosesz per discussion.


src/BinSkim.Driver/RoslynAnalyzer/ActionMap.cs
```diff
@@ -16,26 +16,26 @@ namespace Microsoft.CodeAnalysis.IL
     /// <typeparam name="TKind"></typeparam>
```
chinarosesz: Nit: Remove un-commented params

chinarosesz:
>         #region IDisposable Support

I'm curious what you think about the usage of #region, it can be a bit of a controversial topic, but I find that it's difficult having to expand every time when viewing in Visual Studio. I know there is a setting to automatically expand all, but not all developers turn this on. Region can also be manipulated and encourage long classes, that's another reason why I chose to stop using it over time. 

Given that said, I think this might be generated by the IDE when you implement IDisposable and it inserts it by default. So maybe wasn't intentional.

---
Refers to: src/BinSkim.Sdk/BinaryAnalyzerContext.cs:70 in f2cea9c. [](commit_id = f2cea9c32878d3ed515033e5b413e94640915b82, deletion_comment = False)

michaelcfanning:
>         #region IDisposable Support

I believe in avoiding #region except when it's truly helpful for readability. This code does not meet that standard. There are also a lot of unhelpful/obvious comments in here. Will fix.

---
In reply to: [581515151](https://github.com/microsoft/binskim/pull/259#issuecomment-581515151) [](ancestors = 581515151)

---
Refers to: src/BinSkim.Sdk/BinaryAnalyzerContext.cs:70 in f2cea9c. [](commit_id = f2cea9c32878d3ed515033e5b413e94640915b82, deletion_comment = False)

PR #264, opened 2020-03-27
Changing BA2008 GUARD:CF -> guard:cf

JacksonTech:
As per Microsoft docs:

https://docs.microsoft.com/en-us/cpp/build/reference/guard-enable-control-flow-guard

/guard:cf is lowercase. MSVC's compiler options are case-sensitive and
will throw warnings if you pass in /GUARD:CF

The linker options are not case sensitive, so either one works.

This change replaces all /GUARD:CF instances with /guard:cf


michaelcfanning:
@JacksonTech , hope you are well. Thanks very much for this contribution. Unfortunately, I only noticed it after making an intrusive change to many test files (as you can see from merge conflicts here). And so I replaced your change to master and have shipped it. I put a shout-out to you in our release notes, hope that's ok.

This was an important correction to make, so thanks again. Would be interested to hear about your application of BinSkim if you have a moment.

PR #269, opened 2020-04-19
Fix Markdown formatting in ReleaseHistory.md

danpere:
Most (but not all) of the "NuGet Package" links were missing the final `)` so the full URL showed instead of "NuGet Package" being a link.


michaelcfanning:
Thanks everyone.

PR #283, opened 2020-04-30
Refine PDB loading and exception reporting.

michaelcfanning:


michaelcfanning:
Improve PDB trace output. Make PDB path probing more efficient (by removing duplicate probes when utilizing the --local-symbol-paths argument).

PR #339, opened 2021-01-25
Updating sarif-sdk

eddynaka:


antoniovs1029:
Awesome. @eddynaka , when will the --level flag be available on the Binskim.exe? And on Azure DevOps task? thanks

eddynaka:
In the next few weeks, we will release a new version on nuget.org.
I still don't have knowledge of the Azure DevOps task development, so I can't say when.

PR #340, opened 2021-01-26
Verify IntegrityCheck

eddynaka:


src/BinSkim.Rules/PERules/BA2025.EnableIntegrityCheck.cs
```diff
@@ -0,0 +1,67 @@
...
+            nameof(RuleResources.NotApplicable_InvalidMetadata)
+        };
+
+        public override void Analyze(BinaryAnalyzerContext context)
+        {
+            PEBinary target = context.PEBinary();
+            PEHeader peHeader = target.PE.PEHeaders.PEHeader;
+
+            DllCharacteristics dllCharacterstics = peHeader.DllCharacteristics;
+            if (!dllCharacterstics.HasFlag((DllCharacteristics)Enum.Parse(typeof(DllCharacteristics), ImageDllCharacteristicsForceIntegrity)))
```
michaelcfanning, line 46: >Enum.Parse(typeof(DllCharacteristics), ImageDllCharacteristicsForceIntegrity)) [](start = 63, length = 78)

Just embed a value of 0x80 here and cast it to DllCharacteristics. There's no good reason to parse a string representation to it.

src/BinSkim.Rules/PERules/BA2025.EnableIntegrityCheck.cs
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Microsoft.CodeAnalysis.IL.Rules
+{
+    [Export(typeof(Skimmer<BinaryAnalyzerContext>)), Export(typeof(ReportingDescriptor))]
+    public class EnableIntegrityCheck : PEBinarySkimmerBase
+    {
+        /// <summary>
+        /// https://docs.microsoft.com/windows/win32/debug/pe-format#dll-characteristics
+        /// </summary>
+        private const string ImageDllCharacteristicsForceIntegrity = "128";
```
michaelcfanning, line 22: >ImageDllCharacteristicsForceIntegrity [](start = 29, length = 37)

Convert this value to an int and give it a value of 0x80;

src/BinSkim.Rules/PERules/BA2025.EnableIntegrityCheck.cs
```diff
@@ -0,0 +1,67 @@
...
+
+using Microsoft.CodeAnalysis.BinaryParsers;
+using Microsoft.CodeAnalysis.IL.Sdk;
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.Driver;
+
+namespace Microsoft.CodeAnalysis.IL.Rules
+{
+    [Export(typeof(Skimmer<BinaryAnalyzerContext>)), Export(typeof(ReportingDescriptor))]
+    public class EnableIntegrityCheck : PEBinarySkimmerBase
```
michaelcfanning, line 17: >EnableIntegrityCheck [](start = 17, length = 20)

Let's rename this rule to 'RequireSignatureCheck' as that is the way that the feature is described in help.

src/BinSkim.Rules/PERules/BA2025.EnableIntegrityCheck.cs
```diff
@@ -0,0 +1,67 @@
...
+                return;
+            }
+
+            context.Logger.Log(this,
+                RuleUtilities.BuildResult(ResultKind.Pass, context, null,
+                    nameof(RuleResources.BA2025_Pass),
+                        context.TargetUri.GetFileName()));
+        }
+
+        public override AnalysisApplicability CanAnalyzePE(PEBinary target, PropertiesDictionary policy, out string reasonForNotAnalyzing)
```
michaelcfanning, line 61: >CanAnalyzePE [](start = 46, length = 12)

So, it appears that we should only be running this analysis for kernel mode code, which also means native. See the PE.IsKernelMode helper. Now that we know this feature is restricted to these binary kinds, we can elevate the severity of this rule to at least warning, I think. If we understood the exports of the 'certain Windows features' below, we could make it an error (i.e., we'd check the binary for use of those features).

"This option must be set for both 32-bit and 64-bit DLLs that implement kernel-mode code loaded by certain Windows features, and is recommended for all device drivers on Windows Vista, Windows 7, Windows 8, Windows Server 2008, and Windows Server 2012."

michaelcfanning:
You'll need to recompile your tests as kernel mode drivers, I think. I'm not sure this error will apply to exes, since it's a kernel mode thing...

michaelcfanning:
Strings:

HELP:
Forced Integrity checking is a policy that ensures a binary that is being loaded is signed prior to loading. The IMAGE_DLLCHARACTERISTICS_FORCE_INTEGRITY flag is set in the PE header at link time by using the /INTEGRITYCHECK linker flag to indicate that the binary being loaded must be signed. This flag causes the Windows memory manager to enforce a signature check at load time on the binary file. Applications check for presence of the flag when the binary is loaded in order to ensure that the identity of publisher is known. The Forced Integrity policy is enforced on Windows Vista, Windows Server 2008 and later releases when the IMAGE_DLLCHARACTERISTICS_FORCE_INTEGRITY flag is set. The Forced Integrity policy is enforced on both 32 bit and 64 bit systems and in both user and kernel mode. However, Force Integrity is not enforced for boot start drivers. The following Windows features enforce integrity checking on third-party binaries: Windows Security Center, Appinit.dll extensions, Object Manager filter registration, extended process filter registration APIs, and for Windows 7, Windows Biometric Framework, LSA extensions, and NTOS kernel extensions. There is a development cost associated with signing binaries with the IMAGE_DLLCHARACTERISTICS_FORCE_INTEGRITY flag set, so do not set the IMAGE_DLLCHARACTERISTICS_FORCE_INTEGRITY flag unless your component requires it. See [https://social.technet.microsoft.com/wiki/contents/articles/255.forced-integrity-signing-of-portable-executable-pe-files.aspx](https://social.technet.microsoft.com/wiki/contents/articles/255.forced-integrity-signing-of-portable-executable-pe-files.aspx) for more information.


FAIL:
'{0}' is a kernel mode binary that was not compiled with /INTEGRITYCHECK, requiring the Windows memory manager to check for a digital signature when loading the image.

PASS:
'{0}' is a kernel mode binary that was compiled with /INTEGRITYCHECK, requiring the Windows memory manager to check for a digital signature when loading the image.

PR #354, opened 2021-03-14
Adding error if pdb not loaded to rule BA4001

eddynaka:


src/BinSkim.Rules/PERules/WindowsBinaryAndPdbSkimmerBase.cs
```diff
@@ -40,10 +40,8 @@ public sealed override void Analyze(BinaryAnalyzerContext context)
                 target.Pdb.LoadTrace = null;
             }
 
-            if ((!target.PE.IsManaged ||
-                  target.PE.IsMixedMode ||
-                  EnforcePdbLoadForManagedAssemblies)
-                    && target.Pdb == null)
+            if ((!target.PE.IsManaged || target.PE.IsMixedMode) &&
+                (EnforcePdbLoadForManagedAssemblies && target.Pdb == null))
```
michaelcfanning, line 44: >EnforcePdbLoadForManagedAssemblies [](start = 17, length = 34)

Is this change right? Why are we consulting 'EnforcePdbLoadForManagedAssemblies' for cases where we have a managed assembly?

Did you try rerunning the functional tests? On surface, it looks like your change might stop reporting PDB load failures for native binaries. O

michaelcfanning (changes requested):
:clock1:

src/BinSkim.Rules/PERules/WindowsBinaryAndPdbSkimmerBase.cs
eddynaka: the previous condition:
if native or mixed mode or enforced && pdb == null -> we throw exception

the new condition:
if native or mixedmode && (enforce && pdb == null) -> we throw if enforce is true && pdb == null.

all tests are running (i re-enabled in a previous pr)

---
In reply to: [593830623](https://github.com/microsoft/binskim/pull/354#discussion_r593830623) [](ancestors = 593830623)

michaelcfanning (approved):
:shipit:

PR #356, opened 2021-03-21
Improving BA3010 check for full RELRO

### LINKED ISSUE #355: BA3010 doesn't distinguish between Full and Partial RELRO
The current RELRO checker does not seem to distinguish between Full and Partial RELRO. Full RELRO is discussed [here](https://www.trapkit.de/articles/relro/).

```
Analyzing 'test-allno'...
/binskim-bins/test-allno: error BA3001: PIE disabled on executable 'test-allno'.  This means the code section will always be loaded to the same address, even if ASLR is enabled in the Linux kernel.  To address this, ensure you are compiling with '-fpie' when using clang/gcc.
/binskim-bins/test-allno: error BA3002: Stack on 'test-allno' is executable, which means that an attacker could use it as a place to store attack shellcode.  Ensure you are compiling with '-z noexecstack' to mark the stack as non-executable.            
/binskim-bins/test-allno: error BA3003: The stack protector was not found in 'test-allno'.  This may be because the binary has no stack-based arrays, or because '--stack-protector-strong' was not used.                                                    
/binskim-bins/test-allno: error BA3010: The GNU_RELRO segment is missing from this binary, so relocation sections in 'test-allno' will not be marked as read only after the binary is loaded.  An attacker can overwrite these to redirect control flow.  Ensure you are compiling with the compiler flags '-Wl,z,relro' to address this.
/binskim-bins/test-allno: error BA3030: No checked functions are present/used when compiling 'test-allno', and it was compiled with GCC--and it uses functions that can be checked. The Fortify Source flag replaces some unsafe functions with checked versions when a static length can be determined, and can be enabled by passing '-D_FORTIFY_SOURCE=2' when optimization level 2 ('-O2') is enabled.  It is possible that the flag was passed, but that the compiler could not statically determine the length of any buffers/strings.
Analyzing 'test-full-relro'...
/binskim-bins/test-full-relro: error BA3030: No checked functions are present/used when compiling 'test-full-relro', and it was compiled with GCC--and it uses functions that can be checked. The Fortify Source flag replaces some unsafe functions with checked versions when a static length can be determined, and can be enabled by passing '-D_FORTIFY_SOURCE=2' when optimization level 2 ('-O2') is enabled.  It is possible that the flag was passed, but that the compiler could not statically determine the length of any buffers/strings.
Analyzing 'test-part-relro'...
/binskim-bins/test-part-relro: error BA3030: No checked functions are present/used when compiling 'test-part-relro', and it was compiled with GCC--and it uses functions that can be checked. The Fortify Source flag replaces some unsafe functions with checked versions when a static length can be determined, and can be enabled by passing '-D_FORTIFY_SOURCE=2' when optimization level 2 ('-O2') is enabled.  It is possible that the flag was passed, but that the compiler could not statically determine the length of any buffers/strings.
```

[checksec](https://www.trapkit.de/tools/checksec/) makes an appropriate distinction:

```
RELRO           STACK CANARY      NX            PIE             RPATH      RUNPATH      FILE
No RELRO        No canary found   NX disabled   No PIE          No RPATH   No RUNPATH   ./test-allno
Full RELRO      Canary found      NX enabled    PIE enabled     No RPATH   No RUNPATH   ./test-full-relro
Partial RELRO   Canary found      NX enabled    PIE enabled     No RPATH   No RUNPATH   ./test-part-relro
```

toshipiazza:
Fixes #355

### DIFF
--- a/src/BinSkim.Rules/ELFRules/BA3010.EnableReadOnlyRelocations.cs
+++ b/src/BinSkim.Rules/ELFRules/BA3010.EnableReadOnlyRelocations.cs
@@ -35,11 +35,11 @@ namespace Microsoft.CodeAnalysis.IL.Rules
         public override MultiformatMessageString FullDescription => new MultiformatMessageString { Text = RuleResources.BA3010_EnableReadOnlyRelocations_Description };
 
         protected override IEnumerable<string> MessageResourceNames => new string[] {
-                    nameof(RuleResources.BA3010_Pass),
-                    nameof(RuleResources.BA3010_Error_No_BindNow),
-                    nameof(RuleResources.BA3010_Error_No_Relro_Segment),
-                    nameof(RuleResources.NotApplicable_InvalidMetadata)
-                };
+            nameof(RuleResources.BA3010_Pass),
+            nameof(RuleResources.BA3010_Error_No_BindNow),
+            nameof(RuleResources.BA3010_Error_No_Relro_Segment),
+            nameof(RuleResources.NotApplicable_InvalidMetadata)
+        };
 
         public override AnalysisApplicability CanAnalyzeElf(ELFBinary target, Sarif.PropertiesDictionary policy, out string reasonForNotAnalyzing)
         {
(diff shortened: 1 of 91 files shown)


toshipiazza:
This change is stricter than the original rule; so I suspect a lot of Linux projects that previously had no errors reported by BinSkim might now see this error (BA3010). If this is undesirable, I can split it out into another check (BA3011?).

toshipiazza:
Closing in favor of #363, where I split the BIND_NOW check out into another warning (BA3011)

PR #363, opened 2021-04-12
Creates BA3011 check for presence of BIND_NOW flag AKA full RELRO

### LINKED ISSUE #355: BA3010 doesn't distinguish between Full and Partial RELRO

toshipiazza:
Fixes #355


michaelcfanning:
Toshi, thanks for this, we will review asap.

src/BinSkim.Rules/ELFRules/BA3011.EnableBindNow.cs
```diff
@@ -0,0 +1,111 @@
...
+                context.Logger.Log(this,
+                RuleUtilities.BuildResult(ResultKind.Pass, context, null,
+                    nameof(RuleResources.BA3011_Pass),
+                    context.TargetUri.GetFileName()));
+            }
+            else
+            {
+                // Fail
+                context.Logger.Log(this,
+                RuleUtilities.BuildResult(FailureLevel.Error, context, null,
```
eddynaka, line 105: >FailureLevel.Error [](start = 42, length = 18)

@michaelcfanning , should we set this to note? and in the next release, we change to error based on Matt's feedback.

src/BinSkim.Rules/ELFRules/BA3011.EnableBindNow.cs
```diff
@@ -0,0 +1,111 @@
...
+                {
+                    if ((entry.Tag == ELFSharp.ELF.Sections.DynamicTag.BindNow) ||
+                        (entry.Tag == ELFSharp.ELF.Sections.DynamicTag.Flags && (entry.Value & DF_BIND_NOW) != 0) ||
+                        (entry.Tag == ELFSharp.ELF.Sections.DynamicTag.Flags1 && (entry.Value & DF_1_NOW) != 0))
+                    {
+                        return true;
+                    }
+                }
+            }
+            catch (InvalidCastException)
```
michaelcfanning, line 103: >InvalidCastException [](start = 19, length = 20)

Just curious, what's producing these exceptions?

src/BinSkim.Rules/RuleIds.cs
```diff
@@ -48,8 +48,8 @@ internal static class RuleIds
 
         // Skipping some check namespace (BA3004-3009) for future checks
         public const string EnableReadOnlyRelocations = "BA3010";
+        public const string EnableBindNow = "BA3011";
```
michaelcfanning, line 51: >EnableBindNow [](start = 28, length = 13)

Good names! Let's add test cases for this rule, @eddynaka.

michaelcfanning:
Every new rule needs to either be added as 'Note' in the current release, or we need to add it to a major version revision. @eddynaka can you take care of this? Also, we need to document this great new check in our release notes. Finally, can you ensure that our comprehensive rules docs includes this check.

src/BinSkim.Rules/RuleIds.cs
toshipiazza: FYI there are some test cases for this already:

https://github.com/microsoft/binskim/blob/d9afb65c89a621411efded74c27999281d87867e/src/BuildSamples/ELF/build-test-binaries.sh#L59

https://github.com/microsoft/binskim/blob/d9afb65c89a621411efded74c27999281d87867e/src/BuildSamples/ELF/build-test-binaries.sh#L60

https://github.com/microsoft/binskim/blob/d9afb65c89a621411efded74c27999281d87867e/src/BuildSamples/ELF/build-test-binaries.sh#L73

Spoke offline with Eddy, and it may be helpful to update the no_immediate_binding test to explicitly use RTLD_LAZY binding: `-Wl,-z,lazy`

https://github.com/microsoft/binskim/blob/d9afb65c89a621411efded74c27999281d87867e/src/BuildSamples/ELF/build-test-binaries.sh#L74

src/BinSkim.Rules/ELFRules/BA3011.EnableBindNow.cs
toshipiazza: The project I'm working on separates executables into two files like in [this](https://stackoverflow.com/questions/46197810/separating-out-symbols-and-stripping-unneeded-symbols-at-the-same-time) post. The .dbg file contains only debug info, and therefore isn't even executable. But since it's an ELF, it gets parsed by `binskim analyze`. As a result, the output of binskim is meaningless, so I was happy to just silence the exception.

I haven't seen any other case where this happens.

michaelcfanning (approved):
:shipit:

PR #374, opened 2021-05-12
PEBinary should always initialize PE

eddynaka:


src/BinaryParsers/PEBinary/PEBinary.cs
```diff
@@ -22,6 +22,9 @@ public class PEBinary : BinaryBase
             string localSymbolDirectories = null,
             bool tracePdbLoad = false) : base(uri)
         {
+            // We actively verify our ability to parse this binary as a PE.
+            this.PE = new PE(this.TargetUri.LocalPath);
```
michaelcfanning, line 26: >LocalPath [](http://example.com/codeflow?start=44&length=9)

Isn't this going to break PDB only analysis?

Did you try running this version against a standalone PDB?

PR #376, opened 2021-05-14
Enable shadow-stack

### LINKED ISSUE #277: [RULE REQUEST] Enforce CET shadow stack compatibility
# Rule Proposal: Enforce CET shadow stack compatibility to defend against return-oriented programming (ROP) based malware attacks.

#### Rule metadata
- *Id*: BA2025
- *Name*: EnableControlEnforcementTechnologyShadowStack
- Failure level: Warning

#### Rule descriptions
- *Short*: TBD
- *Full*: TBD

#### Platform & applicability
- *Platform*: Windows
- *Applicable to*: native *with PDBs* (required for this analysis)
- *Not applicable to*: *nix binaries

#### User-facing strings
- *Fail*: '{0}' does not enable the Control-flow Enforcement Technology (CET) Shadow Stack mitigation. To resolve this issue, pass /CETCOMPAT on the linker command lines.
- *Pass*:  '{0}' enables the Control-flow Enforcement Technology (CET) Shadow Stack mitigation.

- [CETCOMPAT](https://docs.microsoft.com/en-us/cpp/build/reference/cetcompat?view=msvc-160)
- [Extended DLL Characteristics](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#extended-dll-characteristics)

#### Notes on implementation
- Linker option is only available starting in Visual Studio 2019 version 16.7. Rule should therefore examine the linker version details to ensure the toolchain that compiled the binary supports the flag. If the toolchain is too old, the rule `CanAnalyze` should return `MetadataConditions.ImageCompiledWithOutdatedTools`. [`BA2008.EnableControlFlowGuard`](https://github.com/microsoft/binskim/blob/master/src/BinSkim.Rules/PERules/BA2008.EnableControlFlowGuard.cs) has an example of this pattern.
- It is occasionally helpful to provide a rule property that controls the version parameterization (this is mostly helpful to overcome issues in toolchain versioning). It's not clear this rule requires this kind of configurability. [`BA2008.EnableControlFlowGuard`](https://github.com/microsoft/binskim/blob/master/src/BinSkim.Rules/PERules/BA2008.EnableControlFlowGuard.cs) has an example of providing a knob for minimal linker version that can be used as a model, however.
- User-facing output should summarize key details from [/CETCOMPAT linker option](https://docs.microsoft.com/en-us/cpp/build/reference/cetcompat?view=vs-2017). description and additional  [security mitigation](https://techcommunity.microsoft.com/t5/windows-kernel-internals/understanding-hardware-enforced-stack-protection/ba-p/1247815) guidance.
- Rule metadata should capture relevant expanded information and links to external guidance.
- Note that this PE bit is ignored in cases where feature is unsupported by Windows or runtime hardware. This mitigation is an opt-in Windows feature, allowing some flexibility in requiring compiling with /CETCOMPAT (allowing app developers to opt into actual usage).

#### How to resolve
/CETCOMPAT

eddynaka:
Fixes #277


src/BinSkim.Rules/PERules/BA2025.EnableShadowStack.cs
```diff
@@ -0,0 +1,96 @@
...
+            nameof(RuleResources.NotApplicable_InvalidMetadata)
+        };
+
+        public override AnalysisApplicability CanAnalyzePE(PEBinary target, Sarif.PropertiesDictionary policy, out string reasonForNotAnalyzing)
+        {
+            PE portableExecutable = target.PE;
+            AnalysisApplicability result = AnalysisApplicability.NotApplicableToSpecifiedTarget;
+
+            reasonForNotAnalyzing = MetadataConditions.ImageIsILOnlyAssembly;
+            if (portableExecutable.IsILOnly) { return result; }
```
shaopeng-gh, line 49: other way is to move the assign statement inside the 
if() {assign statement;return result; }
since it is only needed when if is true

eddynaka: pls, do the proper changes :)

shaopeng-gh: got it, updated.

PR #380, opened 2021-05-25
Loading Pdb if not found

eddynaka:


src/BinaryParsers/PEBinary/ProgramDatabase/Pdb.cs
```diff
@@ -86,10 +86,18 @@ private void Init(string peOrPdbPath, string symbolPath, string localSymbolDirec
             }
             catch (COMException ce)
             {
-                throw new PdbException(ce)
+                var pdbException = new PdbException(ce)
                 {
                     LoadTrace = this.loadTrace?.ToString()
                 };
+
+                if (pdbException.ExceptionCode == DiaHresult.E_PDB_NOT_FOUND && peOrPdbPath.EndsWith(".dll", StringComparison.OrdinalIgnoreCase))
```
michaelcfanning, line 94: >pdbException.ExceptionCode [](http://example.com/codeflow?start=20&length=26)

Why are we entering this code path? A path with '.dll' should not arrive here. btw - if we did intend to handle things in the way you've proposed, we would need to handle '.exe' extension and '.sys'.

In general, however, this is a PDB class. it's not the job of the PDB class to correct the situation if someone upstream has tried to initialize it with something that isn't a PDB. So I'd tend to go examine where that is occurring and look at adding path processing there.

PDB loading/resolution is a very tricky topic and the logic has evolved considerably over time. We should consider investing more in testing here, perhaps with mocking to simulate all necessary conditions. Something to keep in mind.

michaelcfanning (changes requested):
:clock1:

src/BinaryParsers/PEBinary/ProgramDatabase/Pdb.cs
eddynaka: when we analyze with the following command:
`analyze "C:\Users\ednakamu\Desktop\binskim\*.dll"`, the value of peOrPdbPath is equal to the dll.

When that happens, we will try to load normally using the method `loadDataForExe`. But that sometimes fails with PDB_NOT_FOUND. In that case, we want to try again reading the pdb manually.

What I tested:
- multiple  combinations of the boolean that configure the reader
- with and without symbolserver

if thats the case, i could make the change in the PEBinary, adding this logic near the caught clause of the LoadPdb method.

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -609,15 +613,10 @@ internal static ManagedPlatform ComputeIsDotNetCore(MetadataReader metadataReade
...
-
-                    default:
-                    {
-                        break;
-                    }
                 }
             }
 
-            throw new InvalidOperationException("Could not identify managed platform.");
+            return ManagedPlatform.Unknown;
```
michaelcfanning, line 619: We now no longer will get early insight that we have an unknown platform. Why is not failing fast here a good idea?

Also, in general, we don't want to remove default handlers in switch statements. If you want this change, you should move it into the default handler. I'm not convinced (yet!) it's a good idea to eliminate the throw. What are your thoughts?

src/BinaryParsers/PEBinary/PEBinary.cs
```diff
@@ -121,5 +93,54 @@ public static bool CanLoadBinary(Uri uri)
             catch (IOException) { return false; }
             catch (UnauthorizedAccessException) { return false; }
         }
+
+        private Pdb LoadPdb()
+        {
+            string pePath = this.PE?.FileName ?? this.TargetUri.LocalPath;
+            if (!TryLoadPdb(pePath, this.symbolPath, this.localSymbolDirectories, this.tracePdbLoad, out Pdb pdb))
```
michaelcfanning, line 100: >TryLoadPdb [](http://example.com/codeflow?start=17&length=10)

If you know that a file with an exe or dll extension will fail, why don't you check for that upfront? Having a TryLoadPdb call nested inside this is a clue that you should consider clarifying inputs upfront.

eddynaka: normally, the first try points to the binary itself and when that happens, we  try to load the pdb using `loadDataForExe`.

But, even in cases where the dll and pdb are side by side, the `loadDataForExe` can not load, so, i'm forcing loading only the pdb, calling `loadDataFromPdb`.

i understood ur concern, but if we read directly the pdb, it will be easier, but we would change the behavior...

we can talk more later to discuss :)

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
eddynaka: This is preventing me to analyze some binaries...

When we have a dll that has aacorlib as reference, we throw an exception, but in that case, the dll is a valid PE to analyze.

we can also discuss later :)

michaelcfanning: send me a binary that throws bad references and I will take a look. the issue here is that someone is now produce PEs where we fundamentally don't understand the platform. we need to fix this! note that it is these binaries in many cases that *are* in fact .net core! and so it's especially important to chase that down given the purpose of this helper.

this is one of those places where the lack of curated metadata is a real disadvantage. i.e., our detections here are heuristic based and fragile over time, as frameworks and compilers change....

eddynaka: added aacorlib as netframework.

michaelcfanning:
I like the change but can we please add some tests? It's fine to check this in and then go author new tests.

michaelcfanning (approved):
:shipit:

PR #389, opened 2021-06-18
Adding supports for analyzing Fat MachO files.

yongyan-gh:
Fat MachO file can have more than 1 binary files inside the file.


src/BinaryParsers/MachOBinary/MachOBinary.cs
```diff
@@ -52,61 +43,54 @@ public static bool CanLoadBinary(Uri uri)
         {
             try
             {
-                return MachOReader.TryLoad(Path.GetFullPath(uri.LocalPath), out _) == MachOResult.OK;
+                return MachOReader.TryLoadFat(
```
eddynaka, line 46: >TryLoadFat [](http://example.com/codeflow?start=35&length=10)

what would happen if the binary isn't FAT?

yongyan-gh: if a FAT MachO it returns  MachOResult.FatMachO
if normal MachO (has only 1 binary) it returns MachOResult.OK
if its not a MachO, it returns MachOResult.NotMachO

eddynaka (approved):

PR #404, opened 2021-07-07
Adding compiler data to AppInsights

eddynaka:


src/BinaryParsers/MachOBinary/MachOBinary.cs
```diff
@@ -68,7 +68,7 @@ public static bool CanLoadBinary(Uri uri)
         /// <summary>
         /// Unit type of Dwarf used..
```
michaelcfanning, line 69: >.. [](http://example.com/codeflow?start=35&length=2)

Duplicated period here.

michaelcfanning:
>         private const string SECTIONNAME_DEBUG_INFO = "__debug_info";

these strings are neither alphabetized nor are they sorted by line length. 

can we sort by line length? makes things more readable.

---
Refers to: src/BinaryParsers/MachOBinary/SingleMachOBinary.cs:16 in aedb1eb. [](commit_id = aedb1eb47e2feb4964603108537f12f87f565d72, deletion_comment = False)

src/BinSkim.Driver/AnalyzeCommand.cs
```diff
@@ -43,6 +43,8 @@ protected override BinaryAnalyzerContext CreateContext(AnalyzeOptions options, I
                 ShouldWarnVerbose = false;
             }
 
+            binaryAnalyzerContext.CompilerDataLogger = new CompilerDataLogger(binaryAnalyzerContext, options.RepositoryUri, options.PipelineName);
```
michaelcfanning, line 46: >CompilerDataLogger [](http://example.com/codeflow?start=34&length=18)

Why do we always create this? Instead of conditionally?

src/BinSkim.Driver/AnalyzeCommand.cs
```diff
@@ -68,6 +70,8 @@ public override int Run(AnalyzeOptions analyzeOptions)
 
             int result = base.Run(analyzeOptions);
 
+            CompilerDataLogger.Flush();
```
michaelcfanning, line 73: >Flush [](http://example.com/codeflow?start=31&length=5)

If a flush is really required should this be in a finally block somewhere? This won't execute in a case where Run raises an exception

src/BinSkim.Driver/AnalyzeCommand.cs
eddynaka: I'm creating every time  because only here i will have all  the information (pipeline, repository).

the appinsights part is static, so it wont keep re-creating it.

src/BinSkim.Driver/AnalyzeOptions.cs
```diff
@@ -43,5 +43,15 @@ public class AnalyzeOptions : AnalyzeOptionsBase
             HelpText = "Emit verbose output. The resulting comprehensive report is designed to provide appropriate evidence for compliance scenarios.")]
         [Obsolete("Use --level and --kind instead.")]
         public bool Verbose { get; set; }
+
+        [Option(
+            "repository-uri",
+            HelpText = "RepositoryUri is required to rule BA4001 and BA4002.")]
```
michaelcfanning, line 49: >RepositoryUri is required to rule BA4001 and BA4002 [](http://example.com/codeflow?start=24&length=51)

Whenever you show an opaque rule id to the user, you should add the readable name. Not everyone can remember what BA4001 is. Also, you've described a fact here 'this data is required to run rule BA4001.ReportPECompilerData' but you haven't described what that data is.

What is this data? Does it make sense that we require a single repo URL for a scan run??

src/BinSkim.Driver/AnalyzeOptions.cs
```diff
@@ -43,5 +43,15 @@ public class AnalyzeOptions : AnalyzeOptionsBase
...
+        public string RepositoryUri { get; set; }
+
+        [Option(
+            "pipeline-name",
+            HelpText = "PipelineName is required to rule BA4001 and BA4002.")]
```
michaelcfanning, line 54: >PipelineName [](http://example.com/codeflow?start=24&length=12)

same note, what is this data? you don't describe it.

src/BinSkim.Rules/DwarfRules/BA4002.ReportDwarfCompilerData.cs
```diff
@@ -91,15 +91,7 @@ private void PrintCompilerData(BinaryAnalyzerContext context, string language, I
...
-                    Console.Write($"{context.TargetUri.LocalPath},");
-                    Console.Write($"{compiler.Compiler},");
-                    Console.Write($"{compiler.Version},");
-                    Console.Write($"{compiler.Version},");
-                    Console.Write($"{language},");
-                    Console.Write($"{file},");
-                    Console.Write($"{file},");
-                    Console.Write($"{context?.Hashes?.Sha256},");
-                    Console.WriteLine();
+                    context.CompilerDataLogger.Write(compiler.Compiler.ToString(), compiler.Version.ToString(), language, file);
```
michaelcfanning, line 94: >CompilerDataLogger [](http://example.com/codeflow?start=28&length=18)

why is it ok that some data has gotten dropped? such as the sha256 has?

src/BinSkim.Rules/PERules/BA4001.ReportPECompilerData.cs
```diff
@@ -48,14 +48,14 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
...
+                context.CompilerDataLogger.PrintHeader();
                 PrintHeader = false;
             }
 
             if (pdb == null)
             {
-                Console.Write(context.TargetUri.LocalPath + ",");
-                Console.WriteLine($",,,,,,{context?.Hashes?.Sha256},{target.PdbParseException.Message}");
+                string errorMessage = target.PdbParseException.Message;
+                context.CompilerDataLogger.WriteException(errorMessage);
```
michaelcfanning, line 58: >CompilerDataLogger [](http://example.com/codeflow?start=24&length=18)

what about the hash data?

src/BinSkim.Sdk/ICompilerDataLogger.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.CodeAnalysis.BinaryParsers.ProgramDatabase;
+
+namespace Microsoft.CodeAnalysis.IL.Sdk
+{
+    public interface ICompilerDataLogger
```
michaelcfanning, line 8: >ICompilerDataLogger [](http://example.com/codeflow?start=21&length=19)

why did you create an interface?

src/BinSkim.Rules/DwarfRules/BA4002.ReportDwarfCompilerData.cs
eddynaka: where I create the CompilerDataLogger, i pass the context. from it I retrieve information such as hash. With that, i can remove from here and just pass the data that i dont have during construction of compilerDataLogger

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            if (!appInsightsRegistered)
+            {
+                Console.WriteLine("RepositoryUri,PipelineName,Target,Compiler Name,Compiler BackEnd Version,Compiler FrontEnd Version,Language,Dialect,Module Name,Module Library,Hash,Error");
+            }
+        }
+
+        public void Write(string compilerData, ObjectModuleDetails omDetails)
+        {
+            string name = omDetails?.Name?.Replace(",", "_");
```
michaelcfanning, line 72: >name [](http://example.com/codeflow?start=19&length=4)

are we really transforming data that previously existed in the PDB into some other normalized form? 

this doesn't seem like a good idea, it seems better to preserve data as it actually exists in the binary.

can you explain the decision? certainly this code needs to be commented if we retain what you've done.

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -0,0 +1,172 @@
...
+            string library = omDetails?.Library?.Replace(",", ";");
+            if (appInsightsRegistered)
+            {
+                string[] compilerDataParts = compilerData.Split(',');
+
+                s_telemetryClient.TrackEvent(CompilerEventName, properties: new Dictionary<string, string>
+                {
+                    { "repositoryUri", repositoryUri ?? string.Empty },
```
michaelcfanning, line 80: >repositoryUri [](http://example.com/codeflow?start=39&length=13)

you say in your help comments that this data is required, but it doesn't look like it's required. it's optional.

i think you meant you say that this data is only consumed by the compiler reporting rules, and not relevant otherwise.

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -0,0 +1,172 @@
...
+                    { "pipelineName", pipelineName ?? string.Empty },
+                    { "target", this.localPath },
```
michaelcfanning, line 82: >localPath [](http://example.com/codeflow?start=37&length=9)

why are we sending absolute paths? is this right? might this data expose some unwanted information, like a user alias?

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -0,0 +1,172 @@
...
+                    { "compilerName", compilerDataParts[0] },
+                    { "compilerBackEndVersion", compilerDataParts[1] },
```
michaelcfanning, line 84: >compilerDataParts[1] [](http://example.com/codeflow?start=48&length=20)

look at these 'magic constants' where we have special knowledge that backend version is in the first piece. 

that tends to be pretty fragile, it might be better for this helper to break out all this data and require the caller to parse into the format.

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -0,0 +1,172 @@
...
+                    { "dialect", string.Empty },
+                    { "moduleName", name ?? string.Empty },
+                    { "moduleLibrary", (name == library ? string.Empty : library) },
+                    { "hash", this.sha256 ?? string.Empty },
+                    { "error", string.Empty }
+                });
+            }
+            else
+            {
+                Console.Write($"{repositoryUri},");
```
michaelcfanning, line 96: >Write [](http://example.com/codeflow?start=24&length=5)

Note that it'd be more performant to concatenate all this information into a single string and then pass it to WriteLine.

michaelcfanning (changes requested):
:clock1:

src/BinaryParsers/MachOBinary/MachOBinary.cs
eddynaka: fixed.

src/BinSkim.Driver/AnalyzeCommand.cs
eddynaka: yeah, AppInsights batches the metrics, so we will need to flush.
Added a try/finally.

src/BinSkim.Driver/AnalyzeCommand.cs
eddynaka: added some conditions to prevent creating this class and not using.
thanks for the feedback.

src/BinSkim.Sdk/ICompilerDataLogger.cs
eddynaka: just deleted.
i thought it was going to be useful.

src/BinSkim.Rules/PERules/BA4001.ReportPECompilerData.cs
eddynaka: the hash is constructed with the CompilerDataLogger, so we don't need to worry here.

eddynaka:
>         private const string SECTIONNAME_DEBUG_INFO = "__debug_info";

fixed.

---
In reply to: [881621666](https://github.com/microsoft/binskim/pull/404#issuecomment-881621666) [](http://example.com/codeflow?ancestors=881621666,881621666)

---
Refers to: src/BinaryParsers/MachOBinary/SingleMachOBinary.cs:16 in aedb1eb. [](commit_id = aedb1eb47e2feb4964603108537f12f87f565d72, deletion_comment = False)

src/BinSkim.Sdk/CompilerDataLogger.cs
eddynaka: fixed

src/BinSkim.Sdk/CompilerDataLogger.cs
eddynaka: we were creating a csv and we were using ',' as separator.
with that, I saw cases where the name contained ',' breaking the csv.

this is why i changed from ',' to '_'

src/BinSkim.Sdk/CompilerDataLogger.cs
eddynaka: I changed the relevant code to throw an  argumentNullException in case repositoryUri or pipelineName is null/empty.

With that, i cleaned the code

src/BinSkim.Driver/AnalyzeOptions.cs
eddynaka: I changed the helptext. Let me know what do you think.

About your question...I thought about it, and  I could not find anything better based on this:
[https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml)

my idea was  to find one pre-defined variable that any pipeline would be  able to get.

src/BinSkim.Driver/AnalyzeOptions.cs
eddynaka: updated the helptext, pls, review :)

src/BinSkim.Sdk/CompilerDataLogger.cs
eddynaka: made a change to transform the localPath to a relativeFilePath base on the targetFileSpecifiers

src/BinSkim.Sdk/CompilerDataLogger.cs
eddynaka: fixed!

src/BinSkim.Sdk/CompilerDataLogger.cs
michaelcfanning: typically, what you would do to workaround this would be to encapsulate the data in double quotes. is that an option here? 

we really should not be modifying the compiler generated data here. how will you convert it back if you need to unify it again with actual PDB data?

michaelcfanning (approved):
:shipit:

PR #413, opened 2021-07-20
Add PIE rules for MachO binaries

yongyan-gh:


src/BinSkim.Rules/MachORules/BA5001.EnablePositionIndependentExecutable.cs
```diff
@@ -0,0 +1,107 @@
...
+    [Export(typeof(Skimmer<BinaryAnalyzerContext>)), Export(typeof(ReportingDescriptor))]
+    public class EnablePositionIndependentExecutableMachO : MachOBinarySkimmer
+    {
+        /// <summary>
+        /// BA5001
+        /// </summary>
+        public override string Id => RuleIds.EnablePositionIndependentExecutableMachO;
+
+        /// <summary>
+        /// "A Position Independent Executable (PIE) relocates all of its sections at load time, including the code section,
```
eddynaka, line 26: >" [](http://example.com/codeflow?start=12&length=1)

you can remove the double-quotes here and in the end.

src/BinSkim.Rules/MachORules/BA5001.EnablePositionIndependentExecutable.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            return (macho.MachO.FileType == ELFSharp.MachO.FileType.Executable ||
+                    macho.MachO.FileType == ELFSharp.MachO.FileType.DynamicLibrary);
+        }
+
+        private static bool HasPieFlag(SingleMachOBinary binary)
+        {
+            if (binary.MachO.FileType == ELFSharp.MachO.FileType.DynamicLibrary)
+            {
+                // for shared libaries the compiler adds -fPIC by default
```
eddynaka, line 93: >fPIC [](http://example.com/codeflow?start=58&length=4)

what if u have a sharedlibrary and manually remove? what would happen? is it possible?

yongyan-gh: yes PIC can be disabled by option `-mdynamic-no-pic`
https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/MachOTopics/1-Articles/dynamic_code.html 

Also there is a check in https://liquid.microsoft.com/Web/Object/Read/MS.Security/Requirements/Microsoft.Security.SystemsADM.10203#Zguide says
>Do not enable any of the following flags:
>
>Allow Executable Stack (--allow_stack_execute)
>Remove PIC References (--mdynamic-no-pic)

Should we keep them in 2 rules or check both condition in current rule?

eddynaka: The point here is: we should always check even if this is a dynamiclibrary or not.
Because, as you mentioned, anyone could disable it and if thats disabled, u would need to flag it, right?

yongyan-gh: I see, the mach-O libraries don't have a flag similar to executables PIE flag in file header. Need to investigate more about how to tell if library is built with -fPIC option.

yongyan-gh: I double checked the Apple's article about PIC https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/MachOTopics/1-Articles/dynamic_code.html#//apple_ref/doc/uid/TP40002528

I think my previous understanding was wrong. It looks like the Position-Independent Code (PIC) technique is built-in with Mach-O
>Mach-O position-independent code design is based on the observation that the __DATA segment is always located at a constant offset from the __TEXT segment. That is, the dynamic loader, when loading any Mach-O file, never moves a file’s __TEXT segment relative to its __DATA segment. Therefore, a function can use its own current address plus a fixed offset to determine the location of the data it wishes to access. All segments of a Mach-O file, not only the __TEXT and __DATA segments, are at fixed offsets relative to the other segments.

But with option `-mdynamic-no-pic`, it still accesses indirect addressing through relocation section.
> This option both reduces the code size of application executables and improves their performance by eliminating position-independent code references,  ___while preserving indirect calls to shared libraries and indirection to undefined symbols.___

eddynaka: I think we can do everything in this rule:
1. check if the user has enabled pie -> if yes, pass
2. check if the user has enabled no pie flag -> if yes, error

what do u think?

src/BinSkim.Rules/MachORules/BA5001.EnablePositionIndependentExecutable.cs
```diff
@@ -0,0 +1,119 @@
...
+    public class EnablePositionIndependentExecutableMachO : MachOBinarySkimmer
+    {
+        /// <summary>
+        /// BA5001
+        /// </summary>
+        public override string Id => RuleIds.EnablePositionIndependentExecutableMachO;
+
+        /// <summary>
+        /// A Position Independent Executable (PIE) relocates all of its sections at load time, including the code section,
+        ///  if ASLR is enabled in the Linux kernel (instead of just the stack/heap).  This makes ROP-style attacks more difficult.
```
eddynaka, line 26: >   [](http://example.com/codeflow?start=85&length=2)

can you clean this double space?
also, in the beginning we have a space. can you also delete that?

src/BinSkim.Rules/MachORules/BA5001.EnablePositionIndependentExecutable.cs
```diff
@@ -0,0 +1,119 @@
...
+                // for executables, checking file's header flag has PIE flag
+                // https://developer.apple.com/library/archive/qa/qa1788/_index.html
+                return binary.MachO.Flags.HasFlag(ELFSharp.MachO.HeaderFlags.PIE);
+            }
+
+            if (binary.MachO.FileType == ELFSharp.MachO.FileType.DynamicLibrary ||
+                binary.MachO.FileType == ELFSharp.MachO.FileType.Object ||
+                binary.MachO.FileType == ELFSharp.MachO.FileType.Debug)
+            {
+                // for libaries, check if compiler includes option "mdynamic-no-pic"
```
eddynaka, line 106: >libaries [](http://example.com/codeflow?start=23&length=8)

libraries

src/BinSkim.Rules/MachORules/BA5001.EnablePositionIndependentExecutable.cs
```diff
@@ -0,0 +1,119 @@
...
+                     macho.MachO.FileType == ELFSharp.MachO.FileType.Debug) &&
+                     IsValidDwarfBinary(macho));
+
+        }
+
+        private static bool HasPieFlag(SingleMachOBinary binary)
+        {
+            if (binary.MachO.FileType == ELFSharp.MachO.FileType.Executable)
+            {
+                // for executables, checking file's header flag has PIE flag
```
eddynaka, line 97: >checking [](http://example.com/codeflow?start=36&length=8)

use the same verbal time. 'for executables, check if...'

eddynaka (approved):
:shipit:

PR #450, opened 2021-08-10
Handle null reference expcetion when reading PE header from pdb files

yongyan-gh:
PDB files don't have PEHeaders, adding checks for null cases.


eddynaka:
We can close this, since we fixed with the other PR.

PR #467, opened 2021-08-24
Send command line by chunk

shaopeng-gh:


src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -120,11 +123,11 @@ public void Write(CompilerData compilerData, ObjectModuleDetails omDetails)
             string library = omDetails?.Library;
             if (this.appInsightsRegistered)
             {
-                s_telemetryClient.TrackEvent(CompilerEventName, properties: new Dictionary<string, string>
+                string commandLineId = Guid.NewGuid().ToString();
```
eddynaka, line 126: >commandLineId [](http://example.com/codeflow?start=23&length=13)

create only when you are sure that you have a commandLine.
otherwise, you will create something that will be discarted

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -208,5 +235,23 @@ public void WriteException(string errorMessage)
                 Console.WriteLine(log);
             }
         }
+
+        private void SendChunkedCommandLine(string sessionId, string commandLineId, string commandLine)
```
eddynaka, line 239: >string sessionId [](http://example.com/codeflow?start=44&length=16)

In theory, you don't need the sessionId, since its a static variable.

eddynaka (approved):
:shipit:

PR #470, opened 2021-08-27
Fix read data out of boundary cause System.AccessViolationException

shaopeng-gh:


src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
```diff
@@ -249,6 +249,11 @@ public byte[] ReadBlock(uint size)
         /// <param name="position">The position.</param>
         public byte[] ReadBlock(uint size, int position)
         {
+            if (position >= Data.Length)
```
eddynaka, line 252: >position >= Data.Length [](http://example.com/codeflow?start=16&length=23)

is it possible to create a binary to test this cases?

shaopeng-gh: yes! have no idea currently but let me check

shaopeng-gh: not able to get the exact same issue but I found another issue which is similar out of boundary. it is when compiled with special machine type.
fixed the error and added 2 test files.

eddynaka (approved):

src/ReleaseHistory.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* BUGFIX: Fix exception System.AccessViolationException caused by trying to read data out of boundary. [#470](https://github.com/microsoft/binskim/pull/470)
```
eddynaka, line 5:
```suggestion
* BUGFIX: Fix exception `System.AccessViolationException` caused by trying to read data out of boundary. [#470](https://github.com/microsoft/binskim/pull/470)
```

PR #474, opened 2021-09-01
Include C++ runtime in the package

### LINKED ISSUE #473: Update BinSkim to redistribute msvcp140.dll in its install directory
This dependency is required by our native PDB reading binaries (which are themselves also redistributable).

@eddynaka @yongyan-gh @shaopeng-gh 

This is a high priority issue, can we get on it immediately?

eddynaka:
Fixes #473


src/ReleaseHistory.md
```diff
@@ -5,6 +5,7 @@
 * BUGFIX: Fix exception `System.AccessViolationException` caused by trying to read data out of boundary. [#470](https://github.com/microsoft/binskim/pull/470)
 * BUGFIX: Fix exception handling when PDB cannot be loaded by `IDiaDataSource`. [#461](https://github.com/microsoft/binskim/pull/461)
 * BREAKING: PDB exceptions will be reported once per target. [#465](https://github.com/microsoft/binskim/pull/465)
+* BUGFIX: Include C++ runtime in the package to prevent `DllNotFoundException` when loading `msdia140.dll`. [#474](https://github.com/microsoft/binskim/pull/474)
```
shaopeng-gh, line 8: >msdia140.dll [](http://example.com/codeflow?start=93&length=12)

is the issue missing dependency to load this dll?
I think we will need 2 sets for x86 and x64:
refs\x64\*
refs\x84\*

or it seems like we currently coded to use x64 only, I guess your new files are all x64.
do you think put in a x64 folder so it is easy to understand:
refs\runtime\x64\*

eddynaka: we only ship x64 versions and it is redundant to have x64 folder.

We can create a new PR cleaning up, removing the extra folders and keep everything in the same level :)

shaopeng-gh: only ship x64 versions --- got it!

PR #482, opened 2021-09-09
Fix opcode handling when reading DWARF line number programs

shaopeng-gh:


src/BinaryParsers/ElfBinary/Dwarf/DwarfEnums.cs
```diff
@@ -415,6 +415,9 @@ public enum DwarfAttribute
     /// </summary>
     public enum DwarfLineNumberStandardOpcode
     {
+        // Standard opcodes (1 to 12)
+        // Extended opcodes (0)
```
eddynaka, line 419: order  by length

eddynaka (approved):

PR #490, opened 2021-09-20
Add test files for different debug options

shaopeng-gh:
1. We are able to parse all of: Full, Portable, PdbOnly, Embedded.
for None it is pdb not found as expected.
2. Also add test files for Rust language and VB language . We are able to parse.


eddynaka (approved):
:shipit:

PR #493, opened 2021-09-28
Send assembly references to Telemetry

yongyan-gh:
Read assembly references from managed assemblies and send to telemetry by chunks.

### DIFF
--- a/src/BinSkim.Sdk/CompilerDataLogger.cs
+++ b/src/BinSkim.Sdk/CompilerDataLogger.cs
@@ -18,6 +18,7 @@ namespace Microsoft.CodeAnalysis.IL.Sdk
         private const int ChunkSize = 8192;
         private const string CompilerEventName = "CompilerInformation";
         private const string CommandLineEventName = "CommandLineInformation";
+        private const string AssemblyReferencesEventName = "AssemblyReferencesInformation";
         private const string SummaryEventName = "AnalysisSummary";
 
         private readonly bool appInsightsRegistered;
@@ -126,6 +127,7 @@ namespace Microsoft.CodeAnalysis.IL.Sdk
             if (TelemetryEnabled)
             {
                 string commandLineId = string.Empty;
+                string assemblyReferencesId = string.Empty;
                 var properties = new Dictionary<string, string>
                 {
                     { "target", this.relativeFilePath },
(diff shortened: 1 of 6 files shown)


src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -15,11 +15,12 @@ namespace Microsoft.CodeAnalysis.IL.Sdk
 {
     public class CompilerDataLogger
     {
-        private const int ChunkSize = 8192;
         private const string CompilerEventName = "CompilerInformation";
         private const string CommandLineEventName = "CommandLineInformation";
+        private const string AssemblyReferencesEventName = "AssemblyReferencesInformation";
         private const string SummaryEventName = "AnalysisSummary";
 
+        private readonly int ChunkSize = 8192;
```
eddynaka, line 23: >8192 [](http://example.com/codeflow?start=41&length=4)

u can remove this, since u are using the value from the constructor

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -15,11 +15,12 @@ namespace Microsoft.CodeAnalysis.IL.Sdk
 {
     public class CompilerDataLogger
     {
-        private const int ChunkSize = 8192;
         private const string CompilerEventName = "CompilerInformation";
         private const string CommandLineEventName = "CommandLineInformation";
+        private const string AssemblyReferencesEventName = "AssemblyReferencesInformation";
         private const string SummaryEventName = "AnalysisSummary";
```
eddynaka, line 21: >SummaryEventName [](http://example.com/codeflow?start=29&length=16)

move this to the first, so we will have ordered by length

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -32,20 +33,32 @@ public class CompilerDataLogger
...
+
+            if (!TelemetryEnabled)
             {
-                string appInsightsKey = RetrieveAppInsightsKey();
-                if (!string.IsNullOrEmpty(appInsightsKey) && Guid.TryParse(appInsightsKey, out _))
+                try
                 {
-                    Initialize(appInsightsKey);
-                    this.appInsightsRegistered = true;
+
```
eddynaka, line 50: can u remove thisi new line?

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -32,20 +33,32 @@ public class CompilerDataLogger
...
+                                  TelemetryClient telemetryClient = null,
+                                  int chunkSize = 8192)
         {
-            try
+            s_telemetryConfiguration = telemetryConfiguration;
+            s_telemetryClient = telemetryClient;
+            s_sessionId = TelemetryEnabled ? Guid.NewGuid().ToString() : null;
+            ChunkSize = chunkSize;
+
+            if (!TelemetryEnabled)
```
shaopeng-gh, line 46: >if (!TelemetryEnabled) [](http://example.com/codeflow?start=12&length=22)

question, why we add this new condition

yongyan-gh: the code in below block is to initialize a TelemetryClient instance to be used to send telemetry data.
if caller already passed in the TelemetryClient instance in constructor, we don't need to initialize a new instance.
Added the TelemetryClient / Configuration as constructor parameter so that we can overwrite the built-in instances. The unit tests can pass in mock client/configuration for testing purpose.

eddynaka: also, since we are creating the compilerdatalogger for each object, we dont need to keep creating an instance of the telemetry.

with that, once it is initialized, it won't create again

shaopeng-gh (approved):
:shipit:

eddynaka (approved):

PR #494, opened 2021-09-30
Remove Microsoft.DiaSymReader and replace the code with existing library

shaopeng-gh:


src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs
```diff
@@ -62,7 +62,7 @@ private void AnalyzeManagedAssemblyAndPdb(BinaryAnalyzerContext context)
             PEBinary target = context.PEBinary();
             Pdb di = target.Pdb;
 
-            if (target.PE.ManagedPdbSourceFileChecksumAlgorithm(di.FileType) != ChecksumAlgorithmType.Sha256)
+            if (target.PE.ManagedPdbSourceFileChecksumAlgorithm(di.FileType, di) != ChecksumAlgorithmType.Sha256)
```
eddynaka, line 65: >ManagedPdbSourceFileChecksumAlgorithm [](http://example.com/codeflow?start=26&length=37)

can you verify if we have enough coverage for the ba2004 rule?

shaopeng-gh: yes, we do have 14 test files for this rule with combination of:
full_exe_sha1.exe
full_sha1.dll
portable_exe_sha1.exe
portable_sha1.dll
netcoreapp3.1
netstandard2.0
Native

yongyan-gh:
>         private const string sha256 = "8829d00f-11b8-4213-878b-770e8597ac16";

can we use a unified definition of signature algorithms instead of definition in each place?

---
Refers to: src/BinaryParsers/PEBinary/PortableExecutable/PE.cs:35 in d233ce5. [](commit_id = d233ce56a313677810c2020722e61f4cf9fe75fa, deletion_comment = True)

shaopeng-gh:
>         private const string sha256 = "8829d00f-11b8-4213-878b-770e8597ac16";

yes, I prefer this as well.

---
In reply to: [931621533](https://github.com/microsoft/binskim/pull/494#issuecomment-931621533) [](http://example.com/codeflow?ancestors=931621533,931621533)

---
Refers to: src/BinaryParsers/PEBinary/PortableExecutable/PE.cs:35 in d233ce5. [](commit_id = d233ce56a313677810c2020722e61f4cf9fe75fa, deletion_comment = True)

eddynaka (approved):

PR #496, opened 2021-10-04
Add new lang code and add baseline file for Objective C

shaopeng-gh:
1.	The language 78 is set by NASM team on purpose, in their source code they indicate since there is no exact match for their language, they use “N” as language indicator and the ASCII is written as 78.
For this one I suggest we can add it and map 78 as “NASM”.

details in this file:
https://github.com/netwide-assembler/nasm/blob/5368e4579403daaf6c12165eb857893f8acfc7f8/output/codeview.c
the line of the code is:
section_write32(sect, 'N'); /* language: '**N**' (**0x4e**) for "**N**ASM"; flags are 0 */

2.	The MSDN documentation is outdated and missing new entries.
With the VS installed we can find the source code definition included in the DIA, which have a few more items added since.
“C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\DIA SDK\include\cvconst.h”
  CV_CFL_JSCRIPT  = 0x0E,
  CV_CFL_MSIL     = 0x0F,  // Unknown MSIL (LTCG of .NETMODULE)
  CV_CFL_HLSL     = 0x10,  // High Level Shader Language
  CV_CFL_OBJC     = 0x11,  // Objective-C
  CV_CFL_OBJCXX   = 0x12,  // Objective-C++
  CV_CFL_SWIFT    = 0x13,  // Swift
} CV_CFL_LANG;


src/BinaryParsers/PEBinary/ProgramDatabase/ObjectModuleDetails.cs
```diff
@@ -226,23 +226,27 @@ public string GetDialect(out string versionNumber)
...
+        VisualBasic = 0x0B,
+        ILASM = 0x0C,
+        Java = 0x0D,
+        JScript = 0x0E,
+        MSIL = 0x0F,
+        HLSL = 0x10, // High Level Shader Language
+        ObjectiveC = 0x11,
+        ObjectiveCxx = 0x12,
+        Swift = 0x13,
+        NASM = 0x4E, // The Netwide Assembler
```
shaopeng-gh, line 249: >NASM = 0x4E, // The Netwide Assembler [](http://example.com/codeflow?start=8&length=37)

added info where this come from in the desciprtion of the PR.

eddynaka (approved):
:shipit:

PR #499, opened 2021-10-06
Collecting repository/project/organization when available

eddynaka:


src/BinSkim.Driver/AnalysisSummaryExtractor.cs
```diff
@@ -51,17 +57,44 @@ public static void UpdateBuildPipelineInfo(AnalysisSummary summary)
...
+
+                summary.RepositoryId = ExtractValueFromEnvironmentVariable(RepositoryIdVar);
+                summary.RepositoryName = ExtractValueFromEnvironmentVariable(RepositoryNameVar);
+
+                summary.ProjectId = ExtractValueFromEnvironmentVariable(ProjectIdVar);
+                summary.ProjectName = ExtractValueFromEnvironmentVariable(ProjectNameVar);
+
+                summary.OrganizationId = ExtractValueFromEnvironmentVariable(OrganizationIdVar);
+                summary.OrganizationName = ExtractValueFromEnvironmentVariable(OrganizationNameVar);
+                summary.OrganizationName = summary.OrganizationName.Replace("https://dev.azure.com/", string.Empty).TrimEnd('/');
```
shaopeng-gh, line 72: >https://dev.azure.com/ [](http://example.com/codeflow?start=77&length=22)

question, will it possible with upper case

shaopeng-gh, line 72: >TrimEnd('/') [](http://example.com/codeflow?start=116&length=12)

Question, will the summary.OrganizationName only possible with
[https://dev.azure.com/A/](https://dev.azure.com/A/)
or also possible with
[https://dev.azure.com/A/B/](https://dev.azure.com/A/B/)

eddynaka: will only contain A/

shaopeng-gh: >summary.OrganizationName [](http://example.com/codeflow?start=43&length=24)

the logic in the method ExtractValueFromEnvironmentVariable() looks like if can not find will return null, so this can be null 
maybe add a "?" here

eddynaka: changed the method that is called to always output an empty string if no envvariable  was found.
that will prevent us to add `?` everywhere.

shaopeng-gh: good idea!

yongyan-gh (approved):
:shipit:

PR #512, opened 2021-11-02
Improving signing and packing process

eddynaka:
With this change we will:
- start signing all files that we ship: no need for intermediate folders.
- fix the package reference from `Binary.Parsers`


CreatePackagesFromLayoutDirectory.cmd
```diff
@@ -18,37 +18,13 @@ if "%Platform%" EQU "" (
 set Platform=x64
 )
 
-set BinaryOutputDirectory=%BinaryOutputDirectory%\%Platform%_%Configuration%\Publish
-set LayoutForSigningDirectory=%BinaryOutputDirectory%\..\LayoutForSigning
-
-:: Copy all multitargeted assemblies to their locations
-call :CopyExeForSigning BinSkim.exe
-call :CopyFilesForMultitargeting BinSkim.dll       || goto :ExitFailed
```
eddynaka, line 26: >CopyFilesForMultitargeting [](http://example.com/codeflow?start=6&length=26)

we don't need to move the files since we signed everything.

PR #515, opened 2021-11-06
Create one session/telemetry configuration per run

eddynaka:
# Description:

When we changed from single thread to multi-thread application, the reporting rule started to behave differently:
- many configurations were being created
- many sessions were being created
- hash wasn't showing correctly since the hash isn't calculated previously

# Fixes:

This PR will fix the following issues:
- session/telemetry objects should not be overridden if already exists
- session/telemetry objects should be created only once (even in a multithread environment)
- hash should have a reference to the context since it will be only created once we start loading a specific file

### DIFF
--- a/src/BinSkim.Sdk/CompilerDataLogger.cs
+++ b/src/BinSkim.Sdk/CompilerDataLogger.cs
@@ -274,7 +274,7 @@ namespace Microsoft.CodeAnalysis.IL.Sdk
             }
             else
             {
-                Console.WriteLine(summary.ToString());
+                Console.WriteLine(summary);
             }
         }
 
--- a/src/BinSkim.Sdk/CompilerDataLogger.cs
+++ b/src/BinSkim.Sdk/CompilerDataLogger.cs
@@ -20,14 +20,18 @@ namespace Microsoft.CodeAnalysis.IL.Sdk
         private const string CommandLineEventName = "CommandLineInformation";
         private const string AssemblyReferencesEventName = "AssemblyReferencesInformation";
 
-        private readonly int ChunkSize;
-        private readonly string sha256;
-        private readonly string relativeFilePath;
+        private string Sha256 => context?.Hashes?.Sha256;
 
-        private static string s_sessionId;
+        private readonly int chunkSize;
+        private readonly string relativeFilePath;
+        private readonly IAnalysisContext context;
+
+        internal static string s_sessionId;
         private static bool s_printHeader = true;
         private static TelemetryClient s_telemetryClient;
         private static TelemetryConfiguration s_telemetryConfiguration;
+
+        private static readonly object s_syncRoot = new object();
 
         public static bool TelemetryEnabled => s_telemetryClient != null;
 
@@ -37,10 +41,10 @@ namespace Microsoft.CodeAnalysis.IL.Sdk
                                   TelemetryClient telemetryClient = null,
                                   int chunkSize = 8192)
         {
-            s_telemetryConfiguration = telemetryConfiguration;
-            s_telemetryClient = telemetryClient;
-            s_sessionId = TelemetryEnabled ? Guid.NewGuid().ToString() : null;
-            ChunkSize = chunkSize;
+            s_telemetryConfiguration ??= telemetryConfiguration;
+            s_telemetryClient ??= telemetryClient;
+            s_sessionId ??= Guid.NewGuid().ToString();
+            this.chunkSize = chunkSize;
 
             if (!TelemetryEnabled)
             {
@@ -58,7 +62,7 @@ namespace Microsoft.CodeAnalysis.IL.Sdk
                 }
             }
 
-            this.sha256 = analysisContext?.Hashes?.Sha256 ?? string.Empty;
+            this.context = analysisContext;
             this.relativeFilePath = analysisContext?.TargetUri?.LocalPath ?? string.Empty;
 
             foreach (string path in targetFileSpecifiers)
@@ -77,9 +81,15 @@ namespace Microsoft.CodeAnalysis.IL.Sdk
         {
             if (s_telemetryConfiguration == null && s_telemetryClient == null)
             {
-                s_sessionId = Guid.NewGuid().ToString();
-                s_telemetryConfiguration = new TelemetryConfiguration(instrumentationKey);
-                s_telemetryClient = new TelemetryClient(s_telemetryConfiguration);
+                lock (s_syncRoot)
+                {
+                    if (s_telemetryConfiguration == null && s_telemetryClient == null)
+                    {
+                        s_sessionId = Guid.NewGuid().ToString();
+                        s_telemetryConfiguration = new TelemetryConfiguration(instrumentationKey);
+                        s_telemetryClient = new TelemetryClient(s_telemetryConfiguration);
+                    }
+                }
             }
         }
 
@@ -152,7 +162,7 @@ namespace Microsoft.CodeAnalysis.IL.Sdk
                     { "moduleName", compilerData.ModuleName ?? string.Empty },
                     { "moduleLibrary", (compilerData.ModuleName == compilerData.ModuleLibrary ? string.Empty : compilerData.ModuleLibrary ?? string.Empty) },
                     { "sessionId", s_sessionId },
-                    { "hash", this.sha256 },
+                    { "hash", this.Sha256 },
                     { "error", string.Empty }
                 };
 
@@ -183,7 +193,7 @@ namespace Microsoft.CodeAnalysis.IL.Sdk
             }
             else
             {
-                string log = $"{this.relativeFilePath},{compilerData},{this.sha256},";
+                string log = $"{this.relativeFilePath},{compilerData},{this.Sha256},";
                 Console.WriteLine(log);
             }
         }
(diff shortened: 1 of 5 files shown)


src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -20,15 +20,19 @@
         private const string CommandLineEventName = "CommandLineInformation";
         private const string AssemblyReferencesEventName = "AssemblyReferencesInformation";
 
-        private readonly int ChunkSize;
-        private readonly string sha256;
+        private string Sha256 => context?.Hashes?.Sha256;
```
eddynaka, line 23: >context?.Hashes?.Sha256 [](http://example.com/codeflow?start=33&length=23)

since we changed how hashes are generated, we should make a reference for the context so we will have the correct value of the hash when needed.

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -75,12 +79,18 @@
~
~        public static void Initialize(string instrumentationKey)
         {
             if (s_telemetryConfiguration == null && s_telemetryClient == null)
             {
-                s_sessionId = Guid.NewGuid().ToString();
-                s_telemetryConfiguration = new TelemetryConfiguration(instrumentationKey);
```
eddynaka, line 81: >s_syncRoot [](http://example.com/codeflow?start=22&length=10)

we should synchronize the telemetry creation to create one telemetry and one session

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -37,10 +41,10 @@
                                   TelemetryClient telemetryClient = null,
                                   int chunkSize = 8192)
         {
-            s_telemetryConfiguration = telemetryConfiguration;
-            s_telemetryClient = telemetryClient;
-            s_sessionId = TelemetryEnabled ? Guid.NewGuid().ToString() : null;
-            ChunkSize = chunkSize;
+            s_telemetryConfiguration ??= telemetryConfiguration;
```
eddynaka, line 44: >telemetryConfiguration [](http://example.com/codeflow?start=41&length=22)

we should only use telemetryConfiguration/telemetryClient if the respective static value is null. This should only be used during tests.

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -20,15 +20,19 @@ public class CompilerDataLogger
         private const string CommandLineEventName = "CommandLineInformation";
         private const string AssemblyReferencesEventName = "AssemblyReferencesInformation";
 
-        private readonly int ChunkSize;
-        private readonly string sha256;
+        private string Sha256 => context?.Hashes?.Sha256;
```
michaelcfanning, line 23: >context [](http://example.com/codeflow?start=33&length=7)

the code previously seemed to return string.empty in the null case, is that still required or is it ok to drop it?

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -20,15 +20,19 @@ public class CompilerDataLogger
...
+
+        private readonly int chunkSize;
         private readonly string relativeFilePath;
+        private readonly IAnalysisContext context;
 
-        private static string s_sessionId;
+        internal static string s_sessionId;
```
michaelcfanning, line 29: >internal [](http://example.com/codeflow?start=8&length=8)

please pull aside all the internal static properties (that are overridden by tests)  and group them together. 

we do this to improve the readability of code, i.e., that accessibility is a clue to the maintainer there is a special external caller (our tests) that update these values.

make sense? you are future-proofing this code to help whoever comes along when those tests fail.

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -77,9 +81,15 @@ public static void Initialize(string instrumentationKey)
         {
             if (s_telemetryConfiguration == null && s_telemetryClient == null)
             {
-                s_sessionId = Guid.NewGuid().ToString();
-                s_telemetryConfiguration = new TelemetryConfiguration(instrumentationKey);
-                s_telemetryClient = new TelemetryClient(s_telemetryConfiguration);
+                lock (s_syncRoot)
+                {
+                    if (s_telemetryConfiguration == null && s_telemetryClient == null)
```
michaelcfanning, line 86: >s_telemetryClient [](http://example.com/codeflow?start=60&length=17)

I don't understand the visibility of things in your class, why is s_sessionId internal but s_telemetry not?

Why is it ok to allow s_sessionId to be overridden by tests but then overwritten again if these two values are null?

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -264,17 +274,24 @@ public static void Summarize(AnalysisSummary summary)
...
             else
             {
-                Console.WriteLine(summary.ToString());
+                Console.WriteLine(summary);
             }
         }
 
+        internal static void Reset()
+        {
+            s_sessionId = null;
```
michaelcfanning, line 283: >s_sessionId [](http://example.com/codeflow?start=12&length=11)

why don't you lock here?

src/BinSkim.Sdk/Properties/AssemblyInfo.cs
```diff
@@ -2,6 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Reflection;
+using System.Runtime.CompilerServices;
 
 [assembly: AssemblyTitle("BinSkim SDK")]
 [assembly: AssemblyDescription("BinSkim SDK and utilities for authoring analysis checks")]
+
+[assembly: InternalsVisibleTo("Test.UnitTests.BinSkim.Driver")]
```
michaelcfanning, line 10: >UnitTests [](http://example.com/codeflow?start=36&length=9)

you added this to the wrong file, it should be in Shared\AssemblyInfo.cs

src/BinSkim.Sdk/CompilerDataLogger.cs
eddynaka: when we send a null to AppInsights, it won't render the key value pair. If we try to print it in console, a null is showed as an empty string.

so, in this case, we don't need to create a default to empty

src/BinSkim.Sdk/Properties/AssemblyInfo.cs
eddynaka: fixed!
thanks!

src/BinSkim.Sdk/CompilerDataLogger.cs
eddynaka: Added some docs in this method. We don't need a lock because it will only be called during tests.
All the tests that requires this aren't running in parallel.

src/BinSkim.Sdk/CompilerDataLogger.cs
eddynaka: session/configuration/client can be injected using the constructor but the session is always defined when we have the telemetry enabled.

It's why we don't need to make the configuration/client internal.

src/BinSkim.Sdk/CompilerDataLogger.cs
eddynaka: Got it!
thanks for the feedback!

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -77,9 +81,15 @@ public static void Initialize(string instrumentationKey)
         {
             if (s_telemetryConfiguration == null && s_telemetryClient == null)
             {
-                s_sessionId = Guid.NewGuid().ToString();
-                s_telemetryConfiguration = new TelemetryConfiguration(instrumentationKey);
-                s_telemetryClient = new TelemetryClient(s_telemetryConfiguration);
+                lock (s_syncRoot)
```
eddynaka, line 84: when we changed from single thread to multi thread, we started to create multiple configurations/clients/sessions.

With that in mind, we must synchronize the threads to prevent this behavior.

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -20,14 +20,18 @@ public class CompilerDataLogger
...
+        private string Sha256 => context?.Hashes?.Sha256;
+
+        private readonly int chunkSize;
         private readonly string relativeFilePath;
+        private readonly IAnalysisContext context;
+
+        internal static string s_sessionId;
 
-        private static string s_sessionId;
         private static bool s_printHeader = true;
```
michaelcfanning, line 31: >private static bool s_printHeader = true; [](http://example.com/codeflow?start=8&length=41)

add private static ctor, all static initialization is explicit in a common place.

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -20,14 +20,18 @@ public class CompilerDataLogger
         private const string CommandLineEventName = "CommandLineInformation";
         private const string AssemblyReferencesEventName = "AssemblyReferencesInformation";
 
-        private readonly int ChunkSize;
-        private readonly string sha256;
+        private string Sha256 => context?.Hashes?.Sha256;
+
+        private readonly int chunkSize;
         private readonly string relativeFilePath;
```
michaelcfanning, line 26: >relativeFilePath [](http://example.com/codeflow?start=32&length=16)

just retrieve this from the context as you do hashes

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -37,10 +41,10 @@ public class CompilerDataLogger
                                   TelemetryClient telemetryClient = null,
                                   int chunkSize = 8192)
         {
-            s_telemetryConfiguration = telemetryConfiguration;
-            s_telemetryClient = telemetryClient;
-            s_sessionId = TelemetryEnabled ? Guid.NewGuid().ToString() : null;
-            ChunkSize = chunkSize;
+            s_telemetryConfiguration ??= telemetryConfiguration;
```
michaelcfanning, line 44: >s_telemetryConfiguration ??= telemetryConfiguration [](http://example.com/codeflow?start=12&length=51)

ensure both are null or non-null in the ctor.

PR #523, opened 2021-11-11
Releasing stable version v1.9.0

eddynaka:
This PR is making the proper changes for us to release a new version that contains all binaries signed + the fix about the session.


michaelcfanning (approved):
:shipit:

PR #527, opened 2021-12-10
Address ARM64 CETCOMPAT compatability

### LINKED ISSUE #508: "Control-flow Enforcement Technology (CET) Shadow Stack mitigation" , BA2025 should not be applied for arm64 binaries
```
C:\i\openvpn-pipeline-artifacts\arm64\openvpn.exe: warning BA2025: 'openvpn.exe' does not enable the Control-flow Enforcement Technology (CET) Shadow Stack mitigation. To resolve this issue, pass /CETCOMPAT on the linker command lines.
```

however, CET is not available for arm64

marmegh:
Fixes #508

Added detection of ARM64, two test cases, and a not applicable test for BA2025.


src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -802,6 +802,8 @@ public bool IsWixBinary
             }
         }
 
+        public bool IsARM64Binary => Machine == Machine.Arm64;
```
michaelcfanning, line 805: >IsARM64Binary [](http://example.com/codeflow?start=20&length=13)

Let's remove this, as we don't have any other pattern which defines a helper for machine types. Instead, the rule can simply access the Machine property directly on this class.

Context here is that this PE abstraction has a bunch of legacy helpers on it that haven't been cleanly designed, and that I'd hoped would wither away. :)

src/BinSkim.Rules/PERules/BA2025.EnableShadowStack.cs
```diff
@@ -60,6 +60,12 @@ public override AnalysisApplicability CanAnalyzePE(PEBinary target, Sarif.Proper
                 return notApplicable;
             }
 
+            if (portableExecutable.IsARM64Binary)
```
michaelcfanning, line 63: >portableExecutable.IsARM64Binary [](http://example.com/codeflow?start=16&length=32)

So, you'll change this to portableExecutable.Machine == Machine.Arm64...

src/BinSkim.Rules/PERules/BA2025.EnableShadowStack.cs
```diff
@@ -60,6 +60,12 @@ public override AnalysisApplicability CanAnalyzePE(PEBinary target, Sarif.Proper
...
+            {
+                reasonForNotAnalyzing = MetadataConditions.ImageIsARM64BitBinary;
```
michaelcfanning, line 65: >ImageIsARM64BitBinary [](http://example.com/codeflow?start=59&length=21)

Please update casing everywhere to 'Arm', this is the appropriate casing in API contexts. You will see a different casing in some of our file names, etc., but that's ok.

src/BinSkim.Sdk/MetadataConditions.cs
```diff
@@ -40,5 +40,6 @@ public static class MetadataConditions
         public static readonly string ImageIsPreVersion7WindowsCEBinary = SdkResources.MetadataCondition_ImageIsPreVersion7WindowsCEBinary;
         public static readonly string ImageIsNativeUniversalWindowsPlatformBinary = SdkResources.MetadataCondition_ImageIsNativeUniversalWindowsPlatformBinary;
         public static readonly string MachOIsNotExecutableDynamicLibraryOrObject = SdkResources.MetadataCondition_MachOIsNotExecutableDynamicLibraryOrObject;
+        public static readonly string ImageIsARM64BitBinary = SdkResources.MetadataCondition_ImageIsARM64BitBinary;
```
michaelcfanning, line 43: >ImageIsARM64BitBinary [](http://example.com/codeflow?start=38&length=21)

Casing of this field is 'ImageIsArm64BitBinary.

Also, you may observe that all these fields are sorted by line length, so please move this line to the appropriate spot (somewhere near 'ElfIsCoreNoneOrObject', I think).

src/BinSkim.Sdk/SdkResources.resx
```diff
@@ -240,4 +240,7 @@
   <data name="MetadataCondition_ImageIsNotMachO" xml:space="preserve">
     <value>image is not a MachO binary</value>
   </data>
+  <data name="MetadataCondition_ImageIsARM64BitBinary" xml:space="preserve">
+    <value>image is an ARM 64 bit binary</value>
```
michaelcfanning, line 244: >ARM 64 [](http://example.com/codeflow?start=23&length=6)

ok, so this is a user-facing string and not a .NET API, so we prefer the casing that customers see on the web when querying around for this kind of thing. I think 'ARM64' is best, what do you think? Notice you've introduced a space in your rendering, 'ARM 64', which I propose to eliminate.

src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
```diff
@@ -277,13 +278,23 @@ private BinaryAnalyzerContext CreateContext(TestMessageLogger logger, Properties
                 context.Rule = skimmer;
 
                 AnalysisApplicability applicability = skimmer.CanAnalyze(context, out string reasonForNotAnalyzing);
+
                 if (applicability != expectedApplicability)
                 {
                     sb.AppendLine("CanAnalyze did not correct indicate target applicability (unexpected return was " +
```
michaelcfanning, line 284: >correct [](http://example.com/codeflow?start=54&length=7)

s/be 'correctly'. You didn't make this mistake but would you mind correcting it?

src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
```diff
@@ -277,13 +278,23 @@ private BinaryAnalyzerContext CreateContext(TestMessageLogger logger, Properties
...
                         applicability + "): " +
                         Path.GetFileName(target));
                     continue;
                 }
+
+                if (expectedReasonForNotAnalyzing != null && reasonForNotAnalyzing != expectedReasonForNotAnalyzing)
```
michaelcfanning, line 290: >expectedReasonForNotAnalyzing != null && reasonForNotAnalyzing != expectedReasonForNotAnalyzing [](http://example.com/codeflow?start=20&length=95)

why do we require the null check here??

src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
```diff
@@ -277,13 +278,23 @@ private BinaryAnalyzerContext CreateContext(TestMessageLogger logger, Properties
...
+                {
+                    sb.AppendLine("Cannot Analyze but unexpected reason identified (unexpected return was " +
+                        reasonForNotAnalyzing + "but " +
+                        expectedReasonForNotAnalyzing + "was expected): " +
```
michaelcfanning, line 294: >was [](http://example.com/codeflow?start=57&length=3)

Looks like you are missing a space at the beginning of this literal. This raises a very common, more advanced piece of guidance I give all engineers: you should always provoke and examine the output of your test failures to find bugs like this and ensure that your new testing is providing a good experience in failure conditions (i.e., actionable for the poor soul who is responsible for fixing the future break).

It's great that you've added this code path by the way!!

src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
```diff
@@ -277,13 +278,23 @@ private BinaryAnalyzerContext CreateContext(TestMessageLogger logger, Properties
...
                 {
                     sb.AppendLine("CanAnalyze did not correct indicate target applicability (unexpected return was " +
                         applicability + "): " +
                         Path.GetFileName(target));
                     continue;
                 }
+
+                if (expectedReasonForNotAnalyzing != null && reasonForNotAnalyzing != expectedReasonForNotAnalyzing)
+                {
+                    sb.AppendLine("Cannot Analyze but unexpected reason identified (unexpected return was " +
```
michaelcfanning, line 292: >Cannot [](http://example.com/codeflow?start=35&length=6)

So, this code follows the current convention of using the '+' to concatenate. You could convert to using interpolated strings instead. If you do, maybe clean up the rest of the file for interpolated string usage? e.g.:

Notice some other small clean-ups: 1) fixed casing of 'Analyze', 2) enquoted the dynamic injected values in apostrophes. This is very helpful in test output and I will offer this suggestion repeatedly. i.e., (We expected 'foo' but saw 'bar') is more readable than (We expected foo but saw bar)

                    sb.AppendLine(
                        $"Cannot analyze but unexpected reason identified (unexpected return was " +
                        $"'{reasonForNotAnalyzing}' but " +
                        $"'{expectedReasonForNotAnalyzing}' was expected): " +
                        Path.GetFileName(target));

michaelcfanning:
Solid change! Well done. You'll be looking in good shape after addressing my trivial suggestions. :)

src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
marmegh: Not all tests that use this method take advantage of the additional functionality yet. We definitely want them to; but that should be part of a separate PR. We may encounter scenarios where the originally intended reasons for the test passing are no longer true; for that reason I did not pursue updating all tests yet.

src/BinSkim.Sdk/SdkResources.resx
marmegh: Is "MetadataCondition_ImageIsARM64BitBinary" good as is? Or does it need to be updated to MetadataCondition_ImageIsArm64BitBinary? I believe I found all of the others.

src/BinSkim.Rules/PERules/BA2025.EnableShadowStack.cs
michaelcfanning: Nice! btw - if you feel like these issues are resolved, please feel free to mark them that way.

src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
```diff
@@ -131,12 +131,22 @@ public RuleTests(ITestOutputHelper output)
...
+                            expectedText,
+                            Path.GetFileName(target)));
                 }
                 else
                 {
                     other.Remove(target);
-                    sb.AppendLine("Expected '" + ruleName + "' " + expectedText + " but check " + actualText + " for: " + Path.GetFileName(target));
+                    sb.AppendLine(
+                        string.Format(
+                            "Expected '{0}' {1} but check {2} for: {3}",
```
michaelcfanning, line 145: >Expected '{0}' {1} but check {2} for: {3}", [](http://example.com/codeflow?start=29&length=43)

This format string looks odd to me, what is this error condition actually saying? Do you have an example we could review?

src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
```diff
@@ -1099,6 +1126,15 @@ public void BA2025_EnableShadowStack_Fail()
             }
         }
 
+        [Fact]
+        public void BA2025_EnableShadowStack_NotApplicable()
+        {
+            this.VerifyApplicability(
+                skimmer: new EnableShadowStack(),
+                applicabilityConditions: new HashSet<string>(),
```
michaelcfanning, line 1134: >new HashSet<string>() [](http://example.com/codeflow?start=41&length=21)

In general, I think requiring that a caller instantiate an object which is empty indicates a design that could be improved, i.e., passing 'null' here may be a preferable indicator of no applicability conditions. 

Make sense? This is a fairly subtle API consideration, not a big deal. Sometimes it is important to preserve null for some specific scenario, as distinct from a non-null object which is empty.

src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
marmegh: "Expected 'DoNotModifyStackProtectionCookie' failure but check succeeded for: __security_cookies.exe"

I'm not entirely certain why we have this at all. As written, even if we have a regression in the code we shouldn't hit this condition.

src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
```diff
@@ -277,11 +288,27 @@ private BinaryAnalyzerContext CreateContext(TestMessageLogger logger, Properties
...
                 AnalysisApplicability applicability = skimmer.CanAnalyze(context, out string reasonForNotAnalyzing);
+
                 if (applicability != expectedApplicability)
                 {
-                    sb.AppendLine("CanAnalyze did not correct indicate target applicability (unexpected return was " +
-                        applicability + "): " +
-                        Path.GetFileName(target));
+                    sb.AppendLine(
+                        string.Format(
+                            "Can analyze did not correctly indicate target applicability (unexpected return was '{0}'): '{1}'",
```
eddynaka, line 296: >Can analyze [](http://example.com/codeflow?start=29&length=11)

Here, it should be 'CanAnalyze', since this is the name of the method.

eddynaka (approved):
:shipit:

src/BinSkim.Sdk/MetadataConditions.cs
```diff
@@ -22,6 +22,7 @@ public static class MetadataConditions
         public static readonly string ImageIsNot32BitBinary = SdkResources.MetadataCondition_ImageIsNot32BitBinary;
         public static readonly string ImageIsNot64BitBinary = SdkResources.MetadataCondition_ImageIsNot64BitBinary;
         public static readonly string ElfIsCoreNoneOrObject = SdkResources.MetadataCondition_ElfIsCoreNoneOrObject;
+        public static readonly string ImageIsArm64BitBinary = SdkResources.MetadataCondition_ImageIsARM64BitBinary;
```
michaelcfanning, line 25: >MetadataCondition_ImageIsARM64BitBinary [](http://example.com/codeflow?start=75&length=39)

Casing of this is still wrong, s/be MetadataCondition_ImageIsArm64BitBinary. Sorry to nickle-and-dime you, generally we follow the .NET guidelines in all source code expressed stuff (like resx string names). Actual user-facing strings can retain capitalization exceptions if it's part of the public brand.

Note instances of 'Elf' and 'Wix' in this file, where a user-facing string might prefer 'ELF' or 'WIX'.

src/BinSkim.Sdk/SdkResources.resx
michaelcfanning: Yes, please update as you suggest. Also, we should probably use 'ARM64 binary' or 'ARM 64-bit binary' in the user-facing string, as each is a bit more readable. I think the latter is good, what do you think?

src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
michaelcfanning: btw - as you resolve things, you can mark them as 'resolved' in code flow. Then your reviewers can 'close' them. :)

src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
michaelcfanning: AFAIK, I think we should! Try moving a test file from the 'Failed' directory to the 'Pass' directory in your enlistment and we should hit this code path.

D:\src\binskim\src\Test.FunctionalTests.BinSkim.Rules\FunctionalTestsData\BA2012.DoNotModifyStackProtectionCookie

michaelcfanning: By the way, if you'd really like to improve this code path, what I'd do is to leave an example as a source code comment.

// Generates message such as the following:
// Expected 'DoNotModifyStackProtectionCookie' failure but check succeeded for: __security_cookies.exe"

Please note that you appear to have added apostrophes where they aren't needed, for 'failure' on line 135 and line 145. Ideally, grab current state of these formatted strings from the debugger to make sure they are good (and once you confirm, put an example in the source code comment, then someone won't enquote something in apostrophes when they shouldn't (I led us here by making this suggestion, of course, without noting that not all dynamic args might be good to enquote).

Also, for extra points, please add the rule id to the output, as this is super helpful in locating the test binarys, the directories for which are prefixed with this info, e.g., :

// Expected 'BA20212:DoNotModifyStackProtectionCookie' failure but check succeeded for: __security_cookies.exe"

michaelcfanning (approved):
:shipit:

michaelcfanning:
Well done Mary (and Eddy, thanks for your help on it), this change is now in a perfect condition. :)

---
In reply to: [830692580](https://github.com/microsoft/binskim/pull/527#issuecomment-830692580) [](http://example.com/codeflow?ancestors=830692580,830692580)

PR #528, opened 2021-12-11
Update README.md

KennieNP:
Added paragraph on how to get the executable file from the nuget package

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,13 @@ This repository contains the source code for BinSkim, a Portable Executable (PE)
 2. Read the **[User Guide](./docs/UserGuide.md)**
 3. Find out more about the Static Analysis Results Interchange Format **([SARIF](https://github.com/sarif-standard/sarif-spec/))** used to output Binskim results
 
+### How to extract the exe file from the nuget package
+If you only want to run the Binskim tool without installing anything, then you can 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,6 +19,13 @@ This repository contains the source code for BinSkim, a Portable Executable (PE)
 2. Read the **[User Guide](./docs/UserGuide.md)**
 3. Find out more about the Static Analysis Results Interchange Format **([SARIF](https://github.com/sarif-standard/sarif-spec/))** used to output Binskim results
 
+### How to extract the exe file from the nuget package
```
michaelcfanning, line 22: > How to extract the exe file from the nuget package [](http://example.com/codeflow?start=3&length=51)

We should package BinSkim as a .NET tool, then this would all turn into a one-liner:

dotnet tool install Microsoft.CodeAnalysis.BinSkim --tool-path <PATH>

michaelcfanning (approved):
:shipit:

michaelcfanning:
Thanks for the contribution!!

README.md
eddynaka: We should start a conversation with the guardian team to see if they will be able to handle a dotnet tool. I will search in my emails, I think I did start that thread in the past.

PR #531, opened 2021-12-16
Fixing `NullReferenceException` when `--Hashes` and telemetry rules are enabled

eddynaka:
With the single thread analysis, the argument `--Hashes` was translated to `--insert Hashes`, adding the deprecated message.
When we moved to the multithreaded analysis, without that condition we would not generate the artifacts.
Since the summary was expecting the artifacts, it was accessing the number of elements directly.


src/BinSkim.Driver/AnalyzeCommand.cs
```diff
@@ -135,30 +146,5 @@ public override int Run(AnalyzeOptions analyzeOptions)
...
-                ToolVersion = tool.Driver.Version,
-                NormalizedPath = string.Join(";", options.TargetFileSpecifiers.Select(p => System.IO.Path.GetDirectoryName(p)).Distinct()),
-                SymbolPath = options.SymbolsPath,
-                FileAnalyzed = artifacts.Count,
-                // FileNotAnalyzed =
-                StartTimeUtc = invocation.StartTimeUtc,
-                EndTimeUtc = invocation.EndTimeUtc,
-                TimeConsumed = invocation.EndTimeUtc - invocation.StartTimeUtc,
-            };
-        }
```
eddynaka, line 162: this code was moved to the AnalysisSummaryExtractor in previous PRs.

src/BinSkim.Driver/AnalysisSummaryExtractor.cs
```diff
@@ -40,7 +40,7 @@ public static AnalysisSummary ExtractAnalysisSummary(SarifLog sarifLog, AnalyzeO
                 ToolVersion = tool.Driver.Version,
                 NormalizedPath = string.Join(";", options.TargetFileSpecifiers.Select(p => System.IO.Path.GetDirectoryName(p)).Distinct()),
                 SymbolPath = options.SymbolsPath,
-                FileAnalyzed = artifacts.Count,
+                FileAnalyzed = artifacts?.Count ?? 0,
```
eddynaka, line 43: >artifacts?.Count ?? 0 [](http://example.com/codeflow?start=31&length=21)

this will prevent the exception in case the SARIF does not have artifacts.

This would happen if: you don't have the --insert hashes or --Hashes option enabled but you have the telemetry rules.

src/BinSkim.Driver/AnalyzeCommand.cs
```diff
@@ -85,6 +85,17 @@ public override int Run(AnalyzeOptions analyzeOptions)
                 analyzeOptions.Kind = new List<ResultKind> { ResultKind.Fail, ResultKind.NotApplicable, ResultKind.Pass };
             }
 
+#pragma warning disable CS0618 // Type or member is obsolete
+            if (analyzeOptions.ComputeFileHashes)
```
eddynaka, line 89: >analyzeOptions.ComputeFileHashes [](http://example.com/codeflow?start=16&length=32)

this change is needed to prevent a breaking change in the guardian tasks.
We need to support --Hashes

src/BinSkim.Driver/AnalyzeOptions.cs
```diff
@@ -48,5 +48,10 @@ public class AnalyzeOptions : AnalyzeOptionsBase
             "ignorePdbLoadError",
             HelpText = "If enabled, BinSkim won't break if we have a 'PdbLoadingException'.")]
         public bool IgnorePdbLoadError { get; set; }
+
+        [Option(
+            "binskimAppInsightsKey",
+            HelpText = "An ApplicationInsightsKey to which will be send data. If not available, it will read from the environment variable.")]
+        public string BinskimAppInsightsKey { get; set; }
```
eddynaka, line 55: >BinskimAppInsightsKey [](http://example.com/codeflow?start=22&length=21)

To facilitate the guardian integration, we will expose a new parameter and also be able to read the environment variable.

The reason: from azure pipelines, creating the variable it isn't passing the value correctly (the variable is always unavailable).

michaelcfanning: We can't accept a change that encourages users to explicitly provide sensitive data on the command-line, this is a very poor security practice, as command-lines are persisted/logged in many contexts.

eddynaka: thanks for the feedback.
i will remove the change and keep it clean to fix the issue that i faced.

eddynaka: Just updated and sent an email with the details.
thanks for the help!

src/BinSkim.Driver/AnalyzeCommand.cs
shaopeng-gh: maybe we should make this change in SDK instead?
AnalyzeCommandBase has this check,
MultithreadedAnalyzeCommandBase does not,

if the change is made in Binskim then only Binskim is fixed but not SDK

eddynaka: I was wondering about that.
Since we want to deprecate --hashes, i would move that logic to AnalyzeCommand and remove the reference from AnalyzeCommandBase.

With that, in the futhre, the options would not have --hashes by default and binskim itself would implement it.

marmegh (approved):
:shipit:

src/BinSkim.Driver/AnalysisSummaryExtractor.cs
```diff
@@ -40,7 +40,7 @@ public static AnalysisSummary ExtractAnalysisSummary(SarifLog sarifLog, AnalyzeO
...
                 // FileNotAnalyzed =
```
michaelcfanning, line 44: >// FileNotAnalyzed = [](http://example.com/codeflow?start=16&length=20)

can you remove this please? it isn't helpful to have it here. 

please add a TODO, or just open an issue

michaelcfanning (approved):
:shipit:

src/BinSkim.Driver/AnalysisSummaryExtractor.cs
eddynaka: created a work item for that.

PR #533, opened 2021-12-16
Fix: Error BA2006: '...' was compiled with one or more modules which were not built using minimum required tool versions

shaopeng-gh:
# Description:

reported issue with 
error BA2006: '...' was compiled with one or more modules which were not built using minimum required tool versions 
this error happens in new version 1.9.0
tested it passes in old version 1.7.2

# Fix:

related to changeset https://github.com/microsoft/binskim/pull/344
location: file BA2006.BuildWithSecureTools.cs line 104

we should add a skip if the compiler is not the ones we currently support those we can check min version.
After we widen the support we should remove this skip.


src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -116,6 +116,13 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
                 Symbol om = omView.Value;
                 ObjectModuleDetails omDetails = om.GetObjectModuleDetails();
 
+                if (omDetails.WellKnownCompiler != WellKnownCompilers.MicrosoftC
+                    && omDetails.WellKnownCompiler != WellKnownCompilers.MicrosoftCxx)
+                {
+                    // TODO: https://github.com/Microsoft/binskim/issues/114
+                    continue;
+                }
```
shaopeng-gh, line 124: In changeset
[https://github.com/microsoft/binskim/pull/344](https://github.com/microsoft/binskim/pull/344)
We have this check and skip in BA2011, I think we should have it for BA2006 as well.

BA2006 had this and were removed:
if (omDetails.WellKnownCompiler != WellKnownCompilers.MicrosoftNativeCompiler)
{
 continue;
}

The reason is currently the only know compiler in the enum WellKnownCompilers are Microsoft compilers, and the version to compare are only Microsoft VS versions. We should remove this check after we also have the codes and determined min versions for all other compilers.

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/Binskim.linux-x64.dll.sarif
```diff
@@ -772,13 +772,10 @@
           "rules": [
             {
               "id": "BA2001",
+              "name": "LoadImageAboveFourGigabyteAddress",
```
shaopeng-gh, line 775: >"name": "LoadImageAboveFourGigabyteAddress", [](http://example.com/codeflow?start=14&length=44)

all json node ordering change are not caused by this change, we did not run the baseline for sometime, 
running re-baseline from current main will also get this change.

shaopeng-gh:
>           "kind": "pass",

skip those re-ordering changes,
this is a real change for this pr, 

clang version 13 is latest version should be supported, after this fix now it will not compare {13.0.0.0} to VS Version(17, 0, 65501, 17013) and report it.
(skipped if not MS compiler)


---
Refers to: src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/clangcl.pe.cpp.codeview.exe.sarif:431 in d845a11. [](commit_id = d845a118c4be0b2083d1961bcbdba767aeae2582, deletion_comment = False)

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -116,6 +116,13 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
                 Symbol om = omView.Value;
                 ObjectModuleDetails omDetails = om.GetObjectModuleDetails();
 
+                if (omDetails.WellKnownCompiler != WellKnownCompilers.MicrosoftC
+                    && omDetails.WellKnownCompiler != WellKnownCompilers.MicrosoftCxx)
+                {
+                    // TODO: https://github.com/Microsoft/binskim/issues/114
```
michaelcfanning, line 122: >TODO [](http://example.com/codeflow?start=23&length=4)

TODO: MikeFan (1/6/2022)
We need to take a step back and comprehensively review our compiler/language support.
[https://github.com/Microsoft/binskim/issues/114](https://github.com/Microsoft/binskim/issues/114)

michaelcfanning:
>                     //case Language.MASM:

Comment this whole block, using C

/* 
repeat your TODO information here.

case Language.MASM

*/

---
Refers to: src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs:142 in af6ff95. [](commit_id = af6ff95e94c45281e81f6b3f5783108534ab9024, deletion_comment = False)

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/Corrupted_Native_x86_VS2013_Default.exe.sarif
```diff
@@ -50,6 +50,7 @@
           }
         }
       ],
+      "automationDetails": {},
```
michaelcfanning, line 53: This is valid SARIF, automationDetails has no required members, *but* it's inefficient, we don't need to serialize an empty object. In general, we don't want our log file persistence to change version-over-version.

shaopeng-gh:
>                     //case Language.MASM:

fixed

---
In reply to: [1006978150](https://github.com/microsoft/binskim/pull/533#issuecomment-1006978150) [](http://example.com/codeflow?ancestors=1006978150,1006978150)

---
Refers to: src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs:142 in af6ff95. [](commit_id = af6ff95e94c45281e81f6b3f5783108534ab9024, deletion_comment = False)

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
shaopeng-gh: fixed

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/Binskim.linux-x64.dll.sarif
shaopeng-gh: thanks for the feedback, working on
1. update SDK to have those re-ordering reverted.
2. review
3. change only one ordering so that name is right after id.

shaopeng-gh: I have created a fix in SDK
[https://github.com/microsoft/sarif-sdk/pull/2420](https://github.com/microsoft/sarif-sdk/pull/2420) (please review together)
and then point binskim to that branch and run re-baseline to verify the fix.
Now this PR is clearer to review.

After all looks good to you, we should get the desired ordering from you and I will update both PR with that order.(you mentioned one name field)

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/Corrupted_Native_x86_VS2013_Default.exe.sarif
shaopeng-gh: fixed together in the SDK PR, link in the other comment with json node ordering.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -132,33 +141,43 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
...
+                    Language data is not always included if it is only compiled with SymTagCompiland without SymTagCompilandDetails
+                    https://docs.microsoft.com/en-us/visualstudio/debugger/debug-interface-access/compilanddetails?view=vs-2022
+                    Compiland information is split between symbols with a SymTagCompiland tag (low detail)
+                    and a SymTagCompilandDetails tag (high detail).
                     case Language.Unknown:
                     {
                         minCompilerVersion =
                             context.Policy.GetProperty(MinimumToolVersions)[nameof(Language.Unknown)];
                         break;
                     }
```
marmegh, line 179: This one wasn't commented out in previous iterations. Was this change intentional?

marmegh:
Is there supposed to be a sarif file for the clangcl.pe.c.codeview.exe in the expected folder?

shaopeng-gh:
good question, I think this was the similar thing that confused Michael in the review :)
I added test files in the Functional tests folder, and there is a "modified" file in the expected folder.
Because if add new test file in the Baseline folder we are expecting "new" file not "modified" file.

the cpp one was already in the baseline, I added it also to the Functional folder because I think it benefit that if future regression we might see a very clear test failure with the rule id.
I also draft an additional exe for c to cover c.

related info:
1.if you put exe in the Baseline folder and run .\UpdateBaselines.ps1 it will update the .sarif file in the expected folder.
2. for exe in Functional folder it is used for loop all files and check if pass or fail depend on the folder name it is put in. No .sarif file need to be generated.


---
In reply to: [1007837347](https://github.com/microsoft/binskim/pull/533#issuecomment-1007837347) [](http://example.com/codeflow?ancestors=1007837347,1007837347)

src/ReleaseHistory.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* BUGFIX: Fix Error BA2006: '...' was compiled with one or more modules which were not built using minimum required tool versions [533](https://github.com/microsoft/binskim/pull/533)
```
marmegh, line 5: >* BUGFIX: Fix Error BA2006: '...' was compiled with one or more modules which were not built using minimum required tool versions  [](http://example.com/codeflow?start=0&length=130)

This seems a little too generic. Can we specify which instances this is being fixed for?

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -116,6 +116,15 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
                 Symbol om = omView.Value;
                 ObjectModuleDetails omDetails = om.GetObjectModuleDetails();
 
+                if (omDetails.WellKnownCompiler != WellKnownCompilers.MicrosoftC
+                    && omDetails.WellKnownCompiler != WellKnownCompilers.MicrosoftCxx)
```
eddynaka, line 120: should this be:
```
omDetails.WellKnownCompiler != WellKnownCompilers.MicrosoftC &&
omDetails.WellKnownCompiler != WellKnownCompilers.MicrosoftCxx
```

eddynaka:
Just remembering that we must review the sarif-sdk change before reviewing this. Because we cannot merge this without merging the other.

---
In reply to: [1007848170](https://github.com/microsoft/binskim/pull/533#issuecomment-1007848170) [](http://example.com/codeflow?ancestors=1007848170,1007837347)

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
shaopeng-gh: fixed

eddynaka:
Re-open the PR and point the submodule to main.
it won't have any change in the testing since u were already pointing to ur branch

---
In reply to: [1007979314](https://github.com/microsoft/binskim/pull/533#issuecomment-1007979314) [](http://example.com/codeflow?ancestors=1007979314,1007837347)

PR #543, opened 2022-01-06
FluentAssertions test code minor syntax change

shaopeng-gh:
# Description:

FluentAssertions auto upgrade by dependabot, 
new version has a minor change that does not allow to use Should().Equals()
it need to use methods in the class like:
Should().Be
Should().BeEquivalentTo

# Fixes:

We only have 1 test method using it, changed to use Be() which is a exact match, base on the code logic.


eddynaka (approved):
:shipit:

PR #545, opened 2022-01-10
BUGFIX: Fix incorrect analysis for non-Microsoft compiler on BA2006.BuildWithSecureTools

shaopeng-gh:
# Note:

this is the same issue with https://github.com/microsoft/binskim/pull/533,
The only difference is after we merged in SDK changes, this is now pointing to SDK main instead of private branch.
And then re-run the baseline. no new files changed.
And changed the history.md file with this new PR number "545"

# Description:

reported issue with 
error BA2006: '...' was compiled with one or more modules which were not built using minimum required tool versions 
this error happens in new version 1.9.0
tested it passes in old version 1.7.2

# Fix:

related to changeset https://github.com/microsoft/binskim/pull/344
location: file BA2006.BuildWithSecureTools.cs line 104

we should add a skip if the compiler is not the ones we currently support those we can check min version.
After we widen the support we should remove this skip.


src/ReleaseHistory.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* BUGFIX: Fix Error BA2006: '...' was compiled with one or more modules which were not built using minimum required tool versions [545](https://github.com/microsoft/binskim/pull/545)
```
eddynaka, line 5: >Fix Error BA2006: '...' was compiled with one or more modules which were not built using minimum required tool versions [](http://example.com/codeflow?start=10&length=119)

We are fixing the incorrect analysis for compilers that aren't Microsoft.

Maybe something like: "Fix incorrect analysis for non-Microsoft compiler on BA2006.RuleName."

shaopeng-gh: agree, this is both short and clear.

michaelcfanning (approved):
:shipit:

PR #548, opened 2022-01-19
Support analyzing binaries build with Clang when command line was recorded

shaopeng-gh:
# Description:

We used to have understanding that GCC will include command line and parameter in the debug info, 
like what we have with the Microsoft pdb files. And Clang does not.

Now with more in depth look into each documents, that is just the "default" settings, using GCC can turn it off and using Clang can turn it on it is just one parameter.

# Fix:

now we understand that GCC can turn it off, we can not assume it will always have the command line and parameter in the debug info like the code we currently have, and if Clang includes it we can also analyze it.

so now for the rules depends on examine command line, we check if command line and parameter are included,
instead of checking GCC or Clang.


shaopeng-gh:
> CodeFlow for GitHub does not download executable files. If you trust this file and would like to run it, please download it yourself through GitHub.

explain for the test files,
both gcc and clang I generated recent versions.
gcc 7-11
clang 10-13

we have 2 rules, one for "stackclashprotection" and one for "stackprotetor"
as in my test file name, "stackprotetor" is enabled. "stackclashprotection" is not.

so these files will be in stackprotetor test either pass or not applicable,
and in stackclashprotection test either fail or not applicable

not applicable when commandline is not recorded. or it is very old gcc pre-8 we explicitly not supporting.


---
Refers to: src/Test.FunctionalTests.BinSkim.Rules/FunctionalTestsData/BA3003.EnableStackProtector/NotApplicable/clang.linux.10.stackprotector.recordcommandline_default:1 in 7fe5907. [](commit_id = 7fe5907dbf7f2bf616681500ddfcb1197ca469f6, deletion_comment = False)

src/BinaryParsers/ElfBinary/Dwarf/DwarfCompileCommandLineInfo.cs
```diff
@@ -18,10 +18,12 @@ public class DwarfCompileCommandLineInfo
         public string CommandLine { get; set; }
         public DwarfLanguage Language { get; set; }
         public string CompileDirectory { get; set; }
+        public bool ParametersInCluded { get; set; } = false;
```
yongyan-gh, line 21: >ParametersInCluded [](http://example.com/codeflow?start=20&length=18)

ParametersIncluded

shaopeng-gh: fixed, thanks

src/BinaryParsers/ElfBinary/Dwarf/DwarfCompileCommandLineInfo.cs
```diff
@@ -18,10 +18,12 @@ public class DwarfCompileCommandLineInfo
         public string CommandLine { get; set; }
         public DwarfLanguage Language { get; set; }
         public string CompileDirectory { get; set; }
+        public bool ParametersIncluded { get; set; } = false;
+        public DwarfCommandLineType DwarfCommandLineType { get; set; } = DwarfCommandLineType.Unknown;
```
eddynaka, line 22: >DwarfCommandLineType [](http://example.com/codeflow?start=36&length=20)

rename this to CommandLineType, similar to what we used in DwarfLanguage -> Language.

eddynaka:
>         private CanAnalyzeDwarfResult VerifyDwarfBinary(IDwarfBinary binary)

the logic is very similar to the one we have in the enablestackclashprotection. can you make a helper and use it in both places?

---
Refers to: src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs:168 in 210feb2. [](commit_id = 210feb2e8eef2d14a2da95f5fe270c4852f39cb2, deletion_comment = False)

src/ReleaseHistory.md
```diff
@@ -1,5 +1,9 @@
 # BinSkim Release History
 
+## Unreleased
+
+* FEATURE: support analyzing binaries build with Clang when command line was recorded. [#548](https://github.com/microsoft/binskim/pull/548)
```
marmegh, line 5: nit: Capitalize `Support` to match the rest of the file.

marmegh, line 5: >* FEATURE: support analyzing binaries build with Clang when command line was recorded. [](http://example.com/codeflow?start=0&length=86)

It looks like there's also a change to which sarif-sdk is referenced. Can we get documented here and in the description the nature of that update?

src/BinSkim.Sdk/SdkResources.resx
```diff
@@ -243,4 +243,7 @@
   <data name="MetadataCondition_ImageIsArm64BitBinary" xml:space="preserve">
     <value>image is an ARM64 binary</value>
   </data>
+  <data name="MetadataCondition_ImageBuiltWithoutRecordCommandLine" xml:space="preserve">
+    <value>image is built without record command line</value>
```
yongyan-gh, line 247: >is built [](http://example.com/codeflow?start=17&length=8)

'image was compiled without {record command line}'
is it possible to have specific argument here (e.g. -grecord-gcc-switches) for each type of binary?

src/ReleaseHistory.md
shaopeng-gh: great the dependabot lag updating the sdk sub module,
I just see it got updated in main today,
so I have merged from latest main, this change is gone now.

shaopeng-gh: agree, fixed

shaopeng-gh:
>         private CanAnalyzeDwarfResult VerifyDwarfBinary(IDwarfBinary binary)

great input, there are 2 methods the same, i moved both to a file, same structure as how we reuse CanAnalyze() code for
BA2011.EnableStackProtection.cs
BA2012.DoNotModifyStackProtectionCookie.cs

---
In reply to: [1018617012](https://github.com/microsoft/binskim/pull/548#issuecomment-1018617012) [](http://example.com/codeflow?ancestors=1018617012,1018617012)

---
Refers to: src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs:168 in 210feb2. [](commit_id = 210feb2e8eef2d14a2da95f5fe270c4852f39cb2, deletion_comment = False)

src/BinaryParsers/ElfBinary/Dwarf/DwarfCompileCommandLineInfo.cs
shaopeng-gh: agree

src/BinSkim.Sdk/SdkResources.resx
shaopeng-gh: fixed the first part, for the second part a binary can have many modules with different compiler so can not single one, will keep it as it is

shaopeng-gh:
**Parked since many issues and questions,
Note down for later reference.**

Instead of currently saying "not applicable", we want mark rule as failed if "command line not available" , apply to both macho, FLF (Shaopeng: also PE).
(Shaopeng: also have concern this will result in many fails, and if user able to resolve 3rd party libs)

Need a new rule.

Need learn form the result of a large set of files.

Need more understanding of language and command line vs. binary, compilation unit, symbol
e.g. if 2 symbol in a compilation unit can have different language

Need complicated test binaries that has many compilers, compilation units, symbols, our current test files are very basic.

Verify resulting user facing strings, output is actionable and user know how to fix by looking at it

"clang version"
 1. verify with other Linux distribution, e.g. Debian
 2. verify macho clang files

Since we want to change to fail the rule instead of not applicable, most code should be in Analyze() instead of Can Analyze()

various refactoring/naming issues:
GetDwarfCommandLineType(string commandline) -->
GetCompilerType(string commandline)
CommandLineUtilities.cs --> should not in rules folder, move to root of binary parser
static bool analyze(IDwarfBinary binary, out List<DwarfCompileCommandLineInfo> failedList)
 --> don't put static method inside method, move outside
DwarfCompileCommandLineInfo --> sth like DwarfCompileSummary, tbd
public List<DwarfCompileCommandLineInfo> CommandLineInfos { get; }
--> public IList<DwarfCompileCommandLineInfo> CommandLineInfos { get; }
DwarfCommandLineType --> sth like Compiler, tbd

PR #530, opened 2021-12-15
Add new PE CV_CFL_LANG language code for ALIASOBJ and Rust

shaopeng-gh:
# Description:

Add new PE CV_CFL_LANG language code for ALIASOBJ and RUST.

    L"Objective-C",                    // CV_CFL_OBJC
    L"Objective-C++",                  // CV_CFL_OBJCXX
    L"Swift",                          // CV_CFL_SWIFT
    L"ALIASOBJ",                       // CV_CFL_ALIASOBJ
    L"Rust",                           // CV_CFL_RUST

PR #551, opened 2022-01-26
BUGFIX: Fix `BA2014.DoNotDisableStackProtectionForFunctions` to eliminate false positive reports that `GsDriverEntry` has disabled the stack protector.

shaopeng-gh:
# Description:

User reported issue by email:

> There’s some issues with functions that are in the AllowedDeclspecSafebuffersFunctionNames but not in ApprovedFunctionsThatDisableStackProtection. My only remaining issues with Binskim are these 4FPs (due to overcomplicated logic inside Binskim source code where left hand does not know what the right hand was coding): image
> 
> Could you please modify the source code for BinSkim such that ApprovedFunctionsThatDisableStackProtection include the AllowedDeclspecSafebuffersFunctionNames? ‘Allowed’ list probably contains the functions necessary to bootstrap GS support for drivers therefore they have to also be ‘approved’.
> 
> Thanks.

# Fix:

My understanding of the issue is Gs initialization functions should be also allowed in BA2014.
Fixed and added test file.


shaopeng-gh:
>               "userfn_use_safebuffers_2;userfn_use_safebuffers_1"

explain for the 4 test files:
case of: 
1. GsDriverEntry that is in the allow list
2. user functions not in the allow list (I added 2 functions)
3. GsDriverEntry  + user functions
4. None

---
Refers to: src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/Native_x64_VS2019_CPlusPlus_GsDriverEntry_And_UserFnUseSafeBuffers.exe.sarif:669 in 761f609. [](commit_id = 761f609a6f2844c4920adf429745ca144ae92bb8, deletion_comment = False)

src/ReleaseHistory.md
```diff
@@ -1,5 +1,9 @@
 # BinSkim Release History
 
+## Unreleased
```
shaopeng-gh, line 3: >Unreleased [](http://example.com/codeflow?start=3&length=10)

One thing need to be done is to add the build script for the test files I added, 
talked to Mary we suggested a location in this PR:
[https://github.com/microsoft/binskim/pull/530](https://github.com/microsoft/binskim/pull/530)

Once we settle the location I will add in this PR at the same location and format.

shaopeng-gh: done, added at the same place with same format.

docs/FunctionalTestBuildScripts.md
```diff
@@ -0,0 +1,29 @@
+# Build Scripts
+
+This file records scripts used to compile the test files, in alphabetical order.
+Base scenario is a simple hello world program built with different parameters for testing purpose.
+Test files are located in [BaselineTestsData](https://github.com/microsoft/binskim/tree/main/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData) and [FunctionalTestsData](https://github.com/microsoft/binskim/tree/main/src/Test.FunctionalTests.BinSkim.Rules/FunctionalTestsData).
+
+## Native_x64_VS2019_CPlusPlus_GsDriverEntry_And_UserFnUseSafeBuffers.exe
```
shaopeng-gh, line 7: >Native_x64_VS2019_CPlusPlus_GsDriverEntry_And_UserFnUseSafeBuffers.exe [](http://example.com/codeflow?start=3&length=70)

explain for the 4 test files:

case of: 
1. only GsDriverEntry that is in the allow list  ---- result should be pass
2. only user functions not in the allow list (I added 2 functions)  ---- result should be fail
3. GsDriverEntry  + user functions  ---- result should be fail
4. None  ---- result should be pass

src/BinSkim.Rules/PERules/BA2014.DoNotDisableStackProtectionForFunctions.cs
```diff
@@ -59,6 +59,9 @@ private static StringSet BuildApprovedFunctionsStringSet()
                 "_TlgWrite",
                 "__vcrt_trace_logging_provider::_TlgWrite"
             };
+
+            // Also allow Gs initialization functions.
+            result.UnionWith(StackProtectionUtilities.GSInitializationFunctionNames);
```
marmegh, line 64: >result.UnionWith(StackProtectionUtilities.GSInitializationFunctionNames); [](http://example.com/codeflow?start=12&length=73)

Is there any way we could get unit test(s) to cover this change?

marmegh:
Release History? Since this is going to be changing an aspect of the results, we probably need a release history entry to describe any changes in behavior.

shaopeng-gh:
yes we do have.

---
In reply to: [1057456130](https://github.com/microsoft/binskim/pull/551#issuecomment-1057456130) [](http://example.com/codeflow?ancestors=1057456130,1057456130)

src/BinSkim.Rules/PERules/BA2014.DoNotDisableStackProtectionForFunctions.cs
shaopeng-gh, line 64: >StackProtectionUtilities.GSInitializationFunctionNames [](http://example.com/codeflow?start=29&length=54)

tried can not put this in to StringSet(){}

src/BinSkim.Rules/PERules/BA2014.DoNotDisableStackProtectionForFunctions.cs
```diff
@@ -59,6 +59,9 @@ private static StringSet BuildApprovedFunctionsStringSet()
                 "_TlgWrite",
                 "__vcrt_trace_logging_provider::_TlgWrite"
             };
+
+            // Also allow Gs initialization functions.
```
shaopeng-gh, line 63: >// Also allow Gs initialization functions. [](http://example.com/codeflow?start=12&length=42)

removed the comment.

src/BinSkim.Rules/PERules/BA2014.DoNotDisableStackProtectionForFunctions.cs
shaopeng-gh: added

shaopeng-gh:
updated the release history entry per input.

---
In reply to: [1057467235](https://github.com/microsoft/binskim/pull/551#issuecomment-1057467235) [](http://example.com/codeflow?ancestors=1057467235,1057456130,1057456130)

src/BinSkim.Rules/PERules/BA2014.DoNotDisableStackProtectionForFunctions.cs
michaelcfanning: Oops, of course! Because it is a collection, sorry for the dumb advice. :)

michaelcfanning (approved):
:shipit:

PR #555, opened 2022-02-01
Fix `MultithreadedAnalyzeCommandBase` artifacts generation and enforcing JSON properties ordering

eddynaka:
# Description
The previous version of the sarif-sdk had a bug that was generating an artifact for each file that was analyzed. This update will guarantee that we will only generate an artifact if we have a result.

The sarif-sdk submodule change is also enforcing a specific ordering of the `name`/`helpuri` in the `reportingDescriptor`. This is expected since we want the `name` to be near the `id`.

# Test
Checked all rebaselined files and only found two `unexpected` changes related to the artifacts. The reason we have those is that we are not producing any results, so the artifacts are getting dropped from the SARIF log.

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -1,5 +1,9 @@
 # BinSkim Release History
 
+## Unreleased
+
+* BUGFIX: Fix `MultithreadedAnalyzeCommandBase` artifacts generation and enforcing JSON properties ordering. [#555](https://github.com/microsoft/binskim/pull/555)
+
 ## **v1.9.1** [NuGet Package](https://www.nuget.org/packages/Microsoft.CodeAnalysis.BinSkim/1.9.1)
 
 * BUGFIX: Fix incorrect analysis for non-Microsoft compiler on BA2006.BuildWithSecureTools. [545](https://github.com/microsoft/binskim/pull/545)
--- a/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/Binskim.linux-x64.dll.sarif
+++ b/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/Binskim.linux-x64.dll.sarif
@@ -772,10 +772,10 @@
           "rules": [
             {
               "id": "BA2001",
+              "name": "LoadImageAboveFourGigabyteAddress",
               "fullDescription": {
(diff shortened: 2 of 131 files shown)


src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/Corrupted_Native_x86_VS2013_Default.exe.sarif
```diff
@@ -38,18 +38,6 @@
           "executionSuccessful": false
         }
       ],
-      "artifacts": [
```
eddynaka, line 41: >artifacts [](http://example.com/codeflow?start=7&length=9)

We will only emit an artifact if we have a result.

eddynaka:
>       ],

we will only emit an artifact if we have a result. This was part of the sarif-sdk change to improve the file size generated when utilizing the MultithreadedAnalyzeCommandBase class:
[https://github.com/microsoft/sarif-sdk/pull/2433](https://github.com/microsoft/sarif-sdk/pull/2433)

---
Refers to: src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/gcc.stack_protector.a.sarif:52 in 64f53be. [](commit_id = 64f53be44cd6359a001acd2fae7ea3bd20b16c3b, deletion_comment = True)

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/Binskim.linux-x64.dll.sarif
```diff
@@ -772,10 +772,10 @@
           "rules": [
             {
               "id": "BA2001",
+              "name": "LoadImageAboveFourGigabyteAddress",
```
eddynaka, line 775: >name [](http://example.com/codeflow?start=15&length=4)

with the sarif-sdk change, we are enforcing a specific order where name will be close to id and helpuri will be near the end.

Reviewed all files and only two has `unexpected` changes that are commented as well.

src/ReleaseHistory.md
```diff
@@ -1,5 +1,9 @@
 # BinSkim Release History
 
+## Unreleased
+
+* BUGFIX: Fix `MultithreadedAnalyzeCommandBase` artifacts generation and enforcing JSON properties ordering. [#555](https://github.com/microsoft/binskim/pull/555)
```
michaelcfanning, line 5: >MultithreadedAnalyzeCommandBase [](http://example.com/codeflow?start=15&length=31)

I'd document this isn't a BinSkim fix, it's an SDK update. 

Update SARIF-SDK in order to fix `blah blah

michaelcfanning:
>       "artifacts": [

this is a bug, it shouldn't have gone away. Instead, the change should be that we see the addition of the artifact index.

---
Refers to: src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/clang.stack_protector.a.sarif:41 in 3fa6d69. [](commit_id = 3fa6d6951d8014d45256f3aeab2787798728cf0e, deletion_comment = True)

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/Corrupted_Native_x86_VS2013_Default.exe.sarif
michaelcfanning: There should be an artifact entry for every artifact that is explicitly referenced. There's not a good reason to arbitrarily restrict this to results. We've lost important information here.

michaelcfanning (approved):
:shipit:

PR #558, opened 2022-02-02
Improving common path function

eddynaka:
# Description

In this change, I'm improving how the method `ReturnCommonPathRootFromTargetSpecifiersIfOneExists` generates the common path. The method will now generate the common path from the target specifiers if that exists.

The current behavior will generate the following:
```
c:\path1\ C:\path2\ -> no common path
c:\path1\ C:\path1\path2\ -> no common path
c:\path1\ C:\path1\path2\1.exe -> no common path
```

The new behavior will generate the following:
```
c:\path1\ c:\path2\ -> C:\
c:\path1\ c:\path1\path2\ -> c:\path1\
c:\path1\ c:\path1\path2\1.exe -> c:\path1\
```

# Tests

The following tests were added:
- Shuffling the array to guarantee that won't matter the order of the input
- UNC examples
- Paths with `..`
- Paths with duplicated information
- Paths that point to files and folders
- Paths with distinct cases

### DIFF
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -89,11 +89,22 @@ namespace Microsoft.CodeAnalysis.IL
                 }
             }
 
-            return smallestPath.Length == 0 || smallestPath[^1] != '\\'
-                ? string.Empty
-                : smallestPath.EndsWith(@"\\")
-                    ? smallestPath.Substring(0, smallestPath.Length - 1)
-                    : smallestPath;
+            if (smallestPath.Length == 0)
+            {
+                return string.Empty;
+            }
+
+            if (smallestPath[^1] != '\\')
+            {
+                // In this case, smallestPath is equal to 'c:\path1\partial-path' (this is an incomplete path).
+                // Once we execute, our smallestPath will be transformed into 'c:\path1\'.
+                string[] parts = smallestPath.Split('\\');
+                smallestPath = string.Join(@"\", parts[..^1]) + @"\";
+            }
+
+            return smallestPath.EndsWith(@"\\")
+                ? smallestPath.Substring(0, smallestPath.Length - 1)
+                : smallestPath;
         }
 
         public override int Run(AnalyzeOptions analyzeOptions)
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -76,7 +76,7 @@ namespace Microsoft.CodeAnalysis.IL
 
             var fileSpecifierDirectories = new SortedSet<string>(StringComparer.OrdinalIgnoreCase);
 
-            // Normalizing all 'targetFileSpecifiers', guaranteeing that they will always end with only one slash.
+            // Normalizing all 'targetFileSpecifiers', ensuring that they will always end with only one slash.
             foreach (string targetFileSpecifier in targetFileSpecifiers)
             {
                 string targetFileDirectory = Path.GetDirectoryName(Path.GetFullPath(targetFileSpecifier));
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -71,13 +71,29 @@ namespace Microsoft.CodeAnalysis.IL
 
         internal static string ReturnCommonPathRootFromTargetSpecifiersIfOneExists(IEnumerable<string> targetFileSpecifiers)
         {
+            var fileSpecifierDirectories = new SortedSet<string>(targetFileSpecifiers.Select(s => (Path.GetDirectoryName(Path.GetFullPath(s)) + @"\")),
+                                                                 StringComparer.OrdinalIgnoreCase);
 
-            var fileSpecifierDirectories = new HashSet<string>(targetFileSpecifiers.Select(s => Path.GetDirectoryName(Path.GetFullPath(s)) + @"\"),
-                                                               StringComparer.OrdinalIgnoreCase);
+            string smallestPath = fileSpecifierDirectories.First();
+            fileSpecifierDirectories.Remove(smallestPath);
 
-            return fileSpecifierDirectories.Count == 1
-                ? fileSpecifierDirectories.First()
-                : string.Empty;
+            for (int i = 0; i < smallestPath.Length; i++)
+            {
+                foreach (string specifier in fileSpecifierDirectories)
+                {
+                    if (char.ToLowerInvariant(smallestPath[i]) != char.ToLowerInvariant(specifier[i]))
+                    {
+                        smallestPath = smallestPath.Substring(0, i);
+                        break;
+                    }
+                }
+            }
+
+            return smallestPath.Length == 0 || smallestPath[^1] != '\\'
+                ? string.Empty
+                : smallestPath.EndsWith(@"\\")
+                    ? smallestPath.Substring(0, smallestPath.Length - 1)
+                    : smallestPath;
         }
 
         public override int Run(AnalyzeOptions analyzeOptions)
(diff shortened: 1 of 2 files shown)


src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -71,13 +71,40 @@ protected override void InitializeConfiguration(AnalyzeOptions options, BinaryAn
...
+            {
+                return string.Empty;
+            }
+
+            if (smallestPath[^1] != '\\')
+            {
+                // In this case, smallestPath is equal to 'c:\path1\partial-path' (this is an incomplete path).
+                // Once we execute, our smallestPath will be transformed into 'c:\path1\'.
+                string[] parts = smallestPath.Split('\\');
+                smallestPath = string.Join(@"\", parts[..^1]) + @"\";
```
eddynaka, line 102: >..^1 [](http://example.com/codeflow?start=55&length=4)

this is a new syntax where it will get all elements besides the last.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -71,13 +71,40 @@ protected override void InitializeConfiguration(AnalyzeOptions options, BinaryAn
...
+            }
+
+            return smallestPath.EndsWith(@"\\")
```
eddynaka, line 105: >@"\\" [](http://example.com/codeflow?start=41&length=5)

when we have a path that is:
C:\*.dll, we are removing *.dll and adding a new \, becoming C:\\

So, this code is intended to remove the double slash at the end

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -71,13 +71,47 @@
...
+            string smallestPath = fileSpecifierDirectories.First();
+            fileSpecifierDirectories.Remove(smallestPath);
 
-            return fileSpecifierDirectories.Count == 1
-                ? fileSpecifierDirectories.First()
-                : string.Empty;
+            for (int i = 0; i < smallestPath.Length; i++)
+            {
+                // We don't need to iterate all characters of the smallestPath if we don't have anything
+                // to compare against.
```
michaelcfanning, line 83: >ToLowerInvariant [](http://example.com/codeflow?start=29&length=16)

are you subject to Turkish i issues here?

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -71,13 +71,47 @@
...
+                foreach (string specifier in fileSpecifierDirectories)
+                {
+                    if (char.ToLowerInvariant(smallestPath[i]) != char.ToLowerInvariant(specifier[i]))
+                    {
+                        smallestPath = smallestPath.Substring(0, i);
+                        break;
+                    }
+                }
+            }
+
```
michaelcfanning, line 98: >smallestPath [](http://example.com/codeflow?start=33&length=12)

'smallestPath'

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -71,13 +71,47 @@
...
+            if (smallestPath.Length == 0)
+            {
```
michaelcfanning, line 100: >smallestPath [](http://example.com/codeflow?start=33&length=12)

Can you just use LastIndexOf on the slash?

eddynaka: yes, we can.
re-worked the code a bit to improve that.

thanks for the suggestion!

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
eddynaka: just updated the comments.

thanks again for the suggestions! :)

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
eddynaka: Since I'm calling chart.ToLowerInvariant, this wont face the `Turkish i issue`. Thanks for pointing that out, I didn't know about that issue.

eddynaka:
Right now, the method `ReturnCommonPathRootFromTargetSpecifiersIfOneExists` is an internal method in the BinSkim MulltithreadedAnalyzeCommand and we could move it to the sarif-sdk, so other AnalyzeCommands could also use or even SarifLogger.

eddynaka:
Also, no release notes were provided, since this is a new function that is not exposed to customer and was not published in any BinSkim version.

---
In reply to: [1028279270](https://github.com/microsoft/binskim/pull/558#issuecomment-1028279270) [](http://example.com/codeflow?ancestors=1028279270,1028279270)

src/Test.UnitTests.BinSkim.Driver/MultithreadedAnalyzeCommandTests.cs
```diff
@@ -74,9 +116,31 @@ public void MultithreadedAnalyzeCommand_ReturnCommonPathRootFromTargetSpecifiers
...
+                    sb.AppendLine($"The test was expecting '{testCase.ExpectedCommonPath}' but found '{commonPath}' when shuffed.");
+                }
             }
 
             sb.Length.Should().Be(0, sb.ToString());
         }
+
+        private static string[] Shuffle(string[] data)
+        {
+            var rand = new Random();
```
michaelcfanning, line 134: >Random [](http://example.com/codeflow?start=27&length=6)

Don't ever introduce randomization into tests without first providing a generated seed to the random class. Next you need to report the seed that was used.

Otherwise you may not easily be able to reproduce a rarely generated error!

michaelcfanning: Go look at how this is done in security utilities.

[https://github.com/microsoft/security-utilities/pull/15/files#diff-8e7029bcfb4b2255f03585b66339672c7bb45d04e8835b4200b82f42740d0460](https://github.com/microsoft/security-utilities/pull/15/files#diff-8e7029bcfb4b2255f03585b66339672c7bb45d04e8835b4200b82f42740d0460)

eddynaka: fixed!
thanks for that and added that to my notes :)

src/Test.UnitTests.BinSkim.Driver/MultithreadedAnalyzeCommandTests.cs
```diff
@@ -74,9 +125,30 @@ public void MultithreadedAnalyzeCommand_ReturnCommonPathRootFromTargetSpecifiers
                 {
                     sb.AppendLine($"The test was expecting '{testCase.ExpectedCommonPath}' but found '{commonPath}'.");
                 }
+
+                // Testing the same string array in a random order.
+                // This will guarantee that the sorting is working as expected.
+                commonPath = MultithreadedAnalyzeCommand.ReturnCommonPathRootFromTargetSpecifiersIfOneExists(Shuffle(testCase.TargetFileSpecifiers));
+                if (commonPath != testCase.ExpectedCommonPath)
+                {
+                    sb.AppendLine($"The test was expecting '{testCase.ExpectedCommonPath}' but found '{commonPath}' when shuffed with seed '{s_randomSeed}'.");
```
michaelcfanning, line 134: >shuffed [](http://example.com/codeflow?start=121&length=7)

s/be 'shuffled'.

michaelcfanning (approved):
:shipit:

src/Test.UnitTests.BinSkim.Driver/MultithreadedAnalyzeCommandTests.cs
michaelcfanning: you are approved to merge whenever you like. Nice work here!

eddynaka: as always, thank you for the guidance!

PR #552, opened 2022-01-26
Improve compiler telemetry to better fit tool architecture.

michaelcfanning:

PR #559, opened 2022-02-02
Rule `BA4001.ReportPortableExecutableCompilerData` will now analyze all PE binaries

eddynaka:
# Description

This change will enable us to analyze any PE file.

We discovered this issue in this [PR](https://github.com/microsoft/binskim/pull/552) where we saw valid PEs being skipped for being a managed IL library.

Previously, the rule `BA4001.ReportPortableExecutableCompilerData` was inheriting from `WindowsBinaryAndPdbSkimmerBase`. Looking at the `CanAnalyze`, we can see a few cases where a binary might be rejected (generate the result `NotApplicable`): `IsWixBinary`, `IsILLibrary`, or `IsDotNetCoreBootstrapExe`.

Now, I'm changing its implementation to inherit from `PEBinarySkimmerBase`, which will just check if we have a Portable Executable or not: https://github.com/microsoft/binskim/blob/2ae90725e27667593d6071b74e53609f31130b85/src/BinSkim.Rules/PERules/PEBinarySkimmerBase.cs#L9-L24

### DIFF
--- a/src/BinSkim.Rules/PERules/BA4001.ReportPortableExecutableCompilerData.cs
+++ b/src/BinSkim.Rules/PERules/BA4001.ReportPortableExecutableCompilerData.cs
@@ -15,15 +15,13 @@ using Microsoft.CodeAnalysis.Sarif.Driver;
 namespace Microsoft.CodeAnalysis.IL.Rules
 {
     [Export(typeof(Skimmer<BinaryAnalyzerContext>)), Export(typeof(ReportingDescriptor)), Export(typeof(IOptionsProvider))]
-    public class ReportPECompilerData : WindowsBinaryAndPdbSkimmerBase, IOptionsProvider
+    public class ReportPortableExecutableCompilerData : PEBinarySkimmerBase, IOptionsProvider
     {
         /// <summary>
         /// BA4001. This reporting rule writes compiler data to AppInsights and 
         /// a CSV file (if configured) for every compilation unit that's scanned.
         /// </summary>
         public override string Id => RuleIds.ReportPortableExecutableCompilerData;
-
-        public override bool LogPdbLoadException => false;
 
         /// <summary>
         /// This rule emits CSV data to the console for every compiler/language/version
@@ -47,7 +45,7 @@ namespace Microsoft.CodeAnalysis.IL.Rules
             return AnalysisApplicability.ApplicableToSpecifiedTarget;
         }
 
-        public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext context)
+        public override void Analyze(BinaryAnalyzerContext context)
         {
             if (!context.CompilerDataLogger.Enabled)
             {
--- a/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/BinSkim.win-x64.ni.dll.sarif
+++ b/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/BinSkim.win-x64.ni.dll.sarif
@@ -557,32 +557,8 @@
           ]
         },
         {
-          "ruleId": "BA4001",
-          "ruleIndex": 23,
-          "kind": "notApplicable",
-          "level": "none",
-          "message": {
-            "id": "NotApplicable_InvalidMetadata",
-            "arguments": [
-              "BinSkim.win-x64.ni.dll",
-              "ReportPECompilerData",
-              "image is a managed IL library (i.e., ahead of time compiled) assembly"
-            ]
-          },
-          "locations": [
-            {
-              "physicalLocation": {
-                "artifactLocation": {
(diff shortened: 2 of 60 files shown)


src/BinSkim.Rules/PERules/BA4001.ReportPortableExecutableCompilerData.cs
```diff
@@ -15,16 +15,14 @@
...
-    public class ReportPECompilerData : WindowsBinaryAndPdbSkimmerBase, IOptionsProvider
+    public class ReportPortableExecutableCompilerData : PEBinarySkimmerBase, IOptionsProvider
     {
         /// <summary>
         /// BA4001. This reporting rule writes compiler data to AppInsights and 
         /// a CSV file (if configured) for every compilation unit that's scanned.
         /// </summary>
         public override string Id => RuleIds.ReportPortableExecutableCompilerData;
 
-        public override bool LogPdbLoadException => false;
```
eddynaka, line 26: >LogPdbLoadException [](http://example.com/codeflow?start=29&length=19)

the PEBinarySkimmerBase does not log anything in case we don't find a PDB. Also, this property is only available for the WindowsBinaryAndPdbSkimmerBase.

eddynaka: Also, with this change, we will try to analyze any PE available. If we find the PDB, great! If not, it will just skip.

src/BinSkim.Rules/PERules/BA4001.ReportPortableExecutableCompilerData.cs
```diff
@@ -15,16 +15,14 @@
 namespace Microsoft.CodeAnalysis.IL.Rules
 {
     [Export(typeof(Skimmer<BinaryAnalyzerContext>)), Export(typeof(ReportingDescriptor)), Export(typeof(IOptionsProvider))]
-    public class ReportPECompilerData : WindowsBinaryAndPdbSkimmerBase, IOptionsProvider
+    public class ReportPortableExecutableCompilerData : PEBinarySkimmerBase, IOptionsProvider
```
eddynaka, line 18: >ReportPortableExecutableCompilerData [](http://example.com/codeflow?start=17&length=36)

I saw that  the file name didn't match with the class name. So, I did this renaming.

This will update all SARIF files, showing a renaming.

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/macho.clang-lib.fat.o.sarif
```diff
@@ -830,10 +830,10 @@
...
+              "name": "LoadImageAboveFourGigabyteAddress",
               "fullDescription": {
                 "text": "64-bit images should have a preferred base address above the 4GB boundary to prevent triggering an Address Space Layout Randomization (ASLR) compatibility mode that decreases security. ASLR compatibility mode reduces the number of locations to which ASLR may relocate the binary, reducing its effectiveness at mitigating memory corruption vulnerabilities. To resolve this issue, either use the default preferred base address by removing any uses of /baseaddress from compiler command lines, or /BASE from linker command lines (recommended), or configure your program to start at a base address above 4GB when compiled for 64 bit platforms (by changing the constant passed to /baseaddress or /BASE). Note that if you choose to continue using a custom preferred base address, you will need to make this modification only for 64-bit builds, as base addresses above 4GB are not valid for 32-bit binaries."
               },
-              "helpUri": "https://github.com/microsoft/binskim/blob/main/docs/BinSkimRules.md#rule-BA2001LoadImageAboveFourGigabyteAddress",
```
eddynaka, line 836: >helpUri [](http://example.com/codeflow?start=15&length=7)

the three mach-o files are not being rebased correctly.
Since i changed the rulename, i need to manually rebase.

we should investigate why the rebaselining isn't working for mach-o files.

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/BinSkim.win-x64.ni.dll.sarif
```diff
@@ -556,33 +556,9 @@
...
-                "artifactLocation": {
-                  "uri": "file:///Z:/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/BinSkim.win-x64.ni.dll",
-                  "index": 0
-                }
-              }
-            }
-          ]
-        },
         {
           "ruleId": "BA5001",
```
marmegh, line 560: >"ruleId": "BA5001", [](http://example.com/codeflow?start=10&length=19)

Is this supposed to be changing from `BA4001` to `BA5001`? Seeing this in a few different sarif files.

PR #563, opened 2022-02-11
BUGFIX: Fix missing item in MinimumToolVersionsMap when user only overwrite part of the settings

shaopeng-gh:
# Description:

User reported issue by email:

new issue with Binskim 1.9.2
```
2022-02-10T15:55:59.5540580Z [Standard]   D:\a\_work\1\a\OpenSSH-Win64\sftp.exe : error ERR998.ExceptionInAnalyze : BA2006 : An exception of type 'KeyNotFoundException' was raised analyzing 'sftp.exe' for check 'BuildWithSecureTools' (which has been disabled). The exception may have resulted from a problem related to parsing the analysis target, and not specific to the rule, however.
2022-02-10T15:55:59.5542301Z [Standard]   KeyNotFoundException: The given key 'C' was not present in the dictionary.
2022-02-10T15:55:59.5543351Z [Standard]      at System.Collections.Concurrent.ConcurrentDictionary`2.ThrowKeyNotFoundException(Object key)
2022-02-10T15:55:59.5544773Z [Standard]      at Microsoft.CodeAnalysis.IL.Rules.BuildWithSecureTools.AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext context) in D:\a\r1\a\src\BinSkim.Rules\PERules\BA2006.BuildWithSecureTools.cs:line 86
2022-02-10T15:55:59.5546453Z [Standard]      at Microsoft.CodeAnalysis.IL.Rules.WindowsBinaryAndPdbSkimmerBase.Analyze(BinaryAnalyzerContext context) in D:\a\r1\a\src\BinSkim.Rules\PERules\WindowsBinaryAndPdbSkimmerBase.cs:line 63
2022-02-10T15:55:59.5548328Z [Standard]      at Microsoft.CodeAnalysis.Sarif.Driver.MultithreadedAnalyzeCommandBase`2.AnalyzeTargetHelper(TContext context, IEnumerable`1 skimmers, ISet`1 disabledSkimmers)
```

# Cause:

The issue is user has a config file and in config file:

```
<Properties Key="BA2006.BuildWithSecureTools.Options" Type="PropertiesDictionary">
    <Property Key="AdvancedMitigationsEnforced" Value="None" Type="AdvancedMitigations" />
    <Properties Key="AllowedLibraries" Type="IL.Rules.StringToVersionMap" />
    <Properties Key="MinimumToolVersions" Type="IL.Rules.StringToVersionMap">
      <Property Key="MinimumCompilerVersion" Value="17.0.65501.17013" Type="System.Version" />
      <Property Key="MinimumXboxCompilerVersion" Value="16.0.11886.0" Type="System.Version" />
    </Properties>
    <Property Key="RuleEnabled" Value="Default" Type="Driver.RuleEnabledState" />
  </Properties>
```

The MinimumToolVersions contains only the ones user want to overwrite, but it does not contain all possible settings, current code when read it, will not combine with the other settings that are not overwritten in the config file so we missed them.
e.g
[nameof(Language.C)] = new Version(17, 0, 65501, 17013),
[nameof(Language.Cxx)] = new Version(17, 0, 65501, 17013),
[nameof(Language.Unknown)] = new Version(int.MaxValue, int.MaxValue, int.MaxValue, int.MaxValue),

# Fix:

propose to change the code to make sure the other settings that are not overwritten in the config file, are also added to the dictionary.

### DIFF
--- a/src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
+++ b/src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
@@ -66,6 +66,29 @@ namespace Microsoft.CodeAnalysis.IL.Rules
             new PerLanguageOption<AdvancedMitigations>(
                 AnalyzerName, nameof(AdvancedMitigationsEnforced), defaultValue: () => AdvancedMitigations.None);
 
+        public override void Initialize(BinaryAnalyzerContext context)
+        {
+            if (context.Policy == null) { return; }
+
+            StringToVersionMap currentMinimumToolVersions = context.Policy.GetProperty(MinimumToolVersions, cacheDefault: false);
+            StringToVersionMap defaultMinimumToolVersions = BuildMinimumToolVersionsMap();
+
+            // If user only overwrite part of the settings, currentMinimumToolVersions will only have those settings.
+            // This is to make sure other default settings are applied as well.
+            foreach (KeyValuePair<string, Version> defaultMinimumToolVersion in defaultMinimumToolVersions)
+            {
+                currentMinimumToolVersions.TryAdd(defaultMinimumToolVersion.Key, defaultMinimumToolVersion.Value);
+            }
+
+            context.Policy.SetProperty(
+                MinimumToolVersions,
+                currentMinimumToolVersions,
+                cacheDescription: false,
+                persistToSettingsContainer: true);
+
+            return;
+        }
+
         public override AnalysisApplicability CanAnalyzePE(PEBinary target, PropertiesDictionary policy, out string reasonForNotAnalyzing)
         {
             PE portableExecutable = target.PE;
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -1,5 +1,9 @@
 # BinSkim Release History
 
+## Unreleased
+
+* BUGFIX: Fix ` KeyNotFoundException` raised on `BA2006.BuildWithSecureTools` when required properties are removed from the config file. [#565](https://github.com/microsoft/binskim/pull/565)
+
 ## **v1.9.2** [NuGet Package](https://www.nuget.org/packages/Microsoft.CodeAnalysis.BinSkim/1.9.2)
 
 * BUGFIX: Fix `MultithreadedAnalyzeCommandBase` artifacts generation and enforcing JSON properties ordering. [#555](https://github.com/microsoft/binskim/pull/555)
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* BUGFIX: Fix ` KeyNotFoundException` raised on `BA2006.BuildWithSecureTools` when required properties are removed from the config file. [#565](https://github.com/microsoft/binskim/pull/565)
+* BUGFIX: Fix `KeyNotFoundException` exception raised by `BA2006.BuildWithSecureTools` when individual `MinimumToolVersions` properties are removed from XML configuration. [#565](https://github.com/microsoft/binskim/pull/565)
 
 ## **v1.9.2** [NuGet Package](https://www.nuget.org/packages/Microsoft.CodeAnalysis.BinSkim/1.9.2)
 


src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -66,6 +66,29 @@ public IEnumerable<IOption> GetOptions()
             new PerLanguageOption<AdvancedMitigations>(
                 AnalyzerName, nameof(AdvancedMitigationsEnforced), defaultValue: () => AdvancedMitigations.None);
 
+        public override void Initialize(BinaryAnalyzerContext context)
```
michaelcfanning, line 69: >Initialize [](http://example.com/codeflow?start=29&length=10)

This is a very clever fix, Shaopeng, and we could consider taking it. But there's not a test or release note with your proposal.

I strongly encourage the team to please always produce release notes and tests *before* diving into code.

I would have merged this change if those things had been available. :)

I will continue to repeat these points to the team over and over, that is my role. 

Authoring a release note is a trivial task. Now, authoring a comprehensive set of unit tests might be more expense than a draft PR warrants. In *all* cases, however, if you haven't investigated the existing test strategy, that's a gap.

michaelcfanning: @eddynaka or @shaopeng-gh, if you see these, please add the following PR note to this change, which I would like to merge. If someone has time to add a unit test as part of that, do so, otherwise please open an issue and we'll add a test as an offline activity.

## BUGFIX: Fix `KeyNotFoundException` on `BA2006.BuildWithSecureTools` when required properties are removed from the config file. [#565](https://github.com/microsoft/binskim/pull/565)

eddynaka: sounds good.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -66,6 +66,29 @@ public IEnumerable<IOption> GetOptions()
...
+            // This is to make sure other default settings are applied as well.
+            foreach (KeyValuePair<string, Version> defaultMinimumToolVersion in defaultMinimumToolVersions)
+            {
+                currentMinimumToolVersions.TryAdd(defaultMinimumToolVersion.Key, defaultMinimumToolVersion.Value);
+            }
+
+            context.Policy.SetProperty(
+                MinimumToolVersions,
+                currentMinimumToolVersions,
+                cacheDescription: false,
```
michaelcfanning, line 86: I actually think cache description == true is correct here, but we can scrub for this later. btw - someone should go look at all our other rules for instances of this bug.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -66,6 +66,29 @@ public IEnumerable<IOption> GetOptions()
...
+            if (context.Policy == null) { return; }
+
+            StringToVersionMap currentMinimumToolVersions = context.Policy.GetProperty(MinimumToolVersions, cacheDefault: false);
+            StringToVersionMap defaultMinimumToolVersions = BuildMinimumToolVersionsMap();
+
+            // If user only overwrite part of the settings, currentMinimumToolVersions will only have those settings.
+            // This is to make sure other default settings are applied as well.
+            foreach (KeyValuePair<string, Version> defaultMinimumToolVersion in defaultMinimumToolVersions)
+            {
+                currentMinimumToolVersions.TryAdd(defaultMinimumToolVersion.Key, defaultMinimumToolVersion.Value);
```
eddynaka, line 80: >currentMinimumToolVersions.TryAdd(defaultMinimumToolVersion.Key, defaultMinimumToolVersion.Value); [](http://example.com/codeflow?start=16&length=98)

@michaelcfanning , If we do this, the user would not be able to remove and use the min version since we are always defaulting to our current mapping.

Probably here we would have to set to new Version()

eddynaka: the cases i thought:
1. user is using default -> policy should be the same as the default mapping
2. user customized one version -> that version should be customized
3. user removed one version -> that version should show as 0.0.0.0

src/ReleaseHistory.md
```diff
@@ -1,5 +1,9 @@
 # BinSkim Release History
 
+## Unreleased
+
+* BUGFIX: Fix ` KeyNotFoundException` raised on `BA2006.BuildWithSecureTools` when required properties are removed from the config file. [#565](https://github.com/microsoft/binskim/pull/565)
```
michaelcfanning, line 5: >KeyNotFoundException [](http://example.com/codeflow?start=16&length=20)

There is an unnecessary space before keynotfoundexception. it should be 'raised by' not raised on. We should mention the configuration option where this occurs as it is data a customer is likely to search on (please always try to imagine your way into this).

I will make these changes and update the branch right now.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
michaelcfanning: Good test cases but your expected outcomes aren't quite right. In #3, if a user removes the version, that value should return to the default, which is the value that's encoded in the rule. This is the behavior of the configuration: if you explicitly refer to it, you can override it. If you do *not* refer to it, you get the built-in default behavior.

This is why Shaopeng's fix is correct. At runtime, he traverses the configuration that comes in and re-adds any properties that are absent.

What customers needs to do if they want to, say, permit all C compiler versions, is to leave an explicit reference to the compiler and to explicitly set its value to zero. After this, all compiler references to that version will be accepted.

PR #564, opened 2022-02-11
Adding stress test powershell

eddynaka:
This PR is adding a stress test script where we can execute commands like:
```
.\StressTest.ps1 -BinSkimFolder "C:\my-folder\win-x64" -InputPaths "C:\my-folder-to-analyze\*.dll"
```

You would also be able to configure the number of times:
```
.\StressTest.ps1 -BinSkimFolder "C:\my-folder\win-x64" -InputPaths "C:\my-folder-to-analyze\*.dll" -Times 1000
```


StressTest.ps1
```diff
@@ -0,0 +1,24 @@
+﻿[CmdletBinding()]
+param(
+    [string]
+    $BinSkimFolder = "bld\bin\x64_Release\netcoreapp3.1",
+
+    [string]
+    $OutputFolder = "c:\temp\",
```
michaelcfanning, line 7: >$OutputFolder [](http://example.com/codeflow?start=4&length=13)

Use an environment variable here? %TEMP%?

StressTest.ps1
```diff
@@ -0,0 +1,24 @@
...
+
+    [string]
+    $OutputFile = "binskim",
```
michaelcfanning, line 10: >$OutputFile [](http://example.com/codeflow?start=4&length=11)

It'd be nice if this was an identifier encoded in something else, like 'BinSkimHASHESRUN'

StressTest.ps1
```diff
@@ -0,0 +1,24 @@
...
+
+    [string]
+    $InputPaths,
+
+    [int]
+    $Times = 1000
```
michaelcfanning, line 16: >$Times [](http://example.com/codeflow?start=4&length=6)

Add a sentinel value that runs indefinitely? Or should people just set a super large value?

StressTest.ps1
michaelcfanning, line 7: >$OutputFolder [](http://example.com/codeflow?start=4&length=13)

shouldn't we consider creating a special directory in temp for this data?

StressTest.ps1
```diff
@@ -0,0 +1,24 @@
+﻿[CmdletBinding()]
+param(
+    [string]
+    $BinSkimFolder = "bld\bin\x64_Release\netcoreapp3.1",
```
michaelcfanning, line 4: >$BinSkimFolder [](http://example.com/codeflow?start=4&length=14)

Can we put this test file in a better place and use a relative path to get to this folder?

michaelcfanning, line 4: >netcoreapp3 [](http://example.com/codeflow?start=42&length=11)

is this script cross-platform friendly or not? you have hard-coded directory separators here. 

do we care if this script if cross-platform friendly?

StressTest.ps1
```diff
@@ -0,0 +1,24 @@
...
+
+    [int]
+    $Times = 1000
+)
+
+for ($i = 0; $i -lt $Times; $i++)
+{
+    $command = $BinSkimFolder + "\BinSkim.exe analyze --recurse --hashes --force --output " + $OutputFolder + $OutputFile + $i + ".sarif " + $InputFolder
+    Write-Host $command
+    Invoke-Expression $command
```
michaelcfanning, line 23: >$command [](http://example.com/codeflow?start=22&length=8)

I think a special mode in this script where it breaks in the case that the output differs from previous run could be helpful.

Then stress could run indefinitely until there's a break.

StressTest.ps1
eddynaka: Removed this, since i changed to a while loop.
Maybe we could add an argument like "stop when reach 100000".

This will prevent it from running forever

StressTest.ps1
eddynaka: done.

StressTest.ps1
eddynaka: Created a scripts directory.
we should also move the other files in the future.

StressTest.ps1
marmegh: What about parameter sets?

scripts/StressTest.ps1
```diff
@@ -0,0 +1,57 @@
...
+    
+    if ($i -ne 0)
+    {
+        $PreviousOutputFilePath = $OutputFolder + $OutputFileName + ($i - 1) + ".sarif"
+    }
+    
+    $command = $BinSkimFolder + "\BinSkim.exe analyze --recurse --hashes --force --output " + $CurrentOutputFilePath + " " + $InputPaths
+    Invoke-Expression $command
+
+    if ($PreviousOutputFilePath -ne "")
```
marmegh, line 41: The if statements can be combined:
`if ($i -ne 0)
    {
        $PreviousOutputFilePath = $OutputFolder + $OutputFileName + ($i - 1) + ".sarif"
        $CurrentFileSize = (Get-Item $CurrentOutputFilePath).Length/1KB
        $PreviousFileSize = (Get-Item $PreviousOutputFilePath).Length/1KB
        if ($CurrentFileSize -eq $PreviousFileSize)
        {
            Remove-Item $PreviousOutputFilePath
        }
        else
        {
            break
        }
    }`

scripts/StressTest.ps1
```diff
@@ -0,0 +1,57 @@
...
+        $PreviousFileSize = (Get-Item $PreviousOutputFilePath).Length/1KB
+
+        if ($CurrentFileSize -eq $PreviousFileSize)
+        {
+            Remove-Item $PreviousOutputFilePath
+        }
+        else
+        {
+            break
+        }
```
marmegh, line 53: Can we maybe note the problem, like:
`Write-Host "Previous output was $($PreviousFileSize) but latest is $($CurrentFileSize)."`

scripts/StressTest.ps1
```diff
@@ -0,0 +1,57 @@
...
+[CmdletBinding()]
+param(
+    [string]
+    $BinSkimFolder = "..\bld\bin\x64_Release\netcoreapp3.1",
+
+    [string]
+    $OutputFileName = "binskim-hash-run-",
+
+    [string]
+    $InputPaths
```
marmegh, line 21: If you add
`[Parameter(Mandatory=$true)]`
this should block looping on an error because of no input path.

scripts/StressTest.ps1
eddynaka: updated. thanks

scripts/StressTest.ps1
eddynaka: updated. thanks

scripts/StressTest.ps1
eddynaka: updated. thanks

michaelcfanning (approved):
:shipit:

marmegh:
Was it intentional to delete the powershell script?

PR #565, opened 2022-02-11
Fix `KeyNotFoundException` when required properties are removed from the config file

eddynaka:
# Description

With the new release (v1.9.2), some customers encountered a `KeyNotFoundException` when running the rule `BA2006.BuildWithSecureTools`. This issue was occurring because they have a special configuration where they removed required properties from the xml. Below is the example where the property for language `C` got removed:
```xml
  <Properties Key="BA2006.BuildWithSecureTools.Options" Type="PropertiesDictionary">
    <!---->
    <Property Key="AdvancedMitigationsEnforced" Value="None" Type="AdvancedMitigations" />
    <!---->
    <Properties Key="AllowedLibraries" Type="IL.Rules.StringToVersionMap">
      <Property Key="libeay32.lib,unknown" Value="0.0.0.0" Type="System.Version" />
    </Properties>
    <!---->
    <Properties Key="MinimumToolVersions" Type="IL.Rules.StringToVersionMap">
      <Property Key="Cxx" Value="17.0.65501.17013" Type="System.Version" />
      <Property Key="MinimumXboxCompilerVersion" Value="16.0.11886.0" Type="System.Version" />
      <Property Key="Unknown" Value="2147483647.2147483647.2147483647.2147483647" Type="System.Version" />
    </Properties>
```

# Fix

Before accessing the key, we should always verify if the key exists and if not, we should return the min/max Version (depending on the rule). With that in mind, I created an extension in the `StringToVersionMap` to safely retrieve a value based on a Key.

@marmegh


src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -106,10 +109,9 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
 
             var goodCompilers = new HashSet<string>();
 
-            var badModules = new List<ObjectModuleDetails>();
             StringToVersionMap allowedLibraries = context.Policy.GetProperty(AllowedLibraries);
 
-            var languageToBadModules = new Dictionary<Language, List<ObjectModuleDetails>>();
```
eddynaka, line 112: badModules list was being updated and languageToBadModules weren't. Changed the logic to use the dictionary variable and that will update the string minimumRequiredCompilers which was always blank (review baseline)

src/BinSkim.Rules/StringToVersionMap.cs
```diff
@@ -11,4 +11,25 @@ namespace Microsoft.CodeAnalysis.IL.Rules
     public class StringToVersionMap : TypedPropertiesDictionary<Version>
     {
     }
+
+    public static class StringToVersionMapExtensions
```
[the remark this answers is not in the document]
eddynaka: the motive to place here is that any rule that uses StringToVersionMap would be able to use the extension.

src/Test.UnitTests.BinSkim.Driver/MultithreadedAnalyzeCommandTests.cs
```diff
@@ -129,7 +129,7 @@ public void MultithreadedAnalyzeCommand_ReturnCommonPathRootFromTargetSpecifiers
                 // Testing the same string array in a random order.
                 // This will guarantee that the sorting is working as expected.
                 commonPath = MultithreadedAnalyzeCommand.ReturnCommonPathRootFromTargetSpecifiersIfOneExists(Shuffle(testCase.TargetFileSpecifiers));
```
eddynaka, line 131: >Shuffle [](http://example.com/codeflow?start=109&length=7)

we are generating a shuffled array. The order might change the expected common path sensitivity (c:/ vs C:/)

This change will guarantee that the order won't matter.

src/BinSkim.Rules/StringToVersionMap.cs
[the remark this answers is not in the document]
eddynaka: I reviewed all the other usages of this class and none would generate the `KeyNotFoundException`.

src/BinSkim.Rules/StringToVersionMap.cs
```diff
@@ -11,4 +11,25 @@ namespace Microsoft.CodeAnalysis.IL.Rules
...
+    {
+        internal static readonly Version s_maxVersion = new Version(int.MaxValue, int.MaxValue, int.MaxValue, int.MaxValue);
+        internal static readonly Version s_minVersion = new Version(0, 0, 0, 0);
+
+        public static Version GetVersionByKey(this StringToVersionMap stringToVersionMap, string key, bool returnMaxValueIfKeyDoesNotExist = false)
```
eddynaka, line 20: >returnMaxValueIfKeyDoesNotExist [](http://example.com/codeflow?start=107&length=31)

In case the property is missing, this means that the user removed the property because that wasn't relevant. So, we should default to minValue. Otherwise, we would require the user to fix the version of the compiler.

src/ReleaseHistory.md
```diff
@@ -1,5 +1,9 @@
 # BinSkim Release History
 
+## Unreleased
+
+* BUGFIX: Fix `KeyNotFoundException` when required properties are removed from the config file. [#565](https://github.com/microsoft/binskim/pull/565)
```
michaelcfanning, line 5: >KeyNotFoundException [](http://example.com/codeflow?start=15&length=20)

Mention the rules that are problematic, both the rule id and friendly.

src/Test.UnitTests.BinSkim.Rules/BuildWithSecureToolsTests.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    public class BuildWithSecureToolsTests
+    {
+        private static readonly Random s_random;
+        private static readonly int s_randomSeed;
+
+        static BuildWithSecureToolsTests()
+        {
+            s_randomSeed = (int)DateTime.UtcNow.TimeOfDay.TotalMilliseconds;
+            s_random = new Random(s_randomSeed);
```
michaelcfanning, line 26: >s_randomSeed [](http://example.com/codeflow?start=34&length=12)

You always need to report the random seed when you use this pattern.

src/Test.UnitTests.BinSkim.Rules/BuildWithSecureToolsTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        [Fact]
+        public void BuildMinimumCompilersList_ShouldAlwaysBeOrdered()
```
michaelcfanning, line 30: >BuildMinimumCompilersList_ShouldAlwaysBeOrdered [](http://example.com/codeflow?start=20&length=47)

Please identify good tests cases for this and implement them.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -106,10 +107,9 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
...
+            var languagesToBadModules = new Dictionary<Language, List<ObjectModuleDetails>>();
```
michaelcfanning, line 112: >languagesToBadModules [](http://example.com/codeflow?start=16&length=21)

please revert this name, the pattern is key type + value type.

src/Test.UnitTests.BinSkim.Rules/StringToVersionMapTests.cs
```diff
@@ -0,0 +1,84 @@
...
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.IL.Rules
+{
+    public class StringToVersionMapTests
+    {
+        [Fact]
+        public void StringToVersionMap_GetVersionByKey_ShouldNotKeyNotFoundException_IfKeyDoesNotExist()
+        {
+            const string key1 = "key1";
```
michaelcfanning, line 19: >key1 [](http://example.com/codeflow?start=25&length=4)

let''s avoid numbers in variable names.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
eddynaka: done

src/Test.UnitTests.BinSkim.Rules/StringToVersionMapTests.cs
eddynaka: removed the entire class and the extension

src/ReleaseHistory.md
eddynaka: updated.

src/Test.UnitTests.BinSkim.Rules/BuildWithSecureToolsTests.cs
eddynaka: removed this in favor of a simple reversed list

eddynaka:
In the BuildWithSecureToolsTests, I added two tests to verify if an empty list of policies would get the correct value and the case where the user didn't change the configuration.

---
In reply to: [1036948370](https://github.com/microsoft/binskim/pull/565#issuecomment-1036948370) [](http://example.com/codeflow?ancestors=1036948370,1036948370)

src/Test.UnitTests.BinSkim.Rules/BuildWithSecureToolsTests.cs
eddynaka: removed the tests, since the function is no longer used.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -54,6 +56,15 @@
 
         private const string MIN_XBOX_COMPILER_VER = "MinimumXboxCompilerVersion";
 
+        public static PerLanguageOption<Version> MinimumCCompilerVersion { get; } =
```
michaelcfanning, line 59: >static [](http://example.com/codeflow?start=15&length=6)

Please follow the style convention in this file for options, which requires a line between each option.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -54,6 +56,15 @@
...
+            new PerLanguageOption<Version>(AnalyzerName, nameof(Language.C), defaultValue: () => new Version());
+        public static PerLanguageOption<Version> MinimumCxxCompilerVersion { get; } =
+            new PerLanguageOption<Version>(AnalyzerName, nameof(Language.Cxx), defaultValue: () => new Version());
+        public static PerLanguageOption<Version> MinimumUnknownCompilerVersion { get; } =
+            new PerLanguageOption<Version>(AnalyzerName, nameof(Language.Unknown), defaultValue: () => new Version());
+        public static PerLanguageOption<Version> MinimumXboxCompilerVersion { get; } =
+            new PerLanguageOption<Version>(AnalyzerName, MIN_XBOX_COMPILER_VER, defaultValue: () => new Version());
```
michaelcfanning, line 66: >MIN_XBOX_COMPILER_VER [](http://example.com/codeflow?start=57&length=21)

please get rid of this constant and use nameof(PropertyName) as we do elsewhere.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -258,33 +210,23 @@
 
                 if (foundIssue)
                 {
+                    if (!languageToBadModules.TryGetValue(omDetails.Language, out List<ObjectModuleDetails> badModules))
+                    {
+                        badModules = new List<ObjectModuleDetails>();
+                    }
+
                     badModules.Add(omDetails);
+                    languageToBadModules[omDetails.Language] = badModules;
```
michaelcfanning, line 219: >languageToBadModules [](http://example.com/codeflow?start=20&length=20)

you are always setting bad modules back in the dictionary, instead of simply doing so when it couldn't be found. Move this code above into that TryGetValue case.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -258,33 +210,23 @@
...
                 }
                 else
                 {
                     goodCompilers.Add(BuildCompilerIdentifier(omDetails));
```
michaelcfanning, line 223: >goodCompilers [](http://example.com/codeflow?start=20&length=13)

While we're here can you please change this variable name to 'inPolicyCompilers'? This is more descriptive than 'good'.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -258,33 +210,23 @@
 
                 if (foundIssue)
                 {
+                    if (!languageToBadModules.TryGetValue(omDetails.Language, out List<ObjectModuleDetails> badModules))
+                    {
+                        badModules = new List<ObjectModuleDetails>();
```
michaelcfanning, line 215: >badModules [](http://example.com/codeflow?start=24&length=10)

change this name to 'outOfPolicyModules'

michaelcfanning:
>             context.Logger.Log(this,

this code does not meet our style conventions

---
Refers to: src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs:237 in d157a56. [](commit_id = d157a569fcce738e848d0db2f0a45994aaa76eab, deletion_comment = False)

michaelcfanning:
>             string[] sorted = goodCompilers.ToArray();

'sortedInPolicyCompilers'

---
Refers to: src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs:233 in d157a56. [](commit_id = d157a569fcce738e848d0db2f0a45994aaa76eab, deletion_comment = False)

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -299,93 +241,124 @@
...
 
-        private string BuildCompilerIdentifier(ObjectModuleDetails omDetails)
-        {
-            return omDetails.CompilerName + ":" + omDetails.Language + ":" + omDetails.CompilerBackEndVersion;
-        }
-
-        private string BuildMinimumCompilersList(BinaryAnalyzerContext context, Dictionary<Language, List<ObjectModuleDetails>> languageToBadModules)
+        internal void GenerateMessageParametersAndLog(BinaryAnalyzerContext context, Dictionary<Language, List<ObjectModuleDetails>> languageToBadModules)
         {
+            var sb = new StringBuilder();
```
michaelcfanning, line 246: >StringBuilder [](http://example.com/codeflow?start=25&length=13)

Can we use a per-instance member variable for this? Our threading model will permit it, won't it?

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -299,93 +241,124 @@
...
             var languages = new List<string>();
-
-            foreach (Language language in languageToBadModules.Keys)
+            foreach (KeyValuePair<Language, List<ObjectModuleDetails>> kp in languageToBadModules.OrderBy(l => l.Key))
```
michaelcfanning, line 248: >foreach (KeyValuePair<Language, List<ObjectModuleDetails>> kp in languageToBadModules.OrderBy(l => l.Key)) [](http://example.com/codeflow?start=12&length=106)

please clarify this code. in general, I'd prefer not to see computation embedded in a for or foreach construct. 

If we have a need to sort this data, I'd tend to use a data structure that itself sorts keys, please go see if SortedDictionary will serve our purposes here.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -299,93 +241,124 @@
...
             {
-                Version version = context.Policy.GetProperty(MinimumToolVersions)[language.ToString()];
-                languages.Add($"{language} ({version})");
+                sb.Append(kp.Value.CreateOutputCoalescedByCompiler());
+
+                Version version = RetrieveMinimumCompilerVersionByLanguage(context, kp.Key);
```
michaelcfanning, line 252: >kp.Key [](http://example.com/codeflow?start=84&length=6)

when you embed things like 'kp.Key' as an argument, you lose readability. For this code, I might have passed this using the parameter naming convention or created a local that retains information that the key is the language.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
```diff
@@ -299,93 +241,124 @@
...
+                languages.Add($"{kp.Key} ({version})");
             }
-            return string.Join(", ", languages);
+
+            string badModulesText = sb.ToString();
+            string minimumRequiredCompilers = string.Join(", ", languages);
+
+            Debug.Assert(!string.IsNullOrWhiteSpace(badModulesText) ||
```
michaelcfanning, line 259: >string [](http://example.com/codeflow?start=26&length=6)

this code does not meet our style guidelines

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
eddynaka: thanks.
changed to sorted dictionary.

src/BinSkim.Rules/PERules/BA2006.BuildWithSecureTools.cs
eddynaka: yes.
made the change.
thanks

eddynaka:
Closing PR in favor of https://github.com/microsoft/binskim/pull/563

PR #566, opened 2022-02-12
Fix the EnableSecureSourceCodeHashing description

BruceHaley:
In my case the EnableSecureSourceCodeHashing error was not due to incorrect hashing. This added note would have saved me half a day trying to understand and find the cause:
_Note: This error will also result from BinSkim not having access to scan the PDBs._


eddynaka:
I tried to simulate this.
The only case I was able to do us providing a binary with embedded PDB.

Otherwise, I generated the "E_PDB_NOT_FOUND" error.

---
In reply to: [1036889699](https://github.com/microsoft/binskim/pull/566#issuecomment-1036889699) [](http://example.com/codeflow?ancestors=1036889699,1036889699)

BruceHaley:
Hi Eddy, Michael,

Thanks, Eddy, for following up with me in Teams on 2/14. Working together I think we both came to understand how to reproduce the error. In my opinion, this is definitely a BinSkim bug. Let me explain:

I encountered the error in this ADO pipeline called [BinSkim-DotNet-v5](https://fuselabs.visualstudio.com/SDK_v4/_build/results?buildId=290921&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=e900fa28-66c4-55e5-4e9c-0ba63fb92fbd).

The pipeline copied .nupkg files to an artifacts staging diirectory (D:\a\1\a\Signing), unpacked them, then ran BinSkim against that staging directory. BinSkim was given this Target Glob Pattern to scan: 
`$(Build.ArtifactStagingDirectory)\Signing\**\*.dll;$(Build.ArtifactStagingDirectory)\Signing\**\*.exe;`

If BinSkim had scanned only the glob pattern given to it, BinSkim would have correctly reported that it found no .pdb files. 

Instead, BinSkim failed to adhere to the glob pattern given to it. Searching elsewhere on the disk, it found some .pdb files in .obj folders. Those files, being intermediate object files, were not hashed. So BinSkim reported EnableSecureSourceCodeHashing errors. Had BinSkim gone on searching, it might have found those same files in their .bin folders correctly hashed.

I had no clue that BinSkim was searching outside the glob pattern that was given to it. That is why I found the error message so baffling. Fortunately Eddy knew how to debug BinSkim and discovered what was going on.

I would describe the BinSkim bug as follows:

Expected behavior: BinSkim would search within the bounds of the Target Glob Pattern given to it in the ADO task. It would report errors consistent with that search area.

Actual behavior: BinSkim searched outside the bounds of the glob pattern given to it. It did not report errors consistent with that search area.

Thank you.
--Bruce

BruceHaley:
I see now my offered documentation fix is incorrect. The error resulted from BinSkim scanning the intermediate unhashed .pdb files instead of the final hashed ones.

PR #574, opened 2022-02-15
Adding SARIF ReleaseHistory to the package

eddynaka:
This change will add the ReleaseHistory.md from the sarif-sdk submodule to the nupkg.

### DIFF
--- a/src/Nuget/BinSkim.nuspec
+++ b/src/Nuget/BinSkim.nuspec
@@ -19,7 +19,7 @@
   <files>
     <file src="bld\bin\x64_$configuration$\Publish\**\*" target="tools"/>   
     <file src="src\ReleaseHistory.md" target="tools\ReleaseHistory.md" />
-    <file src="src\sarif-sdk\src\ReleaseHistory.md" target="tools\ReleaseHistory.SARIF.SDK.md" />
+    <file src="src\sarif-sdk\src\ReleaseHistory.md" target="tools\ReleaseHistory-Sarif-Sdk.md" />
     <file src="src\BinaryParsers\**\*.cs" target="src\BinaryParsers" />
     <file src="src\BinSkim.Driver\**\*.cs" target="src\BinSkim.Driver" />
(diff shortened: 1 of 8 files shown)


marmegh (approved):
:shipit:

src/Nuget/BinSkim.nuspec
```diff
@@ -19,6 +19,7 @@
   <files>
     <file src="bld\bin\x64_$configuration$\Publish\**\*" target="tools"/>   
     <file src="src\ReleaseHistory.md" target="tools\ReleaseHistory.md" />
+    <file src="src\sarif-sdk\src\ReleaseHistory.md" target="tools\ReleaseHistory.SARIF.SDK.md" />
```
michaelcfanning, line 22: >ReleaseHistory.SARIF.SDK. [](http://example.com/codeflow?start=66&length=25)

Can you do me a favor and please use hyphens here? 

ReleaseHistory-Sarif-Sdk.md

I generally try to avoid introducing 'periods' into file names, as they can confuse code that is looking for a file extension. Consider your file, is '.md' the extension? Or is it perhaps '.sdk.md'? :) As you recall, we were just looking at the '.tar.gz' extension for zip archives on the unix platform.

I know that .NET has a convention of including periods as namespace separators. This works because assembly extensions (.dll and .exe) are very well known but also because the file name as a whole matches something well-known (a root namespace).

michaelcfanning: ReleaseHistory-SarifSdk.md is also fine
ReleaseHistorySarifSdk.md is also fine
ReleaseHistory_SarifSdk.md is also fine

:)

eddynaka: thanks!
I updated to -Sarif-Sdk.md option.

michaelcfanning: thanks. this is a small thing but we should codify an extension moving forward so that we agree on a consistent naming pattern. this is an extremely trivial thing to spend cycles discussing. :)

Issue #554, opened 2022-01-28
This doesn't run on macOS


dalemyers:
As a dotnet tool I'd hope this would run on macOS. Unfortunately, when I try, I get this error:

```
$ dotnet microsoft.codeanalysis.binskim/1.9.1/tools/netcoreapp3.1/win-x64/BinSkim.exe
A fatal error was encountered. The library 'libhostpolicy.dylib' required to execute the application was not found in '[snip]/microsoft.codeanalysis.binskim/1.9.1/tools/netcoreapp3.1/win-x64/'.
Failed to run as a self-contained app.
  - The application was run as a self-contained app because '[snip]/microsoft.codeanalysis.binskim/1.9.1/tools/netcoreapp3.1/win-x64/BinSkim.runtimeconfig.json' did not specify a framework.
  - If this should be a framework-dependent app, specify the appropriate framework in '[snip]/microsoft.codeanalysis.binskim/1.9.1/tools/netcoreapp3.1/win-x64/BinSkim.runtimeconfig.json'.
```

Is there any way to work around this? Thanks!

marmegh:
Hi @dalemyers.
Yes, binskim should be able to run on any machine running .NET core. However, there are limitations trying to run BinSkim on macOS.
It looks like you are attempting to run the windows version of binskim. Please try running the linux version. Based on the location in your error message above, it will likely be in this folder:
`microsoft.codeanalysis.binskim/1.9.1/tools/netcoreapp3.1/linux-x64/`.

dalemyers:
How do we run it on macOS in that case? Running `dotnet BinSkim` doesn't work saying that the command or file wasn't found. And making it executable and running it using `./BinSkim` fails with "exec format error".

marmegh:
@jeffersonking, let's schedule some time to walk through this scenario to repro and identify some stronger guidance for @dalemyers.

dalemyers:
dotnet is available. It's saying that `BinSkim` isn't.

michaelcfanning:
Oh, sorry. 

try 'dotnet BinSkim.dll help' with the working directory set to the build output location.

dalemyers:
I'm not sure what you mean by "working directory set to the build output location". Do you mean simply to set the current directory to `/path/to/extracted/package/tools/netcoreapp3.1/linux-x64` ? 

If so, I did that, then ran `dotnet BinSkim.dll help` and this was the output:

```
[...]

PR #576, opened 2022-02-16
Add osx-x64 release

marmegh:
Addressing issue #554 
@jeffersonking, recommended adding osx-x64 specific release binaries per the [dotnet documentation on runtime identifier (RID)](https://docs.microsoft.com/en-us/dotnet/core/rid-catalog).

@eddynaka and @michaelcfanning, still need to get the result tested on mac to see if this resolves the issue. Will update accordingly and publish for formal review once this step is complete. Feel free to add any feedback before this.

image

Results of test run (verify code signing and codesign validation steps, disabled publish steps): 
image


eddynaka:
@marmegh ,

The change looks ok but we should test it.
My suggestion is: update the yaml pipeline to test on a mac. using the command line, you would be able to verify how to run it and analyze binskim itself.

michaelcfanning:
@eddynaka, can you help make your proposed pipeline change, it's an excellent suggestion.

eddynaka:
Sorry for the delay.
I was able to produce a few changes in some files to test during the build pipeline.

---
In reply to: [1108956840](https://github.com/microsoft/binskim/pull/576#issuecomment-1108956840) [](http://example.com/codeflow?ancestors=1108956840,1108956840)

ado-build.yml
```diff
@@ -32,3 +27,23 @@ jobs:
       script: 'BuildAndTest.cmd'
 
   - task: ComponentGovernanceComponentDetection@0
+    inputs:
+      ignoreDirectories: '..\src\sarif-sdk'
```
eddynaka: we already have component governance in sarif-sdk. We don't need to analyze here as well.

So, let's ignore the entire submodule.

ado-build.yml
```diff
@@ -32,3 +27,23 @@ jobs:
...
+  - task: Bash@3
+    displayName: 'Build and Test'
+    inputs:
+      filePath: 'BuildAndTest.sh'
+
+  - task: Bash@3
+    displayName: 'Run BinSkim'
+    inputs:
+      targetType: 'inline'
+      script: 'dotnet bld/bin/x64_Release/netcoreapp3.1/binskim.dll analyze src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/macho.*'
```
eddynaka, line 49: >dotnet bld/bin/x64_Release/netcoreapp3.1/binskim.dll [](http://example.com/codeflow?start=15&length=52)

we should add this in some doc.
the user would need to execute:
`dotnet binskim.dll analyze ....`

src/Test.FunctionalTests.BinSkim.Driver/AnalyzeCommandTests.cs
```diff
@@ -65,7 +65,11 @@ public void AnalyzeCommand_ReadSarifLog_ShouldBeAbleToReadCurrent()
         [Obsolete]
         public void AnalyzeCommand_Hashes_ShouldUpdateDataToInsert()
```
eddynaka, line 66: >AnalyzeCommand_Hashes_ShouldUpdateDataToInsert [](http://example.com/codeflow?start=20&length=46)

This test were failing in the unix pipelines. With the changes below, i was able to make it pass.

eddynaka:
FYI, the windows pipeline isn't working properly.
I opened an issue in the ComponentGovernance team since its throwing an exception.

---
In reply to: [1113754507](https://github.com/microsoft/binskim/pull/576#issuecomment-1113754507) [](http://example.com/codeflow?ancestors=1113754507,1108956840,1108956840)

eddynaka:
For the point above, i was able to fix it already. :)

---
In reply to: [1113758229](https://github.com/microsoft/binskim/pull/576#issuecomment-1113758229) [](http://example.com/codeflow?ancestors=1113758229,1108956840)

README.md
```diff
@@ -22,9 +22,9 @@ This repository contains the source code for BinSkim, a Portable Executable (PE)
 ### How to extract the exe file from the nuget package
 If you only want to run the Binskim tool without installing anything, then you can 
 1. Download BinSkim from **[NuGet](https://www.nuget.org/packages/Microsoft.CodeAnalysis.BinSkim/)**
-2. Rename the file extension from .nupkg to .zip
+2. Rename the file extension from .nupkg to .zip (ie. via commandline: `rename microsoft.codeanalysis.binskim.x.y.z.nupkg microsoft.codeanalysis.binskim.1.9.4.zip`)
```
eddynaka, line 25: >1.9.4 [](http://example.com/codeflow?start=153&length=5)

update this to x.y.z

src/ReleaseHistory.md
```diff
@@ -5,6 +5,7 @@
 * Bump ELFSharp from 2.13.2 to 2.14.0. [#628](https://github.com/microsoft/binskim/pull/628)
 * Bump System.Reflection.Metadata from 5.0.0 to 6.0.1 and System.Collections.Immutable from 5.0.0 to 6.0.0. [#605](https://github.com/microsoft/binskim/pull/605)
 * Bump ELFSharp from 2.14.0 to 2.15.0. [#631](https://github.com/microsoft/binskim/pull/631)
+* BUGFIX: Unblock running binskim on MacOS. [#576](https://github.com/microsoft/binskim/pull/576)
```
eddynaka, line 8: >BUGFIX [](http://example.com/codeflow?start=2&length=6)

not sure if this is considerable a bugfix.
maybe `feature: enable BinSkim for MacOS`. or something like that.

eddynaka (approved):
:shipit:

src/ReleaseHistory.md
```diff
@@ -5,6 +5,7 @@
 * Bump ELFSharp from 2.13.2 to 2.14.0. [#628](https://github.com/microsoft/binskim/pull/628)
 * Bump System.Reflection.Metadata from 5.0.0 to 6.0.1 and System.Collections.Immutable from 5.0.0 to 6.0.0. [#605](https://github.com/microsoft/binskim/pull/605)
 * Bump ELFSharp from 2.14.0 to 2.15.0. [#631](https://github.com/microsoft/binskim/pull/631)
+* FEATURE: Enable BinSkim for MacOS. [#576](https://github.com/microsoft/binskim/pull/576)
 - Bump Sarif.Sdk by updating submodule from [4e9f606 to fc9a9df](https://github.com/microsoft/sarif-sdk/compare/4e9f606bb0e88428866e253352cdc70dc68f98cb...fc9a9dfb865096b5aaa9fa3651854670940f7459). [#638](https://github.com/microsoft/binskim/pull/638)
```
shaopeng-gh, line 9: >-  [](http://example.com/codeflow?start=0&length=2)

looks like a typo should be * ?

BuildAndTest.sh
```diff
@@ -1,11 +1,12 @@
 #!/bin/bash
```
shaopeng-gh, line 1: there is also a file "UpdateBaselines.sh" which was for Linux only, not sure if we want to update that.

ado-build.yml
shaopeng-gh: +1

I see one candidate place,
at the bottom of the readme.md
we currently have:

`Example: binskim.exe analyze c:\bld\*.dll --recurse --output MyRun.sarif`

I think we can change it to 3 OS version example here.

we can also consider if want to include the path in the example to make it more clear, because the previous step is unzip.

shaopeng-gh (approved):
:shipit:

ado-build.yml
shaopeng-gh: more info fyi
I found that we have this file "BinSkim" in the package
`./BinSkim analyze`
I just tried works in Linux.

marmegh: @shaopeng-gh, could you clarify where you found this file?

BuildAndTest.sh
marmegh: What would be the proposed change? I don't see any reference in that file to BinSkimLinux.sln or the uname variable.

ado-build.yml
shaopeng-gh: download package from
[https://www.nuget.org/packages/Microsoft.CodeAnalysis.BinSkim/](https://www.nuget.org/packages/Microsoft.CodeAnalysis.BinSkim/)
(which is the result of our build and package)
unzip
go to folder \tools\netcoreapp3.1\linux-x64
file "BinSkim" without ext

marmegh: Apologies. I was confused since this change is specific to osx. You were being helpful, I was concerned you were telling me we had a binary for linux in the wrong location.

BuildAndTest.sh
shaopeng-gh: oh because I see this file is running under folder /netcoreapp3.1/linux-x64/
I don't think we need another one running for macOS.

ado-build.yml
shaopeng-gh: macOS also have a same file in same location, after your change:
\tools\netcoreapp3.1\osx-x64\BinSkim

I point out in the case we want to add doc for user
see if we want to say use 
`./BinSkim analyze`
or 
`dotnet binskim.dll analyze `

eddynaka: When I created this, I didn't want to change the current directory that we are right now. So, it is why we have the full path.

But, in the end, it is the same as dotnet binskim.dll analyze.

eddynaka: Mary also updated the readme showing where the files will be. so, I think, Shaopeng's suggestion is to add the dotnet binskim.dll analyze as a new step in the readme

shaopeng-gh: I think, Shaopeng's suggestion is to add the dotnet binskim.dll analyze as a new step in the readme  
------ yes, I think it is not necessary, it should be clear enough, just throwing ideals, let me close this comment.

PR #586, opened 2022-02-25
Rename `BA2026.EnableAdditionalSdlSecurityChecks`

marmegh:
Fix #584.

`EnableAdditionalSdlSecurityChecks` is unnecessarily vague. Renaming to `EnableMicrosoftCompilerSdlSwitch` to increase clarity for customers.

PR #590, opened 2022-02-28
Simplifying baseline

eddynaka:
This change is simplifying our tests, removing from the output SARIF all results related to the `NotApplicable` kind.

This will help us validate PRs like https://github.com/microsoft/binskim/pull/586, where we should only see changes in files that contains that kind of rule/result.


marmegh (approved):
:shipit:

src/Test.FunctionalTests.BinSkim.Driver/BaselineTests.cs
```diff
@@ -110,8 +110,8 @@ private void RunRules(StringBuilder sb, string inputFileName)
                 SarifOutputVersion = Sarif.SarifVersion.Current,
                 TargetFileSpecifiers = new string[] { inputFileName },
                 Traces = Array.Empty<string>(),
-                Level = new List<FailureLevel> { FailureLevel.Error, FailureLevel.Warning, FailureLevel.Note, FailureLevel.None },
```
michaelcfanning, line 113: >FailureLevel.None [](http://example.com/codeflow?start=110&length=17)

So, why did you remove none? Please explain. Either this value is useless (defined as zero) or we're missing test cases. I *think* the SARIF spec has a specific meaning for none (to help with debug messages)...

michaelcfanning: I checked the spec, you can remove this from now. Please read the spec on this area.

michaelcfanning:
This change also baselines macho files.

---
In reply to: [895377195](https://github.com/microsoft/binskim/pull/590#issuecomment-895377195) [](http://example.com/codeflow?ancestors=895377195,895377195)

michaelcfanning (changes requested):
:clock1:

michaelcfanning:
I can't provide a visual validation of your change. Please write some code to verify what you've done. Write two visitors that compare old and new log files and which insure programmatically that only not applicable changes are going away.

---
In reply to: [1054805808](https://github.com/microsoft/binskim/pull/590#issuecomment-1054805808) [](http://example.com/codeflow?ancestors=1054805808,895377195,895377195)

michaelcfanning (approved):
:shipit:

PR #595, opened 2022-03-05
Follow-up changes from Rust PR

shaopeng-gh:
per comments:
1. add rust binary to baseline, after the simplify baseline pr
2. rename FunctionalTestsData to remove s
3. replace empty result with with new one that have result.


shaopeng-gh:
per comment this is hard to review
will split into separated prs.

PR #597, opened 2022-03-05
Remove s from FunctionalTestsData and BaselineTestsData

shaopeng-gh:
# Description:

The **ONLY** change in this PR is to: 
Remove s from FunctionalTestsData and BaselineTestsData


docs/FunctionalTestBuildScripts.md
```diff
@@ -2,7 +2,7 @@
 
 This file records scripts used to compile the test files, in alphabetical order.
```
shaopeng-gh, line 3: >This [](http://example.com/codeflow?start=0&length=4)

Both CodeFlow and GitHub does not show all files for me... may be a max number of files limit?

this is the commit:

[https://github.com/microsoft/binskim/commit/547e02f2a9e6dd819400a3d7a49d7438bf0fc614](https://github.com/microsoft/binskim/commit/547e02f2a9e6dd819400a3d7a49d7438bf0fc614)

if you scroll to last file it is
"src/Test.UnitTests.BinaryParsers/PEBinary/PETests.cs"
but not showing here for me.

michaelcfanning (changes requested):
:clock1:

michaelcfanning:
You don't need to make this change. The note you received the other day was to update the name of a directory named 'TestsData' to 'TestData', following other directories in the repo. Please note there is a directory named 'TestsData', I think in the BinaryParsers directory that needs to be fixed.

shaopeng-gh:
got it, let me also change the "TestsData" folder in the "Test.UnitTests.BinaryParsers"

michaelcfanning:
I changed my mind and merged this PR. :) Although we didn't ask for the change, I think the consistency is good. So just start a different change for the 'TestsData' problem :)

---
In reply to: [1060053412](https://github.com/microsoft/binskim/pull/597#issuecomment-1060053412) [](http://example.com/codeflow?ancestors=1060053412,1060053412)

shaopeng-gh:
sounds good, will do that.

---
In reply to: [1060053808](https://github.com/microsoft/binskim/pull/597#issuecomment-1060053808) [](http://example.com/codeflow?ancestors=1060053808,1060053412)

shaopeng-gh:
sorry this change itself is incomplete, there are a few files not in solution but in the repo, let me fix that first.

---
In reply to: [1060054238](https://github.com/microsoft/binskim/pull/597#issuecomment-1060054238) [](http://example.com/codeflow?ancestors=1060054238,1060053412)

shaopeng-gh:
1. This is ready, missed a file RuleContributions.md under doc.
2. if we want to check in this pr, I will also create a PR for **SDK**, because this is in that repo as well:

image

shaopeng-gh:
Oh btw auto merge will not trigger unless status is changed from "Changes requested" to "Approved".

michaelcfanning (approved):
:shipit:

PR #602, opened 2022-03-07
Replace empty result baseline file

shaopeng-gh:
# Description:

This is just a follow up check in for:
Replace empty result baseline file with new ones that have result.


shaopeng-gh:
> {

Just found that we also check in the files in "Actual"  folder. I think it is not needed?

Is this by design, let me know if want me to create a PR to exclude files under "Actual" folder from checking in?



---
Refers to: src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Actual/clang.object_file.o.sarif:1 in 6a89d55. [](commit_id = 6a89d55fd98d7cbab2cb5230690206fdc87d4e44, deletion_comment = True)

michaelcfanning:
> {

Good catch! Please remove those! 

We should add an entry to our .gitignore to prevent this from happening again...

---
In reply to: [1060128720](https://github.com/microsoft/binskim/pull/602#issuecomment-1060128720) [](http://example.com/codeflow?ancestors=1060128720,1060128720)

---
Refers to: src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Actual/clang.object_file.o.sarif:1 in 6a89d55. [](commit_id = 6a89d55fd98d7cbab2cb5230690206fdc87d4e44, deletion_comment = True)

michaelcfanning (approved):
:shipit:

PR #604, opened 2022-03-09
Fix and enable Linux test for Binskim

shaopeng-gh:
# Description:

Fix and enable Linux test for Binskim


shaopeng-gh:
the reason is we only have x64 in the config

image

michaelcfanning (approved):
:shipit:

PR #610, opened 2022-03-10
Fix baseline tests for Linux and update nonwindowsexpected folder

shaopeng-gh:
# Description:

The task is to understand how "nonwindowsexpected" folder works, and if it works, fix if there is problem, and understand what we should do about this folder when checking in a new bin in baseline.

# Fix:

1. Fix and enable Baseline test in Linux
2. Fix the uri invalid per [[RFC3986](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#RFC3986)] in Linux
3. clean all, and re-generate "nonwindowsexpected" folder after fixes.
4. When adding a new bin, we need to run `./UpdateBaselines.sh` in Linux and checkin the SARIF file in nonwindowsexpected folder.


shaopeng-gh:
>         $TOOLPATH analyze $targetFile --output $outputTemp --kind 'Fail;Pass' --level Error`;Warning`;Note --insert Hashes --remove NondeterministicProperties --config default --quiet --sarif-output-version Current

this change is simple copy the latest parameters we are  currently using from the Windows version to make them in sync.

---
Refers to: src/Test.FunctionalTests.BinSkim.Driver/UpdateBaselines.sh:31 in faaa87e. [](commit_id = faaa87edb368b8eda895fb9c66074bdc7cccdd83, deletion_comment = False)

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/NonWindowsExpected/Binskim.linux-x64.dll.sarif
```diff
@@ -329,7 +41,7 @@
             {
               "physicalLocation": {
                 "artifactLocation": {
-                  "uri": "file://Z:/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Binskim.linux-x64.dll",
```
shaopeng-gh, line 332: >file://Z: [](http://example.com/codeflow?start=26&length=9)

before fix, the uri is invalid per [RFC3986]

michaelcfanning (approved):
:shipit:

michaelcfanning:
AWESOME CHANGE. :)

eddynaka:
I would also suggest to update the docs showing how to update the tests for unix. Since we might need WSL or an actual unix environment (?)

shaopeng-gh:
> I would also suggest to update the docs showing how to update the tests for unix. Since we might need WSL or an actual unix environment (?)

Sounds good,
I found a place `\docs\RuleContributions.md` 
contains: `6. Review impact of new analysis on test set and update baselines if necessary.`
I will create a section `# How to update baseline` in it.

PR #618, opened 2022-03-23
codeQL fixes

marmegh:
The codeQL yml file had a type mismatch. Added this file and others in the folder to solution items to aid in adhering to team standard for validating file changes in the IDE.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -2,7 +2,7 @@ name: CodeQL Analysis
 
 on:
   push:
-    branches: "*"
+    branches: [ "**" ]
   pull_request:
     branches: [ main ]
   schedule:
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -2,7 +2,7 @@ name: CodeQL Analysis
 
 on:
   push:
-    branches: "*"
+    branches: [ "**" ]
```
marmegh, line 5: >branches: [ "**" ] [](http://example.com/codeflow?start=4&length=18)

Opened this file in Visual Studio and this section was highlighted with an error message that the expected type should be an array. Also added the change recommended to ensure branches with slashes are included.

eddynaka: GH accepts both ways. so, either are fine.

PR #621, opened 2022-03-24
CodeQL Fix "This cast is redundant because the expression already has type"

shaopeng-gh:
# Description:

CodeQL reports "This cast is redundant because the expression already has type"

# Fixes:

remove the redundant cast

# Tests:

There is No tests added because there is no behavioral implications of this specific type of  change "cast is redundant".

### DIFF
--- a/src/BinaryParsers/ElfBinary/Dwarf/DwarfCompilationUnit.cs
+++ b/src/BinaryParsers/ElfBinary/Dwarf/DwarfCompilationUnit.cs
@@ -170,7 +170,7 @@ namespace Microsoft.CodeAnalysis.BinaryParsers.Dwarf
 
                         case DwarfFormat.Data8:
                             attributeValue.Type = DwarfAttributeValueType.Constant;
-                            attributeValue.Value = (ulong)debugData.ReadUlong();
+                            attributeValue.Value = debugData.ReadUlong();
                             break;
 
                         case DwarfFormat.SData:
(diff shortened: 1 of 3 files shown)


shaopeng-gh:
This is the same type of fix in SARIF SDK, with tests are missing and did not mention the reason.
Just added in the top description why tests are not authored together with the change.

shaopeng-gh:
fyi the Code QL shows a note type:
"Note": Local scope variable shadows member
which is not new to this change, I do not include in this fix since my PR is batch fix "cast is redundant" only.

eddynaka:
Please, create a follow-up PR fixing that.

---
In reply to: [1078422049](https://github.com/microsoft/binskim/pull/621#issuecomment-1078422049) [](http://example.com/codeflow?ancestors=1078422049,1078422049)

eddynaka:
Do we have tests that covers those changes? If not, can you create some?

michaelcfanning:
Q: do these changes introduce a change in runtime behavior?
A: No.

Do you understand why? This particular analysis is a code clean-up analysis, the devs have introduced casts that are unnecessary. Removing them leaves the code in an identical condition after compilation, it's just that the source code is cleaner.

Make sense? This is why we also might not worry about tests if a change only consisted of removing unnecessary namespaces.

---
In reply to: [1100269302](https://github.com/microsoft/binskim/pull/621#issuecomment-1100269302) [](http://example.com/codeflow?ancestors=1100269302,1100269302)

michaelcfanning (approved):
:shipit:

PR #629, opened 2022-04-11
Updating CodeQL analysis for Dependabot on push

eddynaka:
Dependabot does not have access to "push" and it fails some actions.
With that in mind, the known way to fix this is to disable Dependabot branches from the analysis:

https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/troubleshooting-the-codeql-workflow#error-403-resource-not-accessible-by-integration-when-using-dependabot


michaelcfanning (approved):
:shipit:

PR #652, opened 2022-06-07
Fix missing command line id from compiler events

marmegh:
While reviewing compiler data and command line flags, noticed there was not a means to link command line events to compiler data events. A review of code history identified a regression that caused the compiler event data to be emitted prior to creation of the command line id, thereby dropping the link between the two events.

This change corrects the regression by moving logging of the compiler event after the data chunking events. To prevent regression, implemented a more thorough test validation method.


src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -283,6 +281,9 @@ public void Write(BinaryAnalyzerContext context, CompilerData compilerData)
                                    "assemblyReferences",
                                    compilerData.AssemblyReferences);
             }
+
+            // This event cannot be sent before AssemblyReferences or CommandLine.
```
michaelcfanning, line 285: >This [](http://example.com/codeflow?start=15&length=4)

Please explain why in the comment.

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
```diff
@@ -622,6 +625,135 @@ private void LogExceptionEvents(List<ExecutionException> exceptions, CompilerDat
...
+                        {
+                            commandLineEventsId = telemetryEvent.Properties["commandLineId"];
+                        }
+
+                        if (assemblyEventsId.Equals(string.Empty) && telemetryEvent.Properties.ContainsKey("assemblyReferencesId"))
+                        {
+                            assemblyEventsId = telemetryEvent.Properties["assemblyReferencesId"];
+                        }
+
+                        if (expectedEventName == CompilerDataLogger.AssemblyReferencesEventName && !telemetryEvent.Properties.ContainsKey("assemblyReferencesId"))
```
michaelcfanning, line 659: >AssemblyReferencesEventName [](http://example.com/codeflow?start=68&length=27)

This line is getting very long, consider dropping your second condition in order to improve readability, here and elsewhere. I might actually prefer to do this everywhere to make the convention consistent.

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
```diff
@@ -622,6 +625,135 @@ private void LogExceptionEvents(List<ExecutionException> exceptions, CompilerDat
...
+            var summaryEvents = new List<EventTelemetry>();
+
+            foreach (EventTelemetry telemetryEvent in telemetryGeneratedEvents)
+            {
+                switch (telemetryEvent.Name)
+                {
+                    case CompilerDataLogger.CompilerEventName:
+                        compilerEvents.Add(telemetryEvent);
+
+                        if (commandLineEventsId.Equals(string.Empty) && telemetryEvent.Properties.ContainsKey("commandLineId"))
```
michaelcfanning, line 649: >telemetryEvent [](http://example.com/codeflow?start=72&length=14)

You perform a map look up twice in your code, once for ContainsKey, the second to actually retrieve the value.

Try using a TryGetValue pattern instead. You can also inline a variable for this to assign to using new C# syntax:

if (commandLineEvents.IdEquals(string.Empty) &&
    telemetryEventProperties.TryGetValue("commandLineId", out string eventId))
{
    commandLineEventsId = eventId;
}

michaelcfanning, line 649: >commandLineEventsId.Equals [](http://example.com/codeflow?start=28&length=26)

Why is a null value ok here? Any time you are checking string values in this way, we always need to consider at least three conditions and whether they make sense, null, the empty string, and all whitespace.

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
```diff
@@ -622,6 +625,135 @@ private void LogExceptionEvents(List<ExecutionException> exceptions, CompilerDat
...
+
+                    case CompilerDataLogger.AssemblyReferencesEventName:
+                        assemblyReferencesEvents.Add(telemetryEvent);
+
+                        if (expectedEventName == CompilerDataLogger.AssemblyReferencesEventName
+                            && !expectedChunkedContent.Contains(telemetryEvent.Properties["chunkedassemblyReferences"]))
+                        {
+                            sb.AppendLine(string.Format("Unexpected {0} chunked content: {1}",
+                                CompilerDataLogger.AssemblyReferencesEventName,
+                                telemetryEvent.Properties[$"{CompilerDataLogger.AssemblyReferencesEventName}Id"]));
```
michaelcfanning, line 677: >CompilerDataLogger.AssemblyReferencesEventName [](http://example.com/codeflow?start=61&length=46)

Why are you using this property name construction technique here but not elsewhere? Should 'assemblyReferencesId' and other constants of this kind be declared as constants, rather than embedded literals?

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
```diff
@@ -622,6 +625,135 @@ private void LogExceptionEvents(List<ExecutionException> exceptions, CompilerDat
...
+            var compilerEvents = new List<EventTelemetry>();
+            var assemblyReferencesEvents = new List<EventTelemetry>();
+            var commandLineEvents = new List<EventTelemetry>();
+            var summaryEvents = new List<EventTelemetry>();
+
+            foreach (EventTelemetry telemetryEvent in telemetryGeneratedEvents)
+            {
+                switch (telemetryEvent.Name)
+                {
+                    case CompilerDataLogger.CompilerEventName:
```
michaelcfanning, line 646: >CompilerDataLogger.CompilerEventName [](http://example.com/codeflow?start=25&length=36)

Style nit: please add braces to your case blocks. The indentation of the actual case contents is correct.

michaelcfanning:
You didn't describe the actual problem resolved in this PR in your description. You didn't describe how we discovered the problem, which can often be helpful information (e.g., it is helpful to know whether we're blocking customers or independently identified the issue). You did provide important details on the fix but I don't understand the change complete because the problem itself wasn't communicated. I assume it's because we fail to generate/associate an ID properly with the data we're transmitting? If not correct, can you fill me in? Thanks!

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
```diff
@@ -622,6 +625,135 @@ private void LogExceptionEvents(List<ExecutionException> exceptions, CompilerDat
...
+
+                    case CompilerDataLogger.SummaryEventName:
+                        summaryEvents.Add(telemetryEvent);
+                        break;
+                }
+            }
+
+            if (expectedEventName == CompilerDataLogger.AssemblyReferencesEventName && assemblyReferencesEvents.Count != chunkNumber)
+            {
+                sb.AppendLine(string.Format("Expected {0} {1} events, but {2} were found.",
```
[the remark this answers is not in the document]
marmegh: Actually, these two can be combined and we're still able to denote a difference between the two. Consolidated both and cleared things up a bit.

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
```diff
@@ -622,6 +631,139 @@ private void LogExceptionEvents(List<ExecutionException> exceptions, CompilerDat
...
+            var compilerEvents = new List<EventTelemetry>();
+            var assemblyReferencesEvents = new List<EventTelemetry>();
+            var commandLineEvents = new List<EventTelemetry>();
+            var summaryEvents = new List<EventTelemetry>();
+
+            foreach (EventTelemetry telemetryEvent in telemetryGeneratedEvents)
+            {
+                IDictionary<string, string> properties = telemetryEvent.Properties;
+
+
```
shaopeng-gh, line 652: there are 2 empty lines here, should be 1 only. I thought linter should mark it?

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
```diff
@@ -622,6 +631,139 @@ private void LogExceptionEvents(List<ExecutionException> exceptions, CompilerDat
...
+                        if (expectedEventName == CompilerDataLogger.AssemblyReferencesEventName
+                            && !properties.ContainsKey(CompilerDataLogger.AssemblyReferencesId))
+                        {
+                            sb.AppendLine("Compiler Event is missing the `assemblyReferencesId`");
+                        }
+                        else if (expectedEventName == CompilerDataLogger.CommandLineEventName
+                            && !properties.ContainsKey(CompilerDataLogger.CommandLineId))
+                        {
+                            sb.AppendLine("Compiler Event is missing the `commandLineId`");
```
shaopeng-gh, line 679: >`commandLineId` [](http://example.com/codeflow?start=73&length=15)

I see the comment below about:
$"Expected `{commandLineEventsId}` but found `{currentCommandLineEventId}`."

See if we want to do the same here.

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
shaopeng-gh: Most other language, linters will flag. not in .NET.
:)

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
```diff
@@ -622,6 +631,139 @@ private void LogExceptionEvents(List<ExecutionException> exceptions, CompilerDat
...
+                        {
+                            sb.AppendLine(
+                                $"Unexpected `{CompilerDataLogger.AssemblyReferencesEventName}` chunked content: " +
+                                $"`{properties[CompilerDataLogger.ChunkedAssemblyReferences]}` " +
+                                $"expected: `{expectedChunkedContent}`");
+                        }
+
+                        string currentAssemblyEventId = properties["assemblyReferencesId"];
+
+                        if (assemblyEventsId != properties["assemblyReferencesId"])
```
shaopeng-gh, line 698: >if (assemblyEventsId != properties["assemblyReferencesId"]) [](http://example.com/codeflow?start=24&length=59)

defined currentAssemblyEventId right before this line,
were you planning to use it here:
if (assemblyEventsId != currentAssemblyEventId)

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
```diff
@@ -622,6 +631,139 @@ private void LogExceptionEvents(List<ExecutionException> exceptions, CompilerDat
...
+                            sb.AppendLine("Compiler Event is missing the `commandLineId`");
+                        }
+                        break;
+                    }
+                    case CompilerDataLogger.AssemblyReferencesEventName:
+                    {
+                        assemblyReferencesEvents.Add(telemetryEvent);
+
+                        if (expectedEventName == CompilerDataLogger.AssemblyReferencesEventName
+                            && !expectedChunkedContent.Contains(properties[CompilerDataLogger.ChunkedAssemblyReferences]))
```
shaopeng-gh, line 688: >&& !expectedChunkedContent.Contains(properties[CompilerDataLogger.ChunkedAssemblyReferences])) [](http://example.com/codeflow?start=28&length=94)

Have not check this file before, I ran the test case locally and see the behavior,

expectedChunkedContent:
"Microsoft.DiaSymReader (1.3.0);Newtonsoft.Json (13.0.1)"

properties[CompilerDataLogger.ChunkedAssemblyReferences]:
"DiaSymRead"

This condition will pass the contains will be true, 
wondering why it is "DiaSymRead" not "DiaSymReader" or "Microsoft.DiaSymReader (1.3.0)"
do you know,

shaopeng-gh: ok! I see now, DiaSymRead is split from the whole string and we are comparing using contains.

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
```diff
@@ -622,6 +631,139 @@ private void LogExceptionEvents(List<ExecutionException> exceptions, CompilerDat
...
+                IDictionary<string, string> properties = telemetryEvent.Properties;
+
+
+                if (string.IsNullOrWhiteSpace(commandLineEventsId)
+                    && properties.TryGetValue(CompilerDataLogger.CommandLineId, out string commandLineEventId))
+                {
+                    commandLineEventsId = commandLineEventId;
+                }
+
+                if (string.IsNullOrWhiteSpace(assemblyEventsId)
```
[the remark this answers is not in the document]
marmegh: The Id should be the same for each test. We have a check later to confirm that the ids remain the same.

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
marmegh: Good call.

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
marmegh: Did the same for the command line event too.

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
```diff
@@ -622,6 +625,135 @@ private void LogExceptionEvents(List<ExecutionException> exceptions, CompilerDat
...
+                        }
+
+                        string currentCommandLineEventId = telemetryEvent.Properties["commandLineId"];
+                        if (assemblyEventsId.Equals(string.Empty))
+                        {
+                            commandLineEventsId = currentCommandLineEventId;
+                        }
+                        else if (assemblyEventsId != telemetryEvent.Properties["commandLineId"])
+                        {
+                            sb.AppendLine(string.Format("{0} event detected with unexpected Id. Expected {1}, but found {2}",
```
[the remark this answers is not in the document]
marmegh: The missing space is why I tend to prefer the string.format approach; but opted to take the full suggestion to clear up confusion over where we're including vs. not including the event name.

marmegh:
I've updated the description with more context.

---
In reply to: [1149142777](https://github.com/microsoft/binskim/pull/652#issuecomment-1149142777) [](http://example.com/codeflow?ancestors=1149142777,1149142777)

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
shaopeng-gh: missing the {} after fix:
"Compiler Event is missing the `commandLineId`"

shaopeng-gh (approved):
:shipit:

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
```diff
@@ -622,6 +631,154 @@ private void LogExceptionEvents(List<ExecutionException> exceptions, CompilerDat
...
+
+        private StringBuilder ValidateChunkedContent(StringBuilder sb,
+                                                     int expectedChunkSize,
+                                                     List<EventTelemetry> chunkedEvents)
+        {
+            if (chunkedEvents.Count != expectedChunkSize)
+            {
+                sb.AppendLine(
+                    string.Format(
+                        "Incorrect number of chunkedEvents de tected. Expected {0}, but found {1}",
```
michaelcfanning, line 774: >de tected. [](http://example.com/codeflow?start=59&length=10)

Is there an erroneous space in 'de tected'? :) You can use interpolated strings here as well.

I'm approving this change but if you have a moment to go fix these nits, do that. :)

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -260,29 +266,31 @@ public void Write(BinaryAnalyzerContext context, CompilerData compilerData)
...
+                properties.Add(AssemblyReferencesId, assemblyReferencesId);
 
                 SendChunkedContent(AssemblyReferencesEventName,
                                    assemblyReferencesId,
-                                   "assemblyReferences",
+                                   AssemblyReferences,
                                    compilerData.AssemblyReferences);
             }
+
+            // To prevent loss of relationship between compiler and assembly/commandline data
```
michaelcfanning, line 291: >prevent [](http://example.com/codeflow?start=18&length=7)

This comment would be clearer if it explained how 'loss of relationship' is manifest in the actual data stream. I assume it's because we have no generated ID/guid? can you add just a bit more detail?

michaelcfanning (approved):
:shipit:

PR #657, opened 2022-06-23
[WIP] Add BA2027.EnableSourceLink

### LINKED ISSUE #656: [RULE REQUEST] Warn when SourceLink information is missing
# Rule Proposal: SourceLink presence

#### Rule metadata
- *Id*: BA?????
- *Name*: EnableSourceLink

#### Rule descriptions
- *Short*: SourceLink information should be present in the PDB.
- *Full*: SourceLink information should be present in Windows or Portable PDBs for binaries compiled with the C# or MSVC compilers.

#### Platform & applicability
- *Platform*: Windows and Linux
- *Applicable to*: C#, MSVC. Windows or Portable PDBs.
- *Not applicable to*: Other languages or compilers (e.g. clang, gcc)

#### User-facing strings
- *Fail*: '{0}' does not contain SourceLink information.
- *Pass*:  '{0}' contains SourceLink information.
- *[Other]*:

https://aka.ms/sourcelink

#### Implementation Notes
Given a binary (x.dll), the implementation needs to locate the associated PDB. The PDB may be either a "full" Windows PDB or a Portable PDB. (Embedded PDBs also work.)
For Windows PDBs, the SourceLink document can be retrieved from ISymUnmanagedReader5.GetSourceServerData.
For portable PDBs, the SourceLink document is stored as "custom debug information" with the GUID CC110556-A091-4D38-9FEC-25AB9A351A6A ([SourceLinkKind](https://github.com/dotnet/runtime/blob/main/docs/design/specs/PortablePdb-Metadata.md#source-link-c-and-vb-compilers))
Once you've retrieved the document, it can be parsed as UTF-8 JSON.
For the purposes of this test, it's probably sufficient to just check for the presence of valid JSON.
For a more complete check, we could parse it using SourceLinkMap from [Microsoft.SourceLink.Tools](https://github.com/dotnet/sourcelink).

#### How to resolve
For C# projects:
1. Use .NET Core SDK version 2.1.300 or newer to build your project.
1. Ensure that DebugType is set to Portable or Full.
1. Include the appropriate SourceLink nuget package. See https://github.com/dotnet/sourcelink#using-source-link-in-net-projects

For MSVC projects:
1. Use the MSVC compiler from Visual Studio 2017 Update 9 or newer. The linker needs to support the [/sourcelink switch](https://docs.microsoft.com/en-us/cpp/build/reference/sourcelink?view=msvc-170).
1. Add package references to the appropriate SourceLink provider: See https://github.com/dotnet/sourcelink#using-source-link-in-c-projects

pharring:
Fixes #656

A few notes:

The version of Dia2Lib that we use is a bit out of date. I need IDiaDataSource3 to get access to the raw stream data, so I redefined it locally for now.

I created a SourceLink.sln under src/BuildSamples to generate combinations of C#/C++ with/without SourceLink. The outputs of these builds are then checked in under FunctionalTestData. I also grabbed a couple of existing clang/gcc/macho binaries to test the "NotApplicable" code path.


src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -814,6 +817,17 @@
                 : ChecksumAlgorithmForPortablePdb();
         }
 
+        public string ManagedPdbGetSourceLinkDocument()
+        {
+            if (!TryGetPortablePdbMetadataReader(out MetadataReader metadataReader))
```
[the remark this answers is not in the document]
pharring: Addressed in https://github.com/microsoft/binskim/pull/657/commits/47fdbd0d019f37cbdb257bad91a453d90674f3bc

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -831,16 +845,11 @@
 
         private ChecksumAlgorithmType ChecksumAlgorithmForPortablePdb()
         {
-            if (!this.peReader.TryOpenAssociatedPortablePdb(
-                this.FileName,
-                filePath => File.Exists(filePath) ? File.OpenRead(filePath) : null,
-                out MetadataReaderProvider pdbProvider,
-                out _))
-            {
+            if (!TryGetPortablePdbMetadataReader(out MetadataReader metadataReader))
```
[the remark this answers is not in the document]
pharring: Addressed in https://github.com/microsoft/binskim/pull/657/commits/47fdbd0d019f37cbdb257bad91a453d90674f3bc

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -853,6 +862,22 @@
             return ChecksumAlgorithmType.Unknown;
         }
 
+        private bool TryGetPortablePdbMetadataReader(out MetadataReader metadataReader)
```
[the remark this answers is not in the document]
pharring: Addressed in https://github.com/microsoft/binskim/pull/657/commits/47fdbd0d019f37cbdb257bad91a453d90674f3bc

marmegh:
I see you added test binaries to the functional test folder. You'll want to run `UpdateBaselines.cmd` to automatically generate/update a SARIF file for each test binary including the new rule. Make sure to run BuildAndTest before and after executing this step. It is likely that some tests may no longer be up to date with the rule addition.

pharring:
1. Need to update baselines.
2. Need to fix unit tests.

src/BinaryParsers/PEBinary/ProgramDatabase/Pdb.cs
```diff
@@ -396,6 +399,11 @@ public void Dispose()
             {
                 Marshal.ReleaseComObject(this.session);
             }
+
+            if (this.dataSource != null)
+            {
+                Marshal.ReleaseComObject(this.dataSource);
```
michaelcfanning, line 405: >ReleaseComObject [](http://example.com/codeflow?start=24&length=16)

thanks for cleaning this up.

docs/BinSkimRules.md
```diff
@@ -870,3 +870,21 @@ This rule emits CSV data to the console for every compiler/language/version comb
...
+
+### Messages
+
+#### `Pass`: Pass
+
+The PDB for '{0}' contains SourceLink information.
+
+#### `Error`: Error
+
+The PDB for '{0}' does not contain SourceLink information.
```
michaelcfanning, line 887: >The PDB for '{0}' does not contain SourceLink information [](http://example.com/codeflow?start=0&length=57)

Let's build out this description a tiny bit, see below for a start. What support, exactly, should we document for C++? Does SourceLink have a seamless interop story with existing source server tech, or does it replace it? Does it provide value because we're updating symbol servers to be source link aware? Just trying to clarify my understanding further in support of building a concise desc that answers some typical questions.

'SourceLink information should be present in the PDB of all binaries built with the C# and MSVC compilers. For .NET assemblies. Source Link executes when creating a NuGet package and embeds source control metadata inside assemblies and the package. Developers who download the package and have Source Link enabled in Visual Studio can step into its source code. Frictionless source-driven debugging provides a good user experience for dependency consumers and also accelerates security response in the  event of supply-chain compromise. See [https://aka.ms/sourcelink](https://aka.ms/sourcelink) for more information.'

michaelcfanning, line 887: >The [](http://example.com/codeflow?start=0&length=3)

Our user-facing strings typically publish three data points, 1) the condition that prompted us to fire, 2) the risks associated with ignoring the result, 3) potential fixes.

We've got #1. I suppose the risks are increased engineering and security response costs when source code is not available for debugging. 

ERROR:
The PDB for '{0}' does not contain SourceLink information, increasing engineering and security response costs when source code isn't easily located for debugging; see [https://aka.ms/sourcelink](https://aka.ms/sourcelink) and [https://docs.microsoft.com/en-us/cpp/build/reference/sourcelink](https://docs.microsoft.com/en-us/cpp/build/reference/sourcelink) for information on how to enable.

Our pass messages need to broadcast core benefit:
PASS:

The PDB for '{0}' contains SourceLink information, maximizing engineering and security response cost efficiency when source code is required for debugging and other anlaysis.

src/BinSkim.Rules/RuleIds.cs
```diff
@@ -69,5 +69,8 @@ internal static class RuleIds
         // MachO Checks
         public const string EnablePositionIndependentExecutableMachO = "BA5001";
         public const string DoNotAllowExecutableStack = "BA5002";
+
+        // SourceLink Checks
```
michaelcfanning, line 73: >SourceLink [](http://example.com/codeflow?start=11&length=10)

Ok, so the rule id is always tricky. Mostly our rule IDs are organized around program + debug data representation, BA2XXX are for Windows and BA3XXX are for *nix. Also, these domains currently are focused entirely on security.

Is SourceLink applicable only to MS-generated binaries? Is this data relevant for *nix platforms, say for .NET core? Will *nix or others build their own similar features for native *nix formats + package managers?

This feature does, in fact, have a security element to it: the inability to acquire source code for a compromised (often old) component is a real problem. It's why standards such as the SDL recommend archiving PDBs: persisting/accessing relevant source is an extension of this value.

So we could simply extend the current rule set and make this BA2027.

On the other hand, we could consider this rule the start of a new class of BinSkim analysis, which is more oriented towards general goodness, enabling debugging, performance optimizations, etc.

src/BinSkim.Rules/PERules/BA6001.EnableSourceLink.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                context.Logger.Log(this,
+                    RuleUtilities.BuildResult(ResultKind.Pass, context, null,
+                    nameof(RuleResources.BA6001_Pass),
+                    context.TargetUri.GetFileName()));
+            }
+            else
+            {
+                context.Logger.Log(this,
+                    RuleUtilities.BuildResult(ResultKind.Fail, context, null,
```
michaelcfanning, line 61: >BuildResult [](http://example.com/codeflow?start=34&length=11)

So for the failure case, you don't pass ResultKind.Fail, you call a different overload that accepts the failure severity. For this check, we need a default severity of 'warning', as 'error' in our domain is a serious 'must fix' designation.

Another option for this rule is to make it disabled by default, and require people to opt in. I tend to think we should try to make the 'warning' error level work.

RuleUtilities.BuildResult(FailureLevel.Warning

src/BinSkim.Rules/RuleIds.cs
```diff
@@ -69,5 +69,8 @@ internal static class RuleIds
...
+        public const string EnableSourceLink = "BA6001";
```
michaelcfanning, line 74: >EnableSourceLink [](http://example.com/codeflow?start=28&length=16)

Let's go ahead and proceed with BA2027 and assume the security connection is sufficiently obvious to be in our core charter. :) As I believe it is. Efficient security response is important, it is a truism in our space that attackers have all the time in the world and, when something finally pops, every cycle counts for response engineers.

The primary/happy path value here, of course, is simply making API/dependency consumers happy, i.e., this change maximizes eco-system satisfaction/growth.

src/BuildSamples/SourceLink/Directory.Build.targets
```diff
@@ -0,0 +1,22 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!-- Copy the output assembly and its PDB to the appropriate Pass/Fail folder. -->
+  <PropertyGroup>
+    <!-- EnableSourceLink is set in Microsoft.SourceLink.Common.props -->
```
michaelcfanning, line 6: >props [](http://example.com/codeflow?start=64&length=5)

Just curious, and this target file will naturally be on the path for certain modern VS version users?

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/Binskim.linux-x64.dll.sarif
```diff
@@ -218,7 +218,28 @@
...
+                  "physicalLocation": {
+                    "artifactLocation": {
+                      "uri": "file:///Z:/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Binskim.linux-x64.dll",
+                      "index": 0
+                    }
+                  }
+                }
+              ],
+              "message": {
+                "text": "'Binskim.linux-x64.dll' was not evaluated because its PDB could not be loaded (E_PDB_FORMAT)."
```
michaelcfanning, line 234: >PDB could not be loaded [](http://example.com/codeflow?start=79&length=23)

In general, any transition you see from 'executionSuccessful' == true to the appearance of error notifications will be a usability problem. 

We need to chase down these 'PDB can't be loaded' issues. 

Several  high-level matters will be helpful to settle: 1) you need to make sure your PDB loading only occurs when the target is valid for your analysis, 2) you need to make sure your analysis only occurs if you observe the binary is built with a toolchain that provides source-link support.

i.e., we don't generally push for things like 'you need to update your stale compiler toolchain so you can enable sourcelink'. The pressure to upgrade is handled in the 'BuildWithSecureTools' rule. Assuming that rule is enforcing a min-version that also enables SourceLink, all we need to do is update that rule docs to mention SourceLink as yet another motivation to upgrade your compiler.

You may discover the min-version that rule enforces is too early for SourceLink support, in which case we should consider lifting it.

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/ManagedInteropAssemblyForAtlTestLibrary.dll.sarif
```diff
@@ -315,7 +315,28 @@
       },
       "invocations": [
         {
-          "executionSuccessful": true
+          "toolConfigurationNotifications": [
```
michaelcfanning, line 318: >toolConfigurationNotifications [](http://example.com/codeflow?start=11&length=30)

This notification is a concern because it suggests we used to successfully analyze this binary but now cannot, because we are (for the first time) attempting to location a PDB for it. This could indicate that your rule's CanAnalyze needs to be filtering something it is not. In any case, the new appearance of E_PDB_NO_DEBUG_INFO is probably something we shouldn't see in your baseline updates.

src/BinSkim.Rules/PERules/BA6001.EnableSourceLink.cs
```diff
@@ -0,0 +1,88 @@
...
+        private const string AnalyzerName = RuleIds.EnableSourceLink + "." + nameof(EnableSourceLink);
+
+        public override void Initialize(BinaryAnalyzerContext context)
+        {
+            base.Initialize(context);
+        }
+
+        public override AnalysisApplicability CanAnalyzePE(PEBinary target, PropertiesDictionary policy, out string reasonForNotAnalyzing)
+        {
+            reasonForNotAnalyzing = null;
```
michaelcfanning, line 45: >reasonForNotAnalyzing [](http://example.com/codeflow?start=12&length=21)

Ah, here we are. There are many N/A conditions you probably should be honoring. I think. :) For example, do we require source code in a resource-only binary? Or any case in which a PDB (if one exists) might not have source records? Is that possible?

src/BinSkim.Rules/PERules/BA6001.EnableSourceLink.cs
```diff
@@ -0,0 +1,88 @@
...
+
+using Microsoft.CodeAnalysis.BinaryParsers;
+using Microsoft.CodeAnalysis.BinaryParsers.ProgramDatabase;
+using Microsoft.CodeAnalysis.IL.Sdk;
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.Driver;
+
+namespace Microsoft.CodeAnalysis.IL.Rules
+{
+    [Export(typeof(Skimmer<BinaryAnalyzerContext>)), Export(typeof(ReportingDescriptor)), Export(typeof(IOptionsProvider))]
```
michaelcfanning, line 17: >IOptionsProvider [](http://example.com/codeflow?start=104&length=16)

Strictly speaking you don't need IOptionsProvider support because we have no knobs (yet). :) Have you thought about whether this check might benefit from some configurability? Nothing occurs to me yet, thought I'd ask. If we get to merge and your are still returning an empty array, let's zap your IOptionsProvider support. The reason is that binskim supports an export-config option that produces a default XML config file (to be tweaked by users to configure config-driven analysis). If you implement this interface, that command will find you and potentially record some data in the doc where none is expected.

btw - I may be wrong about that. As I recall, the export-config *always* has a record for every rule, so that you can configure turning it off or changing severity. So it may handle an empty array of options just fine. If you have a moment to check it out, let me know what you find. Isn't important.

src/BinSkim.Rules/PERules/BA6001.EnableSourceLink.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// </summary>
+        public override MultiformatMessageString FullDescription => new MultiformatMessageString { Text = RuleResources.BA6001_EnableSourceLink_Description };
+
+        protected override IEnumerable<string> MessageResourceNames => new string[]
+        {
+            nameof(RuleResources.BA6001_Pass),
+            nameof(RuleResources.BA6001_Error)
+        };
+
+        private const string AnalyzerName = RuleIds.EnableSourceLink + "." + nameof(EnableSourceLink);
```
michaelcfanning, line 36: >RuleIds.EnableSourceLink + "." + nameof(EnableSourceLink); [](http://example.com/codeflow?start=44&length=58)

If we established this code convention today, I think we'd use an interpolated string. Feel free to do that. We can go back later and clean everyone else up.

$"{RuleIds.EnableSourceLink}.{nameof(EnableSourceLink)};

src/BinSkim.Rules/PERules/BA6001.EnableSourceLink.cs
```diff
@@ -0,0 +1,88 @@
...
+        private static bool HasSourceLink(BinaryAnalyzerContext context)
+        {
+            PEBinary target = context.PEBinary();
+            Pdb pdb = target.Pdb;
+
+            // We're just checking for the presence of SourceLink document(s),
+            // not whether they can be read.
+            if (pdb.FileType == PdbFileType.Portable)
+            {
+                string sourceLinkDocument = target.PE.ManagedPdbGetSourceLinkDocument(pdb);
```
michaelcfanning, line 76: Since we don't actually read this document, is there a more efficient API we could create that simply notices where the reader has content? Looking at your helper, can we avoid the ReadUTF8 call and just return true because its Length is >0? That would save something wouldn't it?

return sourceLinkReader.Length == 0 ? null : sourceLinkReader.ReadUTF8(sourceLinkReader.Length);

src/BinSkim.Rules/PERules/BA6001.EnableSourceLink.cs
```diff
@@ -0,0 +1,88 @@
...
+                return !string.IsNullOrEmpty(sourceLinkDocument);
+            }
+            else
+            {
+                IEnumerable<string> sourceLinkDocuments = pdb.WindowsPdbGetSourceLinkDocuments();
+                return sourceLinkDocuments != null && sourceLinkDocuments.Any();
```
michaelcfanning, line 82: >Any [](http://example.com/codeflow?start=74&length=3)

Looking at your yield iterator implementation, looks like this code path is as efficient as it could be, i.e., breaks on the first evidence we've got support.

Just asking: is there no magic value, other (reliable) heuristic we can use to make this determination? I ask merely because as a rule, we try to avoid PDB parsing (and COM interop) wherever possible. Sniffing a magic file in a binary is admirably low-cost. :)

My sense is there's no opportunity here and we've got a reasonably efficient implementation (save the unnecessary presence of ReadUTF8 in the Portable code path, if I'm correct there's a fix we could make here).

src/BuildSamples/SourceLink/Directory.Build.targets
pharring: Yes. It's been so long that I don't remember when support for Directory.Build.props and Directory.Build.targets was added to MSBuild. Definitely there since VS 2017. Likely longer.

src/BinSkim.Rules/PERules/BA6001.EnableSourceLink.cs
pharring: Const string interpolation requires C# 10
I copied this from some other rule, but it turns out it isn't used and can be deleted.

docs/BinSkimRules.md
pharring: Done

pharring: Done.

src/BinSkim.Rules/RuleIds.cs
pharring: BA2027 it is

src/BinSkim.Rules/PERules/BA6001.EnableSourceLink.cs
pharring: Removed the IOptionsProvider

michaelcfanning (approved):
:shipit:

.gitignore
```diff
@@ -19,6 +19,44 @@ src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Actual/
...
+
+*_i.c
+*_p.c
+*_h.h
+*_i.h
+*.ilk
+*.meta
+*.obj
+*.pch
+*.pdb
```
shaopeng-gh, line 37: if we can use the folder ignore it would be better,

above:
`# Build results 
bld/
bin/
obj/`

this particular one *.pdb since we frequently checking in pdb files

docs/BinSkimRules.md
```diff
@@ -196,7 +196,7 @@ The BIND_NOW flag is missing from this binary, so relocation sections in '{0}' w
 
 ---
 
-## Rule `BA3030.UseGccCheckedFunctions`
+## Rule `BA3030.UseCheckedFunctionsWithGcc`
 
 ### Description
```
shaopeng-gh: >C# [](http://example.com/codeflow?start=93&length=2)

Does it support VB?

`Source Link is a language- and source-control agnostic system for providing first-class source debugging experiences for binaries...`

src/BinSkim.Rules/PERules/BA6001.EnableSourceLink.cs
shaopeng-gh: save the unnecessary presence of ReadUTF8 in the Portable code path ------ read form the test file i can see the linker commandline have the:
`/sourcelink`

 /ERRORREPORT:PROMPT /OUT:V:\T\binskim\src\BuildSamples\SourceLink\x64\Release\CPlusPlus_SourceLink.exe /NOLOGO /MANIFEST "/MANIFESTUAC:level='asInvoker' uiAccess='false'" /manifest:embed /DEBUG /PDB:V:\T\binskim\src\BuildSamples\SourceLink\x64\Release\CPlusPlus_SourceLink.pdb /SUBSYSTEM:CONSOLE /OPT:REF /OPT:ICF /LTCG:incremental /LTCGOUT:x64\Release\CPlusPlus_SourceLink.iobj /TLBID:1 /DYNAMICBASE /NXCOMPAT /IMPLIB:V:\T\binskim\src\BuildSamples\SourceLink\x64\Release\CPlusPlus_SourceLink.lib /MACHINE:X64 **/sourcelink:x64\Release\CPlusPlus_SourceLink.sourcelink.json**

[https://docs.microsoft.com/en-us/cpp/build/reference/sourcelink?view=msvc-170](https://docs.microsoft.com/en-us/cpp/build/reference/sourcelink?view=msvc-170)

PR #659, opened 2022-06-25
[RULE REQUEST] Warn for compiler and linker flags that are bad for performance are used for release builds

### LINKED ISSUE #651: [RULE REQUEST] Warn for compiler and linker flags that are bad for performance are used for release builds
# Rule Proposal: Binskim performance checking

This proposal is to add __several__ checks to binskim for "performance" issues that should be flagged.  The specific text below is an _example_ of a rule that should be included.  There are more than this.  These checks should only apply to release builds.  Debug builds are often unoptimized and that is ok.  

There are several compiler and linker flags that are bad for performance or that are implicit.  If the maintainers do not know of these then their performance will be worse without them knowing why.  Binskim would be a great place to expose additional warnings and errors _for release builds_ so that builds running binskim will be informed about these problems.  

For example, enabling /DEBUG in MSVC automatically enables incremental linking.  Incremental linking wastes performance for official release builds that do not benefit from incremental linking.  

Another common performance pitfall is [Runtime Type Information (RTTI)](https://docs.microsoft.com/en-us/cpp/build/reference/gr-enable-run-time-type-information?view=msvc-170) in C++.  This is rarely used but has __significant__ code size overhead.  It is also enabled by default in MSVC so people need to know to disable it or else their binaries have extra code size for no benefit.  

#### Rule metadata
- *Id*:PERF0001 (Windows)
- *Name*: DisableIncrementalLinking

#### Rule descriptions
- *Short*: Incremental linking enabled for release build
- *Full*: Incremental linking support increases binary size and can reduce runtime performance.  Fully optimized release builds should not specify incremental linking.

#### Platform & applicability
- *Platform*: Windows
- *Applicable to*: DLLs and EXEs compiled using MSVC
- *Not applicable to*: non-MSVC compilers

#### User-facing strings
- *Fail*: '{0}' was compiled with incremental linking enabled, which inflates binary size and can reduce performance.
- *Pass*:  '{0}' was compiled with incremental linking disabled.
- *[Other]*: '{0}' was not compiled as release, skipping performance checks

[/INCREMENTAL:NO](https://docs.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=msvc-170)

#### How to resolve
Add `/INCREMENTAL:NO` to the link.exe command line or `<LinkIncremental Condition="$(Configuration) == 'Release'">false</LinkIncremental>` to the msbuild project file.

dmachaj:
Closes #651 
[RULE REQUEST] Warn for compiler and linker flags that are bad for performance are used for release builds

## Why is this change being made?
The purpose of this change is to establish a pattern of identifying and warning about __performance__ problems in BinSkim, in addition to security problems.  BinSkim has a wide (and well deserved) reach so it would be great if it could provide coverage here.  There are a variety of compiler/linker options that are good or bad for performance and in many cases the problems are non-obvious.  BinSkim warning about these non-obvious problems would help improve the performance of software whenever it is run and followed.

## Briefly summarize what changed
This change starts performance warnings with the least controversial problem that I am aware of: incremental linking.  Incremental linking is good for debug builds but it is very bad for optimized release builds that are released publicly.  It leaves padding around the binary so that more code can be shoved in later.  That is wasteful for release builds.

The `BA6001.DisableIncrementalLinking` rule was created.  I semi-arbitrarily picked the 6000 range for performance-related warnings.  This warning tries to only apply to optimized code so it tries to sniff out whether the binary was optimized, or not.  If the binary is unoptimized then clearly performance checking is not needed so it skips analysis.  The current approach for detecting optimization is a heuristic so there is _probably_ a better way to do it.  I don't know it so I'm hoping this PR can help drive discussion.

The other aspect is detecting the linker flags.  There does not seem to be precedent for analyzing the link.exe command line so I had to establish a `LinkerCommandLine` class to encapsulate that.  The docs.microsoft.com documentation for incremental linking was encoded as logic in here to determine whether incremental linking will happen, or not.

The outer rule then checks if optimizations are enabled and incremental linking is also enabled.  If both are true then it is a problem and it logs a warning.

The documentation was also updated as appropriate.

## How was this change tested?
I tested it locally using a variety of binaries with different variations of linker flags.  These binaries are added as test cases.  There is also a new set of functional tests for this rule (which pass; although the NotApplicable test is sketchy).


src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -867,5 +867,43 @@ private ChecksumAlgorithmType ChecksumAlgorithmForFullPdb(Pdb pdb)
...
+                if (om.GetObjectModuleDetails().OptimizationsEnabled)
+                {
+                    ++optimizedCount;
+                }
+            }
+
+            // Arbitrary threshold of 50% optimized.
+            // Things linked into the binary also count as symbols so this can be complicated.  There is probably a better heuristic than this.
+            // A small single-file win32 console application is just over 50% optimized.  A large project is more like 95%.
+            if (((float)optimizedCount / (float)count) > 0.50f)
```
dmachaj, line 887: This heuristic is not ideal.  I'm looking for better options.

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
```diff
@@ -0,0 +1,109 @@
...
+            else if (debugSet)
+            {
+                if (optRef || optIcf || optLbr || order)
+                {
+                    this.IncrementalLinking = false;
+                }
+                else
+                {
+                    this.IncrementalLinking = true;
+                }
```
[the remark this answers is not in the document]
dmachaj: This is checking several bools so I feel like ternary would make readability worse, not better.

docs/BinSkimRules.md
```diff
@@ -870,3 +870,24 @@ This rule emits CSV data to the console for every compiler/language/version comb
 
 ---
 
+## Rule `BA6001.DisableIncrementalLinking`
```
pharring, line 873: This is going to clash with #657 
@michaelcfanning or @marmegh can you advise on how to assign these new proposed rule IDs?

dmachaj: The 6000 range was not claimed when I started coding this change.  If that range will be in use before this merges then I don't mind renumbering to something else (7000's?).  

Thanks for raising that.  I didn't realize there was another PR adding a 6001 rule.

pharring: I don't mind renumbering mine (SourceLink) either. Jumping up by 1,000 each time seems silly. For mine, I can't imagine having more than one or two rules related to Source Link. I certainly don't need to reserve 1,000! Maybe we should both just append onto the 2,000 range.
@michaelcfanning, @marmegh what's the thinking on rule ID assignment? Is there an established convention or policy? For example, the "RULE REQUEST" template asks for a rule ID, but isn't it up to the maintainers to decide what the new ID should be (and avoid clashes like this)?

src/BinSkim.Rules/PERules/BA6001.DisableIncrementalLinking.cs
```diff
@@ -0,0 +1,120 @@
...
+    public class DisableIncrementalLinking : PEBinarySkimmerBase, IOptionsProvider /* Delete this if no special configuration required */
+    {
+        /// <summary>
+        /// BA6001
+        /// </summary>
+        public override string Id => RuleIds.DisableIncrementalLinking;
+
+        /// <summary>
+        /// Recapitulate the full text of the rule description returned below
+        /// here as a summary comment.
```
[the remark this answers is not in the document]
dmachaj: Whoops missed one.

src/BinSkim.Rules/PERules/BA6001.DisableIncrementalLinking.cs
```diff
@@ -0,0 +1,120 @@
...
+                    nameof(RuleResources.BA6001_Warning),
+                    nameof(RuleResources.BA6001_Pass_NonReleaseBuild)
+                };
+
+        public IEnumerable<IOption> GetOptions()
+        {
+            return new List<IOption>
+            {
+                MinimumRequiredLinkerVersion
+            }.ToImmutableArray();
```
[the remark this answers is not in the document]
dmachaj: 🤷 the docs say to begin with [[https://github.com/microsoft/binskim/blob/main/docs/BAXXXX.RuleFriendlyName.cs](https://github.com/microsoft/binskim/blob/main/docs/BAXXXX.RuleFriendlyName.cs)](https://github.com/microsoft/binskim/blob/main/docs/BAXXXX.RuleFriendlyName.cs) and modify it to suit your needs.  I don't have a good grasp of what exactly this solves.  

Although on closer inspection it is probably a result of the `IOptionsProvider` inheritance on line 30.  I'll see if I can eliminate this.

src/BinSkim.Rules/PERules/BA6001.DisableIncrementalLinking.cs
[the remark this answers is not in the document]
dmachaj: Did a pass on all of the comments and cleaned it up a bunch.

docs/BinSkimRules.md
marmegh: @pharring, we're still reviewing both changes for the best rule id range.

src/BinSkim.Rules/PERules/BA6001.DisableIncrementalLinking.cs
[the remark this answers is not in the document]
dmachaj: I don't need this options interface so I removed this whole block.

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -867,5 +867,43 @@
...
+            uint optimizedCount = 0;
+            foreach (DisposableEnumerableView<Symbol> omView in pdb.CreateObjectModuleIterator())
+            {
+                Symbol om = omView.Value;
+                ++count;
+                if (om.GetObjectModuleDetails().OptimizationsEnabled)
+                {
+                    ++optimizedCount;
+                }
+            }
```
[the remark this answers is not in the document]
dmachaj: This method follows the pattern that is already established by the surrounding code in this file.

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -867,5 +867,43 @@
...
+            foreach (DisposableEnumerableView<Symbol> omView in pdb.CreateObjectModuleIterator())
+            {
+                Symbol om = omView.Value;
+                if (om.GetObjectModuleDetails().IncrementalLinkingEnabled)
+                {
+                    // There should be exactly one symbol in the binary containing knowledge about the linker command-line.  If that one symbol
+                    // says that incremental linking was enabled then the whole PE had it enabled too.
+                    return true;
+                }
+            }
```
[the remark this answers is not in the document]
dmachaj: This method follows the pattern that is already established by the surrounding code in this file.

src/BinaryParsers/PEBinary/ProgramDatabase/CompilerCommandLine.cs
```diff
@@ -100,6 +106,17 @@ public CompilerCommandLine(string commandLine)
                                 this.WarningLevel = wChar - '0';
                             }
                         }
+                        else if (argument.EndsWith("O1") || argument.EndsWith("02") || argument.EndsWith("Og") || argument.EndsWith("Os") || argument.EndsWith("Ot") || argument.EndsWith("Ox"))
```
lhecker, line 109: The `"O2"` has a 0 (zero) instead of an O.

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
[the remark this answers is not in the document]
lhecker: While I'd generally agree, I personally think in this case this reads fine:
```cpp
this.IncrementalLinking = !optRef && !optIcf && !optLbr && !order;
```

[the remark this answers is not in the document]
dmachaj: Yeah that's fair.

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
DrusTheAxe: Detection just from a binary? Tough one.

Imports to debug CRTs imply not-optimized

Similarly, if the binary has certain symbols from a statically linked debug CRT it's likely not-optimized. Not sure what symbols to look for. In particular in cases like WinAppSDK using the HybridCRT approach the Standard C library is UCRT and just the Standard C++ stuff (vector etc) are static. Not sure if there's any debug artifacts in there one could sniff out

DrusTheAxe:
How about MSBUILD .props/.targets to do this detection? Projects using MSBUILD (including VS *proj) could import PerformanceWarning.target to get detection at build-time (including dev inner-loop build-time). That would be a big plus over binskim which is rarely (can't be?) run locally as part of inner-loop.

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
dmachaj: That's an interesting suggestion.  Linking to the debug C runtime definitely means a debug build.  However many projects link to the release C runtime, even for debug builds, so it is not a firm guarantee.  That might be a good early-exit approach.  I'll think about this more.

dmachaj:
> How about MSBUILD .props/.targets to do this detection? Projects using MSBUILD (including VS *proj) could import PerformanceWarning.target to get detection at build-time (including dev inner-loop build-time). That would be a big plus over binskim which is rarely (can't be?) run locally as part of inner-loop.

> I've seen build logic like that but can't recall where. Something something UCRT something Windows something?

BinSkim is a tool that is packaged up in a NuGet.  It can be run locally so long as you know to.  That wouldn't be part of the _inner loop_ but these sorts of things don't really change that often so I'm not sure it makes sense to run it with every build.  I think the important part is outer loop coverage.

The other benefit of BinSkim over something like .props/.targets is that it is agnostic of the build tooling.  As long as cl.exe/link.exe are used it doesn't matter if you use CMake, Ninja, nmake, msbuild, a 20 year old batch file, or whatever else to compile your project.  It can still check the results.

DrusTheAxe:
Ohhh I didn't mean to imply either-or. BinSkim is cool but as you note the weakness is it's a step removed from source so you're reduced to digging up dinosaur bones to determine if they were warm- or cold-blooded, their color and if they had feathers. Much easier to do at the source level starting with a Condition="'$(Configuration)'=='Release'". Devs usually build Debug flavors locally so its not really an inconvenience, and even better if that Condition includes an optional opt-out if you define a property <DavidPerfDetector>false</> and if unspecified default to true for Release and false for Debug

The actual detection is easy since you have access to the actual compiler and linker options. WinAppSDK and other projects have various pre-build checks so it's paved ground too

P.S. The debugCRT isn't definitive but seems like a solid way to early exit. Heuristically distinguishing optimized vs not just from a binary is hard

RTTI is a good one to probe for. Fortunately you should be able to spot RTTI metadata in the binary to reliable warnnon that front

Do you have a specific list of what options you *want* to detect? Always good to start with goals... 🙂

docs/BinSkimRules.md
michaelcfanning: Sorry to be late jumping in, these are both excellent contributions and we appreciate them. I just assigned BA2027 to Paul, as source link has a clear security connection. I tend to think that starting a new id line with BA6001 for analysis that's strictly performance relevant makes sense. So you win, @dmachaj, stick with it. :)

I plan to give this PR careful attention tomorrow, when I have some more time. Thanks again for the contribution! This will be an interesting discussion.

dmachaj: @michaelcfanning please let me know if you have any thoughts, questions, or feedback.  I'm especially interested in discussing the expansion of BinSkim into performance warnings.  I have several more that I'd like to follow up with once the pattern is established.  Thanks.

michaelcfanning: I'm very sorry for the delay here, and I'm headed out for a break. We did manage to get @pharring's rule merged and you're next in the queue. That merge means you have file conflicts to resolve, sorry about that. Also, the fact that you're contributing from a fork (which is typical for mature/larger OSS projects) happens to prevent our CI/CD build from verifying your change (because it is hardened to only process branches from trusted contributors). I have therefore made you a write contributor. So if you have time to resolve merge conflicts and move your contribution over to a branch off main and submit a new PR, I'd appreciate it. I will be reviewing your rule as soon as I can get to it, which will be 7/20 (Wednesday) at the earliest. Sorry again for the delay.

Once we get through this first contribution and establish a pattern, the next perf rules s/be smoother sailing. :) Thank you for this contribution!

dmachaj:
@DrusTheAxe @jonwis @pharring @lhecker as an FYI this PR is being closed and is moving to #667 (same changes, not from a fork).

docs/BinSkimRules.md
dmachaj: No worries.  I have merged and un-forked and there is now #667.

PR #660, opened 2022-06-26
Fix flaky test

### LINKED ISSUE #658: Flaky unit test: MultithreadedAnalyzeCommand_ReturnCommonPathRootFromTargetSpecifiersIfOneExists
The MultithreadedAnalyzeCommand_ReturnCommonPathRootFromTargetSpecifiersIfOneExists unit test fails occasionally.
For example, one of the random seeds is 63096267.6169

```
Expected sb.Length to be 0 because The test was expecting 'c:\' but found 'C:\' when shuffled with seed '63096267.6169'.
    , but found 87 (difference of 87).
```

It fails on this test case:
```csharp
new
{
    TargetFileSpecifiers = new[]
    {
        @"c:\*.dll",
        @"c:\*.dll",
        @"C:\*.DLL"
    },
    ExpectedCommonPath = @"c:\"
}
```

The ReturnCommonPathRootFromTargetSpecifiersIfOneExists algorithm is case-insensitive, and all three inputs are equivalent, so it's fair to select any one of them. Due to random shuffling, there's a 1 in 3 chance it'll be the upper-case version causing the test case to fail.

I think the fix is to make the success/fail test case insensitive too.
i.e. instead of:
```csharp
commonPath != testCase.ExpectedCommonPath
```
do this:
```csharp
!string.Equals(commonPath, testCase.ExpectedCommonPath, StringComparison.OrdinalIgnoreCase))
```

pharring:
Fixes #658


michaelcfanning (approved):
:shipit:

src/Test.UnitTests.BinSkim.Driver/MultithreadedAnalyzeCommandTests.cs
```diff
@@ -121,15 +121,15 @@ public void MultithreadedAnalyzeCommand_ReturnCommonPathRootFromTargetSpecifiers
             foreach (var testCase in testCases)
             {
                 string commonPath = MultithreadedAnalyzeCommand.ReturnCommonPathRootFromTargetSpecifiersIfOneExists(testCase.TargetFileSpecifiers);
-                if (commonPath != testCase.ExpectedCommonPath)
+                if (!string.Equals(commonPath, testCase.ExpectedCommonPath, StringComparison.OrdinalIgnoreCase))
```
michaelcfanning, line 124: >OrdinalIgnoreCase [](http://example.com/codeflow?start=93&length=17)

Tough problem, case-sensitivity in this scenario. This looks good as it would break cross-platform testing altogether if we every introduced case-sensitivity in this scenario. Thanks for the fix!

pharring:
@michaelcfanning what's holding up the merge? Is there something else I need to do?

shaopeng-gh (approved):
:shipit:

michaelcfanning:
Sorry for the delay on this, have set auto-merge, so s/be checked in soon.

PR #667, opened 2022-07-19
[RULE REQUEST] Warn for compiler and linker flags that are bad for performance are used for release builds

### LINKED ISSUE #651: [RULE REQUEST] Warn for compiler and linker flags that are bad for performance are used for release builds

dmachaj:
Closes #651 
[RULE REQUEST] Warn for compiler and linker flags that are bad for performance are used for release builds

## Why is this change being made?
The purpose of this change is to establish a pattern of identifying and warning about __performance__ problems in BinSkim, in addition to security problems.  BinSkim has a wide (and well deserved) reach so it would be great if it could provide coverage here.  There are a variety of compiler/linker options that are good or bad for performance and in many cases the problems are non-obvious.  BinSkim warning about these non-obvious problems would help improve the performance of software whenever it is run and followed.

## Briefly summarize what changed
This change starts performance warnings with the least controversial problem that I am aware of: incremental linking.  Incremental linking is good for debug builds but it is very bad for optimized release builds that are released publicly.  It leaves padding around the binary so that more code can be shoved in later.  That is wasteful for release builds.

The `BA6001.DisableIncrementalLinking` rule was created.  I semi-arbitrarily picked the 6000 range for performance-related warnings.  This warning tries to only apply to optimized code so it tries to sniff out whether the binary was optimized, or not.  If the binary is unoptimized then clearly performance checking is not needed so it skips analysis.  The current approach for detecting optimization is a heuristic so there is _probably_ a better way to do it.  I don't know it so I'm hoping this PR can help drive discussion.

The other aspect is detecting the linker flags.  There does not seem to be precedent for analyzing the link.exe command line so I had to establish a `LinkerCommandLine` class to encapsulate that.  The docs.microsoft.com documentation for incremental linking was encoded as logic in here to determine whether incremental linking will happen, or not.

The outer rule then checks if optimizations are enabled and incremental linking is also enabled.  If both are true then it is a problem and it logs a warning.

The documentation was also updated as appropriate.

## How was this change tested?
I tested it locally using a variety of binaries with different variations of linker flags.  These binaries are added as test cases.  There is also a new set of functional tests for this rule (which pass; although the NotApplicable test is sketchy).


docs/BinSkimRules.md
```diff
@@ -914,3 +914,25 @@ This rule emits CSV data to the console for every compiler/language/version comb
 
 ---
 
+## Rule `BA6001.DisableIncrementalLinking`
```
michaelcfanning, line 917: >DisableIncrementalLinking [](http://example.com/codeflow?start=16&length=25)

I like the balance you appear to be striking in analysis. If you see evidence that suggests the standard is relevant (i.e., an actual release build), you fault-in the analysis, otherwise we drop it.

docs/BinSkimRules.md
```diff
@@ -914,3 +914,25 @@ This rule emits CSV data to the console for every compiler/language/version comb
...
+
+### Description
+
+Incremental linking support increases binary size and can reduce runtime performance. Fully optimized release builds should not specify incremental linking.
+
+### Messages
+
+#### `Pass`: Pass
+
+'{0}' was compiled with incremental linking disabled.
```
michaelcfanning, line 927: >was [](http://example.com/codeflow?start=6&length=3)

So, how confident are we in your 'release' build detection heuristic? That will dictate whether we use 'is' or 'appears to be' below. My note is to include this distinguishing characteristic in the output. We generally emit this data: 1) why did we fire or not, that is., what specific data is salient to the result, 2) what is the impact, and (for errors) 3) what are the range of fixes. E.g., 

'{0}' is a fully optimized release binary that is not compiled with incremental linking, minimizing binary sizes with a corresponding increase in runtime performance.

docs/BinSkimRules.md
```diff
@@ -914,3 +914,25 @@ This rule emits CSV data to the console for every compiler/language/version comb
 
 ---
 
+## Rule `BA6001.DisableIncrementalLinking`
+
+### Description
+
+Incremental linking support increases binary size and can reduce runtime performance. Fully optimized release builds should not specify incremental linking.
```
michaelcfanning, line 921: >binary [](http://example.com/codeflow?start=38&length=6)

So, let's maybe add some details on what's going on here, incremental link adds lots of little records, unnecessary data to the PE, yes? Incremental linking disables prevents full optimization (you can't perform whole program LTCG linking for example, right)? I'm guessing that latter issue comprises the greatest perf risk?

Are there other considerations we should mention? an incremental build won't generate a deterministic output, for example. poking around the web, I see a point that the codegen for incremental linking could itself be a source of bugs (idiosyncratic to your non-deterministic build). Makes sense. Anything else?

docs/BinSkimRules.md
```diff
@@ -914,3 +914,25 @@ This rule emits CSV data to the console for every compiler/language/version comb
...
+
+#### `Pass`: Pass
+
+'{0}' was compiled with incremental linking disabled.
+
+#### `Warning`: Warning
+
+'{0}' was compiled with incremental linking enabled, which inflates binary size and can reduce performance.
+
+#### `NonReleaseBuild`: Pass
```
michaelcfanning, line 933: >Pass [](http://example.com/codeflow?start=24&length=4)

So this should actually be emitted as a 'not applicable' message, i.e., this output doesn't appear to be a release build, therefore enforcing this particular standard does not make sense. IOW, we reserve 'pass' for 'all conditions for satisfying this particular quality standard lined up and you hit the mark'.

src/BinaryParsers/PEBinary/ProgramDatabase/CompilerCommandLine.cs
```diff
@@ -100,6 +106,17 @@ public CompilerCommandLine(string commandLine)
                                 this.WarningLevel = wChar - '0';
                             }
                         }
+                        else if (argument.EndsWith("O1") || argument.EndsWith("O2") || argument.EndsWith("Og") || argument.EndsWith("Os") || argument.EndsWith("Ot") || argument.EndsWith("Ox"))
+                        {
+                            // https://docs.microsoft.com/cpp/build/reference/o-options-optimize-code?view=msvc-170
+                            // /O1 /O2 /Og /Os /Ot /Ox are all indicative of optimizations being enabled
+                            this.OptimizationsEnabled = true;
+                        }
+                        else if (argument.EndsWith("Od"))
```
michaelcfanning, line 115: >argument [](http://example.com/codeflow?start=33&length=8)

So what about command-line argument order-sensitivity? Will the compiler error out if I try to optimize with /O1 and also provide /Od? (so that we have a guarantee these two conditions are exclusive between successful compilations). Or does /Od override or not depending on position? Etc. 

The same question exists for the other functionality in this method, I have suspicious we should review the warning level handling.

But since you're code is new and this looks pretty easy to test, I'd like to dig into that for the optimizations settings.

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
```diff
@@ -0,0 +1,102 @@
...
+            bool explicitlyDisabled = false;
+            foreach (string argument in ArgumentSplitter.CommandLineToArgvW(commandLine))
+            {
+                if (!CommandLineHelper.IsCommandLineOption(argument))
+                {
+                    continue;
+                }
+
+                // There are multiple /debug options so use StartsWith
+                if (argument.StartsWith("/debug", System.StringComparison.OrdinalIgnoreCase))
```
michaelcfanning, line 51: >/ [](http://example.com/codeflow?start=41&length=1)

link.exe supports '-' as an arg prefix, is this symbol consistently normalized to '/' or do we need to handle the hyphen case here as well?

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
```diff
@@ -0,0 +1,102 @@
...
+                {
+                    explicitlyEnabled = true;
+                }
+                else if (string.Equals(argument, "/incremental:no", System.StringComparison.OrdinalIgnoreCase))
+                {
+                    explicitlyDisabled = true;
+                }
+            }
+
+            // Explicit enable or disable wins.
```
michaelcfanning, line 81: >Explicit [](http://example.com/codeflow?start=15&length=8)

This is a nice comment that provides some reassurance to me that we understand the command-line interpretation behavior, thanks for adding it.

src/BinaryParsers/PEBinary/ProgramDatabase/ObjectModuleDetails.cs
```diff
@@ -53,6 +65,16 @@ public class ObjectModuleDetails
         /// </summary>
         public int WarningLevel => this.compilerCommandLine.WarningLevel;
 
+        /// <summary>
+        /// Returns whether optimizations were enabled as defined by the various /O switches in the compiland command line
```
michaelcfanning, line 69: >command [](http://example.com/codeflow?start=110&length=7)

The doc comments in this file should all end with a period, have a moment to quickly fix that up?

src/BinSkim.Rules/RuleIds.cs
```diff
@@ -71,5 +71,8 @@ internal static class RuleIds
         // MachO Checks
         public const string EnablePositionIndependentExecutableMachO = "BA5001";
         public const string DoNotAllowExecutableStack = "BA5002";
+
+        // Performance (not security) checks
+        public const string DisableIncrementalLinking = "BA6001";
```
michaelcfanning, line 76: >DisableIncrementalLinking [](http://example.com/codeflow?start=28&length=25)

OK, here comes the quibbling with semantics. :) Incremental linking, in fact, needs to be explicitly opted in, and so 'disabling' it refers to the absence of a specific action. And so I'd reverse it, and frame the rule to advise against opting into this adverse condition. Also might want to qualify the name so that it's clear we're talking release builds.

Note this readable rule name serves as a nice abstract for the check, with useful detail, still pretty concise. That's the goal.

DoNotEnableIncrementalLinkingInReleaseBuilds

docs/BinSkimRules.md
```diff
@@ -914,3 +914,25 @@ This rule emits CSV data to the console for every compiler/language/version comb
...
+
+'{0}' was not compiled as release, skipping performance checks
```
michaelcfanning, line 935: >was [](http://example.com/codeflow?start=6&length=3)

'{0}' appears to be compiled as release and is therefore not subject to performance checks.

src/BinSkim.Rules/PERules/BA6001.DisableIncrementalLinking.cs
```diff
@@ -0,0 +1,93 @@
...
+                };
+
+        private const string AnalyzerName = RuleIds.DisableIncrementalLinking + "." + nameof(DisableIncrementalLinking);
+
+        public override AnalysisApplicability CanAnalyzePE(PEBinary target, Sarif.PropertiesDictionary policy, out string reasonForNotAnalyzing)
+        {
+            PE portableExecutable = target.PE;
+            AnalysisApplicability result = AnalysisApplicability.NotApplicableToSpecifiedTarget;
+
+            reasonForNotAnalyzing = MetadataConditions.ImageIsILOnlyAssembly;
```
michaelcfanning, line 50: >ImageIsILOnlyAssembly [](http://example.com/codeflow?start=55&length=21)

There's probably some analog here for .NET, to support edit-and-continue.

i.e., don't ship managed assemblies compiled for edit-and-continue.

I'm not sure what the risk is, relatively, from making this mistake in .NET vs. native binaries. also not clear what a reliable 'compiled for release' heuristic would be for that platform (but am guessing a fairly solid one could be contrived).

Anyway, just noting, if there's demand in future, we could consider updating this check.

src/BinSkim.Rules/PERules/BA6001.DisableIncrementalLinking.cs
```diff
@@ -0,0 +1,93 @@
...
+        public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext context)
+        {
+            PEBinary target = context.PEBinary();
+            PE pe = target.PE;
+            Pdb pdb = target.Pdb;
+
+            if (!pe.IsMostlyOptimized(pdb))
+            {
+                context.Logger.Log(this,
+                    RuleUtilities.BuildResult(ResultKind.NotApplicable, context, null,
```
michaelcfanning, line 72: >NotApplicable [](http://example.com/codeflow?start=57&length=13)

ah ha! You *are* emitting a not applicable message, it's just that your string name is wrong, it should be

BA6001_NotApplicable_NonReleaseBuild

I actually suggest you create a new MetadataCondition 'NotAReleaseBuild' since we project to have more of these checks. Then, your logging resolves to the following call (taking this approach will keep your output consistent with all other N/A reports).

Notes.LogNotApplicableToSpecifiedTarget(context, MetadataConditions.NotAReleaseBuild);

src/BinSkim.Rules/PERules/BA6001.DisableIncrementalLinking.cs
```diff
@@ -0,0 +1,93 @@
...
+                    nameof(RuleResources.BA6001_Pass_NonReleaseBuild),
+                    context.TargetUri.GetFileName()));
+                return;
+            }
+
+            if (pe.IncrementalLinkingEnabled(pdb))
+            {
+                context.Logger.Log(this,
```
michaelcfanning, line 80: >Logger [](http://example.com/codeflow?start=24&length=6)

Generally, when we log, we provide a code comment that has the complete text of the message that is output to the users, basically, recap the localized message string.

This helps when reading the code and maximizes chances devs will go fix up the messages to improve or update them when logic is changed. The downside is message drift, we need to make sure both copies stay in sync.

Go take a look at any existing security rule, I think we've mostly got them all in shape for this code convention.

docs/BinSkimRules.md
dmachaj: Thank you :).  Yes, that is the goal.  I'm trying to make the analysis as valuable as possible by reducing false positives or cases where it warns for things that don't matter.

docs/BinSkimRules.md
dmachaj: Moderately but not highly confident.

I tried to ask around for better guidance and there didn't seem to really be much.  It is a hard problem to determine whether the intent was a Release or Debug build just by looking at the final binary.  Being part of the build process would provide firm answers to that but would lose the huge scope of BinSkim where it can try to analyze any valid PE file.

The best suggestion for this, which I have not been able to code yet, was to early exit if linking to the debug C runtime.  That should be very strong evidence of a non-release build.  The opposite is not definitive though because it is common to always use the Release C runtime, even for Debug builds.

All that said, I think that these performance rules will always be warnings instead of blocking errors.  That reduces the potential harm significantly.  If we over-eagerly warn about these then there should hopefully not be any damage.

docs/BinSkimRules.md
dmachaj: I think the opposite is true.  Enabling /LTCG (or a few other linker options) will prevent incremental linking.  That limits the scope somewhat.  However, I think it is a confusing pitfall that enabling /DEBUG (produce a PDB) also implicitly enables incremental linking.  That is a non-obvious default that in many cases harms performance.  Hence this rule.

I will try to elaborate further in this description.

docs/BinSkimRules.md
michaelcfanning: That's a good early exit strategy, detection of debug c runtime, that's a good vote for 'debug' and we should take advantage of it, falling back the remaining heuristics otherwise (definitely take your point that the situation will still be ambiguous).

docs/BinSkimRules.md
michaelcfanning: Got it, that definitely argues for a rule. This level of detail is very helpful to include in rule descriptions. This whole nest of command-line args is very involved, thanks for whatever information you can provide.

docs/BinSkimRules.md
michaelcfanning: For performance, people seem generally not be overly upset about false negatives. The rate of discovery for perf issues related to compilation settings is low and the psychology of the issues is different than for security (where people rightly can get agitated over a missed finding). I think firing warnings is a good idea, then we can enable these by default and interleave them into regular analysis. Teams that want to be strict about certain settings can use a config file to elevate to error.

docs/BinSkimRules.md
dmachaj: I had that originally and changed it to a pass before I sent the PR (although it was a few weeks ago now and I'm struggling to remember my full rationale for that).  I'll change it back.

src/BinaryParsers/PEBinary/ProgramDatabase/CompilerCommandLine.cs
dmachaj: It warns if you combine them (VS2022 toolchain):
`1>cl : command line warning D9025: overriding '/O2' with '/Od'`
`1>cl : command line warning D9025: overriding '/Od' with '/O2'`

The second option seems to win.  As a result I _think_ this is correct because whichever comes second will set `this.OptimizationsEnabled` and that is what the compiler will be doing.

docs/BinSkimRules.md
dmachaj: [https://docs.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=msvc-170#remarks](https://docs.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=msvc-170#remarks)

"By default, the linker runs in incremental mode. To override a default incremental link, specify /INCREMENTAL:NO.

An incrementally linked program is functionally equivalent to a program that is non-incrementally linked. However, because it is prepared for subsequent incremental links, an incrementally linked executable, static library, or dynamic-link library file:

* Is larger than a non-incrementally linked program because of padding of code and data. Padding enables the linker to increase the size of functions and data without recreating the file.
* May contain jump thunks to handle relocation of functions to new addresses."

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
```diff
@@ -0,0 +1,102 @@
...
+                }
+                else if (string.Equals(argument, "/order", System.StringComparison.OrdinalIgnoreCase))
+                {
+                    order = true;
+                }
+                else if (string.Equals(argument, "/incremental", System.StringComparison.OrdinalIgnoreCase))
+                {
+                    explicitlyEnabled = true;
+                }
+                else if (string.Equals(argument, "/incremental:no", System.StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
dmachaj: There is no output from link.exe when this is specified multiple times.  I ran `link.exe <otherargs> /INCREMENTAL /INCREMENTAL /INCREMENTAL:NO` and there were no warnings or errors (even on diagnostic output).  I am unsure how to determine whether it _actually_ incrementally linked so I can't tell which option won.  I would guess the last, which would match the /O2 vs. /Od parsing.

I think I can account for this by wiping the opposite option when one of these is encountered.

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
dmachaj: I think you have more confidence than I do!  

I don't work on MSVC and have no knowledge of the inner workings and am attempting to encode the public documentation (from [https://docs.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=msvc-170](https://docs.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=msvc-170)).   This comment is regarding the interpretation in this code.  It could be inaccurate to how link.exe behaves.

src/BinaryParsers/PEBinary/ProgramDatabase/ObjectModuleDetails.cs
dmachaj: Sure.  Leave it better than you found it.

src/BinSkim.Rules/RuleIds.cs
dmachaj: That's more specific so it seems good.  Will change.

src/BinSkim.Rules/PERules/BA6001.DisableIncrementalLinking.cs
dmachaj: I have very limited knowledge of .NET so I am unsure whether a rule like this could apply to managed binaries.  It very well might!  I simply don't know enough.

src/BinSkim.Rules/PERules/BA6001.DisableIncrementalLinking.cs
dmachaj: Will do.

docs/BinSkimRules.md
dmachaj: This one is more complicated so saving for the next round of addressing feedback.

src/BinSkim.Rules/PERules/BA6001.DisableIncrementalLinking.cs
dmachaj: Took me a while to fully figure out how to switch that over but I think it is a nice improvement.  I also fixed the NotApplicable unit test to actually be useful and confirm the inapplicability reason.

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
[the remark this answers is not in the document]
michaelcfanning: OK, sounds like empirically perhaps last setting wins. I can imagine the compiler tools might 'evolve' to support this (tolerance for conflicting settings, honoring last observed value) to create more resilience for build systems that add settings/args to inherited sets of them. Just speculation. We can probably go find stronger confirmation internally for the MSVC toolset.

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
michaelcfanning: Thanks for correcting my optics. :) I can probably locate relevant MSVC parsing logic for us.

src/BinaryParsers/PEBinary/ProgramDatabase/CompilerCommandLine.cs
michaelcfanning: Yes, makes sense and consistent with some of your other testing. As long as our logic for exclusive conditions consistently sets all relevant state in each case, we can assume we'll exit parsing in the same condition as the compiler, in terms of assumed configuration.

docs/BinSkimRules.md
dmachaj: Actually this was easier than I expected.  The `/MTd` and `/MDd` compiler options are a nice clear signal so I have it early-exit as non-optimized when those are used anywhere.  That check is accurate for all of the binaries I have handy for local testing.

src/BinaryParsers/PEBinary/ProgramDatabase/CompilerCommandLine.cs
```diff
@@ -74,6 +79,8 @@ public CompilerCommandLine(string commandLine)
             this.WarningLevel = 0;
             this.WarningsAsErrors = false;
             this.OptimizationsEnabled = false;
+            this.UsesDebugCRuntime = false;
+
             var explicitWarnings = new Dictionary<int, WarningState>();
             foreach (string argument in ArgumentSplitter.CommandLineToArgvW(commandLine))
             {
```
DrusTheAxe, line 86: line 92 - why bother switching on length to subdivide the ifs?

dmachaj: This is a preexisting pattern so I can only speculate.  

The compiler can get both `-MT` and `/MT` as valid options so using `EndsWith("MT")` will work regardless of whether a - or / is used.  You don't want to over-match to other strings so requiring the length to match expectations will make this safe.

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
dmachaj: @michaelcfanning I just wanted to check in regarding whether you were able to locate the parsing logic for the linker command line.  I believe this is the only open issue remaining in this PR.  Thanks.

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
michaelcfanning: Are we going to make a change here? i.e., support both '/debug' and '-debug'?

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
michaelcfanning: I will go check. We don't have to block this immediate PR for that, we can update as necessary once checked in. I'd like to understand whether we need to update the arg parsing to support both the '-' and '/' prefixes first, though (see separate comment).

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
dmachaj: Whoops missed this one.

dmachaj: Fixed now.

michaelcfanning (approved):
:shipit:

PR #668, opened 2022-07-19
Fix failing build pipeline.

marmegh:
Build pipeline was failing because Functional Test Data for BA2027 was not copied to output directory and was missing required PDB files. Added PDB files where available and removed NotApplicable test cases where CanAnalyze would not be called. These changes do not affect rule functionality or impact usage, therefore there is no update to the Release History.


marmegh:
Closing in favor of #669

PR #671, opened 2022-07-21
Fix Linux Tests

shaopeng-gh:
1. Before fix, in Linux and macOS the task of BuildAndTest will always show as green checkmark even there is test failure.
with green checkmark, people will not actively look into test results. This is fixed now.
2.  The newly added tests for windows binary is missing the run in windows only check. No one noticed it due to above. Fixed.
3.  The baseline is different in Linux and macOS vs Windows, and actually we only use re-baseline in Windows. This is a known issue, added run in windows only check.

History md file is not updated due to test fix only.


marmegh:
Great change, Shoapeng. Thank you for following up on these.

marmegh (approved):
:shipit:

PR #672, opened 2022-07-21
Update .gitignore

pharring:
A prior commit added standard lines to ignore build output -- including .pdb -- which caused problems because we do add PDBs to this repo as test data.


.gitignore
```diff
@@ -9,7 +9,7 @@
 bld/
 bin/
 obj/
-src/packages/
+**/packages/*
```
shaopeng-gh, line 12: >* [](http://example.com/codeflow?start=12&length=1)

for people wondering:

we had:
    "\binskim\src\sarif-sdk\packages"
    "\binskim\src\packages"

now one additional:
    "\binskim\src\BuildSamples\SourceLink\packages"

I think the last wildcard is redundant.

shaopeng-gh (approved):
:shipit:

michaelcfanning (approved):
:shipit:

PR #676, opened 2022-07-26
LinkerCommandLine fails to populate in some cases due to case sensitivity

### LINKED ISSUE #675: LinkerCommandLine fails to populate in some cases due to case sensitivity
This is a mistake in the new `LinkerCommandLine` extractor that I recently added to support performance rules.  Our learnings about case (in)sensitivity and both `/out` and `-out` being equivalent was not applied to ObjectModuleDetails.cs.  As a result it is not possible to isolate the linker command line arguments for certain binaries.

The fix is to make these comparisons case-insensitive and to check for both - and /.

dmachaj:
Closes #675

This is a mistake in the new `LinkerCommandLine` extractor that I recently added to support performance rules.  Our learnings about case (in)sensitivity and both `/out` and `-out` being equivalent was not applied to ObjectModuleDetails.cs.  As a result it is not possible to isolate the linker command line arguments for certain binaries.

The fix is to make these comparisons case-insensitive and to check for both - and /.

### DIFF
--- a/src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
+++ b/src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
@@ -122,7 +123,7 @@ namespace Microsoft.CodeAnalysis.BinaryParsers.ProgramDatabase
                 return false;
             }
 
-            return target.Substring(1).StartsWith(argument, System.StringComparison.OrdinalIgnoreCase);
+            return target.AsSpan().Slice(1).StartsWith(argument, System.StringComparison.OrdinalIgnoreCase);
         }
 
         private static bool ArgumentEquals(string target, string argument)
@@ -137,7 +138,7 @@ namespace Microsoft.CodeAnalysis.BinaryParsers.ProgramDatabase
                 return false;
             }
 
-            return target.Substring(1).Equals(argument, System.StringComparison.OrdinalIgnoreCase);
+            return target.AsSpan().Slice(1).Equals(argument, System.StringComparison.OrdinalIgnoreCase);
         }
     }
 }
--- a/src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
+++ b/src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
@@ -48,32 +48,32 @@ namespace Microsoft.CodeAnalysis.BinaryParsers.ProgramDatabase
                 }
 
                 // There are multiple /debug options so use StartsWith
-                if (argument.StartsWith("/debug", System.StringComparison.OrdinalIgnoreCase) || argument.StartsWith("-debug", System.StringComparison.OrdinalIgnoreCase))
+                if (ArgumentStartsWith(argument, "debug"))
                 {
                     debugSet = true;
                 }
-                else if (string.Equals(argument, "/opt:ref", System.StringComparison.OrdinalIgnoreCase) || string.Equals(argument, "-opt:ref", System.StringComparison.OrdinalIgnoreCase))
+                else if (ArgumentEquals(argument, "opt:ref"))
                 {
                     optRef = true;
                 }
-                else if (string.Equals(argument, "/opt:icf", System.StringComparison.OrdinalIgnoreCase) || string.Equals(argument, "-opt:icf", System.StringComparison.OrdinalIgnoreCase))
+                else if (ArgumentEquals(argument, "opt:icf"))
                 {
                     optIcf = true;
                 }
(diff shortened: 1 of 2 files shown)


src/BinaryParsers/PEBinary/ProgramDatabase/ObjectModuleDetails.cs
```diff
@@ -40,7 +40,8 @@ public class ObjectModuleDetails
             this.CompilerBackEndVersion = backEndVersion ?? new Version();
 
             // The command line for link.exe should contain /OUT with the final binary name.  Use that to separate compiler and linker command lines.
-            if ((commandLine != null) && commandLine.Contains("/OUT"))
+            if ((commandLine != null) &&
+                (commandLine.Contains("/OUT", System.StringComparison.OrdinalIgnoreCase) || commandLine.Contains("-OUT", System.StringComparison.OrdinalIgnoreCase)))
```
michaelcfanning, line 44: >commandLine [](http://example.com/codeflow?start=17&length=11)

Good fix. Since you put the hood up here, can we please create a small utility method that performs this case-insensitive, prefix-insensitive ('-' vs. '/') and then consume it here and elsewhere? Didn't want to block your previous PR where the same thought occurred to me. It'd be a good small clean-up change, I think.

dmachaj: Got a recommendation on where such a helper should go?  I don't see an obvious `Utils.cs` to stash it.

dmachaj: Since it is for the linker command-line in both cases it could arguably be a static method on `LinkerCommandLine`.  The OM code calls in and says "Is this a linker command line?" and gets a yes/no response.  Internally it uses the same helper for the other linker options.

dmachaj: I went ahead with wrapping it up into LinkerCommandLine because it is currently specific to that logic.  There are comparisons with `Contains`, `StartsWith`, and `Equals` so it is a little bit messy to have several helpers.

I also needed to substring into the target to exclude the prefix which will generate some garbage.  There should only be one linker command line per binary so the impact of that should be minimal.

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
```diff
@@ -100,5 +100,44 @@ public LinkerCommandLine(string commandLine)
...
             }
         }
+
+        public static bool IsLinkerCommandLine(string commandLine)
+        {
+            // The command line for link.exe should contain /OUT with the final binary name.  Use that to separate compiler and linker command lines.
+            // We should accept both / and - as the argument prefix.  The argument is case insensitive.  We do not know where in the command-line the
+            // OUT argument will be provided so we use Contains.
+            return ((commandLine != null) &&
+                    (commandLine.Contains("/OUT", System.StringComparison.OrdinalIgnoreCase) || commandLine.Contains("-OUT", System.StringComparison.OrdinalIgnoreCase)));
```
michaelcfanning, line 110: >commandLine [](http://example.com/codeflow?start=21&length=11)

This code can call your new helper. :)

michaelcfanning (approved):
:shipit:

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
```diff
@@ -100,5 +100,44 @@ public LinkerCommandLine(string commandLine)
...
+            if (target.Length == 0)
+            {
+                return false;
+            }
+            else if ((target[0] != '/') && (target[0] != '-'))
+            {
+                return false;
+            }
+
+            return target.Substring(1).Equals(argument, System.StringComparison.OrdinalIgnoreCase);
```
michaelcfanning, line 140: >target [](http://example.com/codeflow?start=19&length=6)

we could use a span here, yes? and prevent the garbage string creation.

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
dmachaj: I don't think so because it is `Contains`.  The /OUT option could appear anywhere in the long command-line string.  This is not called with one-by-one arguments like the other helpers.

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
dmachaj: Unsure.  I don't spend much time in C# so I'll have to read up on that.

dmachaj: Thanks @asklar for the offline pointer to some helpful docs on that.  Fortunately that was an easy fix.

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
michaelcfanning: Right, yes, missed that point. We could have created CommandLineContains, I suppose, but with a single upstream caller, this factoring doesn't necessarily make sense for now.

src/BinaryParsers/PEBinary/ProgramDatabase/LinkerCommandLine.cs
michaelcfanning: Excellent, thanks.

PR #685, opened 2022-08-02
* BUGFIX: Fix `error ERR997.ExceptionLoadingPdb : '[filename]' was not evaluated because its PDB could not be loaded (E_PDB_NOT_FOUND).` when reading PE file built with `PDBPageSize:8192` or greater, by upgrading msdia140.dll from `14.27.28826.96` to `14.32.31326.0`.

shaopeng-gh:
# Description:

User reported issue by email:

"RE: BinSkim PDB Error related to PdbPageSize:8192"

Briefly: 
The BinSkim has been running well on our build pipeline, but has hit an error (below) when I am testing my change to enable “PDBPageSize:8192” support in all of our PDB generation.  The .PDB files are present in the same folder as before, but now have a changed internal format. I assert that E_PDB_NOT_FOUND error may be reported on cases where the file is present, but reading its content fails unexpectedly.

`D:\a\_work\1\a\BrowserCommonHarness.exe : error ERR997.ExceptionLoadingPdb : 'BrowserCommonHarness.exe' was not evaluated because its PDB could not be loaded (E_PDB_NOT_FOUND). `

The files do exist, with file names:
BrowserCommonHarness.exe  
BrowserCommonHarness.exe.pdb  

# Cause:

The creating Big PdbPageSize support only starts with Clang 14, and reading it need a newer version of msdia library as well.

# Fix:

update to the latest msdia140.dll comes with the VS 2022.


src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/MixedMode_x64_VS2015_Default.exe.sarif
```diff
@@ -988,20 +988,20 @@
               "id": "BA6001",
               "name": "DisableIncrementalLinkingInReleaseBuilds",
               "fullDescription": {
-                "text": "Incremental linking support increases binary size and can reduce runtime performance. Fully optimized release builds should not specify incremental linking."
+                "text": "Incremental linking support increases binary size and can reduce runtime performance. The support for incremental linking adds padding and other overhead to support the ability to modify a binary without a full link.  The use of incrementally linked binaries may reduce the level of determinism because previous compilations will have lingering effects on subsequent compilations.  Fully optimized release builds should not specify incremental linking."
```
shaopeng-gh, line 991: >"text": "Incremental [](http://example.com/codeflow?start=16&length=20)

I re-ran the baseline to see if my change affect baselines, all the changes are related to previous change `BA6001`.

shaopeng-gh:
btw, I think there are some potential wrong assumptions in the space:
1. assumption with PE we are looking for pdb. 
PE can also build with dwarf.
3. assumption of PE and PDB are build with MS toolchain. 
LLVM can also build PE with PDB
5. assumption the exe and pdb naming with a.exe and a.pdb
the pdb file name is coded inside exe, it can be any naming convention.
6. assuming the pdb file is the one for the exe just by name matches, a.exe and a.pdb.
I delete the pdb for a exe and try re-naming a random pdb back, code loads without question. This can be verified by matching GUID with exe and pdb.
7. PE rules and also how to fix considers only MS toolchain, some of them clang does not support. If we are supporting clang toolchain we may make some modification as appropriate.

src/Test.UnitTests.BinaryParsers/PEBinary/PEBinaryTests.cs
```diff
@@ -32,27 +32,40 @@ internal static string GetTestDirectory(string relativeDirectory)
             return Path.Combine(dirPath, relativeDirectory);
         }
 
-        [Fact]
-        public void PEBinary_PdbAvailable()
+        [Theory]
+        [InlineData("clangcl.14.pe.c.codeview.pdbpagesize_default.exe")]
```
shaopeng-gh, line 36: >14 [](http://example.com/codeflow?start=29&length=2)

The only version of test is 14 because it only starts support at v14, and the LLVM v15 is not yet in released status.

src/Test.UnitTests.BinaryParsers/PEBinary/PEBinaryTests.cs
```diff
@@ -32,27 +32,40 @@ internal static string GetTestDirectory(string relativeDirectory)
...
+        [InlineData("clangcl.14.pe.c.codeview.pdbpagesize_4096.exe")]
+        [InlineData("clangcl.14.pe.c.codeview.pdbpagesize_8192.exe")]
+        [InlineData("clangcl.14.pe.c.codeview.pdbpagesize_16384.exe")]
```
shaopeng-gh, line 39: >16384 [](http://example.com/codeflow?start=58&length=5)

these are not random test, it should be all the supported number:
/PdbPageSize:[4096|8192|16384|32768]
the difference between default and 4096 is whether it is explicitly passed.

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/MixedMode_x64_VS2015_Default.exe.sarif
dmachaj: Sorry about that one.  Somehow this staleness didn't break anything so I didn't notice it when I updated the strings based on PR feedback.  I was seeing these updates too when working on some additional rules and updating the baselines accordingly.

shaopeng-gh: no worries at all I just call it out to save reviewer's when see this diff.

With no enforcement in the build and test no one will remember to run re-baseline.
(
My preference is to change the re-baseline test to strict compare so that we are forced to re-run baseline and review the changes caused, expectedly and unexpectedly :) 
)

src/ReleaseHistory.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* BUGFIX: Fix `error ERR997.ExceptionLoadingPdb : '[filename]' was not evaluated because its PDB could not be loaded (E_PDB_NOT_FOUND).` when reading PE file built with `PDBPageSize:8192` or grater, by upgrading msdia140.dll from `14.27.28826.96` to `14.32.31326.0`. [685](https://github.com/microsoft/binskim/pull/685)
```
marmegh, line 5: >* BUGFIX: Fix `error ERR997.ExceptionLoadingPdb : '[filename]' was not evaluated because its PDB could not be loaded (E_PDB_NOT_FOUND).` when reading PE file built with `PDBPageSize:8192` or grater [](http://example.com/codeflow?start=0&length=197)

Nice description!

src/Test.UnitTests.BinaryParsers/PEBinary/PEBinaryTests.cs
```diff
@@ -32,27 +32,40 @@ internal static string GetTestDirectory(string relativeDirectory)
...
+        [InlineData("clangcl.14.pe.c.codeview.pdbpagesize_32768.exe")]
+        [InlineData("Native_x64_VS2022_PDBPageSize_8192.exe")]
+        [InlineData("Native_x64_VS2013_Default.dll", true)]
```
marmegh, line 42: This type of test adds extra latency to test runtime. We can get the same effect without the latency by adding a foreach loop over a list containing the same strings within the test.

src/Test.UnitTests.BinaryParsers/PEBinary/PEBinaryTests.cs
```diff
@@ -32,27 +32,40 @@ internal static string GetTestDirectory(string relativeDirectory)
...
                 peBinary.StrippedPdb.Should().BeNull();
                 peBinary.PdbParseException.Should().BeNull();
             }
         }
 
-        [Fact]
-        public void PEBinary_NoPdbAvailable()
+        [Theory]
+        [InlineData("clangcl.14.pe.c.codeview.pdbpagesize_8192_pdbmissing.exe")]
+        [InlineData("Native_x86_VS2013_PdbMissing.exe", true)]
```
marmegh, line 60: This type of test adds extra latency to test runtime. We can get the same effect without the latency by adding a foreach loop over a list containing the same strings within the test.

src/ReleaseHistory.md
dmachaj, line 5: [nit] "greater" mis-spelled as "grater".

marmegh:
5: should we then also update to check the EXE to identify the correct PDB? The customer who reported this will not be fixed with this change, then.

---
In reply to: [1203265011](https://github.com/microsoft/binskim/pull/685#issuecomment-1203265011) [](http://example.com/codeflow?ancestors=1203265011,1203265011)

shaopeng-gh:
5: --- No the code is fine we are using library which internally will do the lookup of the file name and load it. Just to note expect it may be different when we do troubleshooting.

---
In reply to: [1215392485](https://github.com/microsoft/binskim/pull/685#issuecomment-1215392485) [](http://example.com/codeflow?ancestors=1215392485,1203265011)

src/ReleaseHistory.md
shaopeng-gh: Fixed, thanks!

src/Test.UnitTests.BinaryParsers/PEBinary/PEBinaryTests.cs
shaopeng-gh: Oh I just discovered this syntax in the existing tests,
looks cool, did not know it is slow, let me change it to a loop.

marmegh:
I have a concern around additional time to run with these changes. Analyzing the four files provided by the customer for repro, took almost 2 minutes. Is this an indicator that we could see an increase in BinSkim timing out for customers with this change. @shaopeng-gh, did you experience any changes in time to run with this change?

marmegh:
Good news here, is that smaller PDBs appear to be parsed with a similar timing as before this change. The issue is going to come with PDBs >4K could result in binskim timeouts. Is there a way to decrease the load time for these?

---
In reply to: [1215773022](https://github.com/microsoft/binskim/pull/685#issuecomment-1215773022) [](http://example.com/codeflow?ancestors=1215773022)

shaopeng-gh:
The file was about 4G, it was slow due to size

---
In reply to: [1215773022](https://github.com/microsoft/binskim/pull/685#issuecomment-1215773022) [](http://example.com/codeflow?ancestors=1215773022,1215773022)

shaopeng-gh:
I think I might mixed up the question about the fix and the testcase.

the run time diff for the fix:

First, Pdb*Page*Size 4K does not indicate it is a huge pdb, 

case 1. normal pdb file size, but using Pdb*Page*Size 4k:
they were errored out, now they are analyzed, this will increase the total run time.
The impact of run time is similar to user has more files.

case 2. huge pdb file size, and using Pdb*Page*Size 4k:
they were errored out, now they are analyzed, this will increase the total run time.
This has big impact, the impact of run time I think also depend on disk speed, loading a large binary locally SSD may be faster than a small file on network share.

case 3&4. any pdb file size, and not using Pdb*Page*Size 4k:
same as before.

Yes, I agree it will increase time and increase the chance of timeout.


---
In reply to: [1215778899](https://github.com/microsoft/binskim/pull/685#issuecomment-1215778899) [](http://example.com/codeflow?ancestors=1215778899,1215773022,1215773022)

shaopeng-gh:
for the testcase:
I have changed to use the loop, and the test cases checked in are normal file size so they are fast.

---
In reply to: [1215849086](https://github.com/microsoft/binskim/pull/685#issuecomment-1215849086) [](http://example.com/codeflow?ancestors=1215849086,1215778899,1215773022,1215773022)

marmegh:
This is an awesome overview of the time consideration. Thank you for expanding on the details.

---
In reply to: [1215849086](https://github.com/microsoft/binskim/pull/685#issuecomment-1215849086) [](http://example.com/codeflow?ancestors=1215849086,1215773022)

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/MixedMode_x64_VS2015_Default.exe.sarif
marmegh: Shaopeng, if you haven't already, please create an Issue to track this suggestion.

marmegh (approved):
:shipit:

marmegh:
I've reviewed the changes and run them locally against the provided repros (both original and generated alternate test cases). @michaelcfanning, your opinion on the timeout risk is likely the main blocker to merging.

---
In reply to: [1215973925](https://github.com/microsoft/binskim/pull/685#issuecomment-1215973925) [](http://example.com/codeflow?ancestors=1215973925,1215773022)

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/MixedMode_x64_VS2015_Default.exe.sarif
shaopeng-gh: Done. 
https://github.com/microsoft/binskim/issues/694

PR #708, opened 2022-09-27
Expose linker command line

shaopeng-gh:
The **linker** command line was not used in any rules and it is hidden as a private field.
Now I have a new internal rule need to use it, so exposing this field the same way as **compiler** command line, 2 lines above.

Did not add release history because not impacting user.


marmegh (approved):
:shipit:

dmachaj (approved):

michaelcfanning (approved):
:shipit:

Issue #707, opened 2022-09-22
BinSkim Reports 32-bit Executables Built with CLR Support as Violating BA2015


JayOfManyTrades:
[BinSkim 1.9.4 testing for C++ binaries.xlsx](https://github.com/microsoft/binskim/files/9628464/BinSkim.1.9.4.testing.for.C%2B%2B.binaries.xlsx)

Dear "BinSkim" Development Team:

I (Jay Schmoll) performed testing and analysis to identify specifics about 'odd' behavior, which "BinSkim" 1.9.4 software exhibits for the "BA2015.EnableHighEntropyVirtualAddresses" and "BA2024.EnableSpectreMitigations" rules.  **I attached a workbook which contains details about my testing and analysis, a summary of which appears below.**

You clearly already know about the observations I made regarding "BA2024.EnableSpectreMitigations"; therefore, I do not have a request regarding them.  **However, I did _not_ find information about the behavior I observed for the "BA2015.EnableHighEntropyVirtualAddresses" rule and, therefore, have the following request.**

### MY REQUEST
Will you please modify "BinSkim" to―for 32-bit (x86) executable programs built from C++ source code **_with_** support for the CLR (Common Language Runtime)―_disregard_ the "BA2015.EnableHighEntropyVirtualAddresses" rule for the programs, like it already disregards the rule for 32-bit (x86) executable programs built from C++ source code **_without_** support for the CLR?

### "BA2015.EnableHighEntropyVirtualAddresses"
This rule applies only to 64-bit binaries.  "BinSkim" 1.9.4 disregards the rule for 32-bit:
- DLLs _whether or not_ they are built with CLR (Common Language Runtime) support.
- executables if they are built _without_ CLR support.

**However, "BinSkim" 1.9.4 undesirably reports 32-bit executables built with CLR support as violating the rule, even though they cannot (because they are 32-bit) support high entropy ASLR (address space layout randomization).**

### "BA2024.EnableSpectreMitigations"
"BinSkim" 1.9.4 reports binaries as violating this rule, even if they are built with spectre mitigation libraries if those libraries are debug (not release) libraries.  My testing and analysis is consistent with information available via the "https://github.com/microsoft/binskim/blob/v1.9.4/docs/RulesAndErrorsTroubleshootingGuide.md" URL's web page, which follows for convenience.
_If the error persists:_
_1. Stay up to date with Visual Studio version (use at least Visual Studio 2019)._
[...]

PR #716, opened 2022-10-17
Fixes for Issue #707 (BinSkim Reports 32-bit Executables Built with CLR Support as Violating BA2015)

JayOfManyTrades:
I created this pull request for my fixes for the "https://github.com/microsoft/binskim/issues/707" issue I submitted.  I very thoroughly tested the changes.

The "https://github.com/microsoft/binskim#submit-pull-requests" URL explains to "Submit a Pull Request to the 'develop' branch."  However, I do not see a `develop` branch in the primary "BinSkim" Git repository.

### DIFF
--- a/src/BinSkim.Rules/PERules/BA2015.EnableHighEntropyVirtualAddresses.cs
+++ b/src/BinSkim.Rules/PERules/BA2015.EnableHighEntropyVirtualAddresses.cs
@@ -82,6 +82,16 @@ namespace Microsoft.CodeAnalysis.IL.Rules
             CoffHeader coffHeader = target.PE.PEHeaders.CoffHeader;
             Characteristics characteristics = coffHeader.Characteristics;
 
+            if (characteristics.ToString() == "ExecutableImage, Bit32Machine")
+            {
+                ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


JayOfManyTrades:
@microsoft-github-policy-service agree

michaelcfanning:
Jay, hello. Sorry for the delay looking at this, we have fallen behind but will be catching up soon. Our plan is to get all open work wrapped up, tested and shipped the week of 11/11.

I think the code fix may be better applied in the `CanAnalyzePE` method. The distinction here is that CanAnalyze is intended to filter out any scan targets for which a specific security standard can't be applied. Any 32-bit binary (that exclusively loads as 32-bit) meets this standard.

The `Analyze` method is intended strictly for scan targets that can literally fail the security standard (i.e., which have the potential to apply a security setting but do not). As another comment, we also generally always report in these methods, i.e., any return here would be associated with a 'Pass' log.

Give me a moment and I'll refactor your change a bit, then ask for you to sanity check what I do. Stand by.

JayOfManyTrades:
Michael, thank you for your thorough, easy-to-understand reply!  Thanks also for offering to make the appropriate changes.  I gladly will test them after you have made them available.

michaelcfanning:
Closing this as replaced by #721

michaelcfanning:
Closing as #721 will push Jay's fix.

PR #719, opened 2022-10-19
BREAKING: Removed SARIF 1.0 support from BinSkim. Now option `-v | --sarif-output-version` does not accept value `OneZeroZero`.

shaopeng-gh:
BREAKING: Removed SARIF 1.0 support from BinSkim. Now option `-v | --sarif-output-version` does not accept value `OneZeroZero`.

I found most of the code is from SARIF SDK and unless we also zap the support in SDK we can not change those codes.
for example, the option itself is from SDK, and the help text will say "Valid values are OneZeroZero and Current".
We do not want that false information.
So a simple way of fix I tried working is to new a same name parameter to hide the underlying one, this way we can show something like "The only valid value is Current." and also only support Current instead of supporting both.
The error message user get this way is unified with other parameter not supported and will print the help page with the info on what is supported.
Added tests show the actual value will be `Current` as we want for various input.


src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -129,14 +129,16 @@ internal static string ReturnCommonPathRootFromTargetSpecifiersIfOneExists(IEnum
 
         public override int Run(AnalyzeOptions analyzeOptions)
         {
-            if (!Environment.GetCommandLineArgs().Any(arg => arg.Equals("--sarif-output-version")))
+            if (!Environment.GetCommandLineArgs().
+                Any(arg => arg.Equals("--sarif-output-version", StringComparison.OrdinalIgnoreCase) ||
+                arg.Equals("-v", StringComparison.OrdinalIgnoreCase)))
```
shaopeng-gh, line 134: this is not related to the change,
the code clearly missed the `-v` variant so I added it btw.

src/BinSkim.Driver/AnalyzeOptions.cs
```diff
@@ -48,5 +48,13 @@ public class AnalyzeOptions : AnalyzeOptionsBase
...
             HelpText = "If enabled, BinSkim won't break if we have a 'PdbLoadingException'.")]
         public bool IgnorePdbLoadError { get; set; }
+
+        [Option(
+            'v',
+            "sarif-output-version",
+            HelpText =
+            "The SARIF version of the output log file. The only valid value is Current.",
+            Default = BinSkimSarifVersion.Current)]
+        public new BinSkimSarifVersion SarifOutputVersion { get; set; } = BinSkimSarifVersion.Current;
```
shaopeng-gh, line 58: >new [](http://example.com/codeflow?start=15&length=3)

please see the desc of the PR for more info.

src/BinSkim.Driver/AnalyzeOptions.cs
```diff
@@ -48,5 +48,13 @@ public class AnalyzeOptions : AnalyzeOptionsBase
             "ignorePdbLoadError",
             HelpText = "If enabled, BinSkim won't break if we have a 'PdbLoadingException'.")]
         public bool IgnorePdbLoadError { get; set; }
+
+        [Option(
```
shaopeng-gh, line 52: >Option [](http://example.com/codeflow?start=9&length=6)

This is the main change of the PR.
The support of the v1 in Binskim, actually comes from SARIF SDK. Without remove the support in SDK we can not really remove it.
The change is to disable it in BinSkim. Please see more details in PR desc.
Now option `-v | --sarif-output-version` does not accept value `OneZeroZero` in BinSkim.

michaelcfanning: I don't think you need this. We should delete it. 

Instead, just find the right place in our code and throw an InvalidOperationException in cases where someone passes SarifOneZeroZero on the command-line.

src/BinSkim.Driver/BinSkimSarifVersion.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.CodeAnalysis.IL
+{
+    public enum BinSkimSarifVersion
```
michaelcfanning, line 6: >BinSkimSarifVersion [](http://example.com/codeflow?start=16&length=19)

I think we can delete this code.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -129,14 +129,16 @@ internal static string ReturnCommonPathRootFromTargetSpecifiersIfOneExists(IEnum
...
             {
-                analyzeOptions.SarifOutputVersion = Sarif.SarifVersion.Current;
```
michaelcfanning, line 134: >SarifOutputVersion [](http://example.com/codeflow?start=31&length=18)

put a check here for use of Sarif.SarifVersion.OneZeroZero and just throw an ArgumentException or InvalidOperationException in that case.

'BinSkim no longer supports emitting SARIF 1.0 (an obsolete format). Pass 'Current' on the command-line or omit this argument entirely.'

shaopeng-gh: I thought about this simple throw as well, I think my way is better because simple throw will still have the wrong info in the tool help.

Let me change to simple throw.

src/BinSkim.Driver/AnalyzeOptions.cs
shaopeng-gh: changed accordingly.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
shaopeng-gh: Changed.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -129,14 +129,23 @@ internal static string ReturnCommonPathRootFromTargetSpecifiersIfOneExists(IEnum
...
         {
-            if (!Environment.GetCommandLineArgs().Any(arg => arg.Equals("--sarif-output-version")))
+            if (!Environment.GetCommandLineArgs().
+                Any(arg => arg.Equals("--sarif-output-version", StringComparison.OrdinalIgnoreCase) ||
+                arg.Equals("-v", StringComparison.OrdinalIgnoreCase)))
             {
                 analyzeOptions.SarifOutputVersion = Sarif.SarifVersion.Current;
             }
 
-            if (s_UnitTestOutputVersion != Sarif.SarifVersion.Unknown)
```
shaopeng-gh, line 137: >s_UnitTestOutputVersion [](http://example.com/codeflow?start=16&length=23)

Multi test run have issue, I changed to instance.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
michaelcfanning: I see your point! We should be thinking about where this area is going. I think the right thing to do moving forward is to actually eliminate this argument entirely, i.e., we don't want the driver SDK itself to provide for SARIF v1 support by default. And so that argues for taking this approach: the help will be broken, but a future update of the driver SDK could drop this argument entirely.

michaelcfanning (approved):
:shipit:

PR #720, opened 2022-10-19
Add BA3003 fallback ELF stack canary check using symbol references. Fixes #712

### LINKED ISSUE #712: [RULE REQUEST] Symbol-based stack canary detection for ELFs (BA3003)
# Rule Proposal: Enhanced EnableStackProtector for Linux binaries

#### Rule metadata
- *Id*: BA3003
- *Name*: EnableStackProtector

#### Rule descriptions
- *Short*: No changes needed to the existing description.
- *Full*: No changes needed to the existing description.

#### Platform & applicability
- *Platform*: Linux
- *Applicable to*: Any ELF binary. Mostly useful for release builds that don't have DWARF information
- *Not applicable to*: Non-ELF binaries.

#### User-facing strings
- *Fail*: No changes needed.
- *Pass*: No changes needed.
- *[Other]*: No changes needed.

While using binskim to check some release builds for a Linux environment, I noticed that it was not reporting missing stack canaries on binaries that checksec flagged.

Adding the `--verbose` flag, I noticed this log:
```
notapplicable BA3003: 'binary-name-omitted' was not evaluated for check 'EnableStackProtector' as the analysis is not relevant based on observed metadata: not compiled with Dwarf debugging.
```

So, it seems binskim cannot check for stack canaries without DWARF information. We could possibly add a fallback check that implements something similar to checksec's method.

Here is how checksec implements canary detection: https://github.com/slimm609/checksec.sh/blob/932b9922d0f1d8bb48cf13453fd5d60157461ae0/src/functions/proccheck.sh#L23

Essentially, it's scanning through the ELF symbol table for references to `__stack_chk_fail`, `__stack_chk_guard`, `__intel_security_cookie`.

#### Implementation Notes

We likely just need to walk through the ELF header structures to find the symbol table. Then, parse the entries in the table and look for any stack check symbols we deem relevant. There are probably more robust ways of checking for stack cookies, but adopting checksec's method is a good start.

#### How to resolve

The existing guidance of adding `-fstack-protector-strong` or `-fstack-protector-all` is good.

connorcashmsft:
This changes BA3003 to check for any references to __stack_chk_fail, __stack_chk_guard, or __intel_security_cookie symbols if DWARF information is not present. That way we can still have some stack canary checking on all binaries. It's not as good as the compile argument check, as it can't tell us any information about how widespread the stack checks are. It only really tells us that they're checked *somewhere* in the binary.

The SARIF output changed a bit, which broke some tests. I copied over new versions for the expected values. The only meaningful changes I noticed were additional BA3003 items for binaries where BA3003 would be silently skipped due to lacking DWARF information.

I also had to replace some binaries under src/Test.FunctionalTests.BinSkim.Rules/FunctionalTestData/BA3003.EnableStackProtector, as some expected-fail binaries actually contained stack canaries and were now passing with this change. See the commit message here for more details: https://github.com/microsoft/binskim/commit/49025fa1b4803a153d682425da10c1d883d3887c. All tests are passing for me now.


connorcashmsft:
@microsoft-github-policy-service agree [company="Microsoft"]

connorcashmsft:

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
```diff
@@ -114,20 +126,49 @@ static bool analyze(IDwarfBinary binary, out List<DwarfCompileCommandLineInfo> f
                 return !failedList.Any();
             }
 
+            static bool analyzeSymbols(ElfBinary binary, out List<SymbolEntry<ulong>> failedList)
```
michaelcfanning, line 129: >analyzeSymbols [](http://example.com/codeflow?start=24&length=14)

I know you didn't establish this pattern but I am unclear: why are we preferring these static local functions? As opposed to simply using a helper method in this rule?

@shaopeng-gh do you know the history?

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
```diff
@@ -69,10 +76,15 @@ public override void Analyze(BinaryAnalyzerContext context)
             IDwarfBinary binary = context.DwarfBinary();
             List<DwarfCompileCommandLineInfo> failedList;
 
-            static bool analyze(IDwarfBinary binary, out List<DwarfCompileCommandLineInfo> failedList)
+            static bool analyzeDwarf(IDwarfBinary binary, out List<DwarfCompileCommandLineInfo> failedList)
             {
                 failedList = new List<DwarfCompileCommandLineInfo>();
 
+                if (binary.CommandLineInfos.Count < 1)
```
michaelcfanning, line 83: >CommandLineInfos [](http://example.com/codeflow?start=27&length=16)

Curious, why did you add this early exit? Strictly speaking, doesn't look necessary, am I correct on that? I notice you didn't introduce a similar early exit in the other static local function, is there a difference between the two cases?

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
```diff
@@ -114,20 +126,49 @@ static bool analyze(IDwarfBinary binary, out List<DwarfCompileCommandLineInfo> f
                 return !failedList.Any();
             }
 
+            static bool analyzeSymbols(ElfBinary binary, out List<SymbolEntry<ulong>> failedList)
+            {
+                failedList = new List<SymbolEntry<ulong>>();
+
+                var symbols = binary.ELF.Sections.FirstOrDefault(s => s.Type == SectionType.DynamicSymbolTable) as SymbolTable<ulong>;
```
michaelcfanning, line 133: >FirstOrDefault [](http://example.com/codeflow?start=50&length=14)

Why do we retrieve/scan only the first section that you retrieve? Does that mean that there is only a single one that meets these criteria?

Does this mean as well that we *always* expect to find one (and therefore enter the foreach that follows?

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
```diff
@@ -114,20 +126,49 @@ static bool analyze(IDwarfBinary binary, out List<DwarfCompileCommandLineInfo> f
...
+                foreach (SymbolEntry<ulong> symbol in symbols.Entries)
+                {
+                    if (symbol.Name != "__stack_chk_fail" && symbol.Name != "stack_chk_guard" && symbol.Name != "__intel_security_cookie")
+                    {
+                        failedList.Add(symbol);
+                    }
+                    else
+                    {
+                        return true;
```
michaelcfanning, line 142: >return [](http://example.com/codeflow?start=24&length=6)

So, what's our guarantee at this return location that failedList is empty?

michaelcfanning, line 142: >return [](http://example.com/codeflow?start=24&length=6)

OK, so I *think* what we're trying to do here is return true if we find any evidence of a stack protector function, yes? If so, fair enough. But why do we need to maintain a list of symbols that don't comprise a stack protector function? We wouldn't report on this,  right? 

If I'm correct, you can remove all references to the failed list and simply return false if we successfully exit the foreach loop.

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
```diff
@@ -114,20 +126,49 @@ static bool analyze(IDwarfBinary binary, out List<DwarfCompileCommandLineInfo> f
...
-                        RuleUtilities.BuildResult(FailureLevel.Error, context, null,
-                            nameof(RuleResources.BA3003_Error),
-                            context.TargetUri.GetFileName(),
-                            DwarfUtility.GetDistinctNames(failedList, context.TargetUri.GetFileName())));
-                    return;
+                    // Analysis using DWARF info failed.
+                    // This could be because the binary simply doesn't have DWARF info.
+                    // So we'll fall back to a symbol search similar to checksec, just to be sure.
+                    List<SymbolEntry<ulong>> failedList_symbols;
+                    if (!analyzeSymbols(elf, out failedList_symbols))
```
michaelcfanning, line 156: >failedList_symbols [](http://example.com/codeflow?start=49&length=18)

Yes, I see no utilization of this data, so looks like we can drop it. Let's move these static local functions into standard C# methods as well. Unless someone can clarify the specific technical value of embedding them locally, of course. :)

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/Binskim.linux-x64.dll.sarif
```diff
@@ -210,10 +212,7 @@
                 "equivalentBinScopeRuleReadableName": "WXCheck"
               }
             }
-          ],
-          "properties": {
```
michaelcfanning, line 214: >properties [](http://example.com/codeflow?start=11&length=10)

I will do you a favor and rebaseline main quickly. That will allow us to be better focused on your changes. Stand by and I will advice when you can merge from main.

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/Binskim.linux-x64.dll.sarif
```diff
@@ -96,6 +96,8 @@
       "tool": {
         "driver": {
           "name": "testhost",
+          "organization": "Microsoft Corporation",
```
michaelcfanning, line 99: >Microsoft [](http://example.com/codeflow?start=27&length=9)

OK, so what actually appears to have happened here is that you copied over test output from our literal test run?

Instead, try doing this from a console window (starting at the root of your BinSkim enlistment):

cd .\src\Test.FunctionalTests.BinSkim.Driver
UpdateBaselines.cmd

This should undo most of these changes once you commit them.

What I did was to enlist clean in main, run baseline update and confirmed the checked in files are good.

We could try to be a bit more friendly and rewrite the test output so that it doesn't have the deltas that burned you, that would be a nice tech debt clean-up later...

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
connorcashmsft: Yeah it's not *really* necessary. I just wanted to be explicit in returning false if there is no DWARF command line information. It could be added to `analyzeSymbols` too without changing the behavior.

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
connorcashmsft: My understanding is that ELFs should only have one dynamic symbol table. I don't actually know how well that assumption holds up in practice. It would be easy enough to simply iterate through any sections that ELFSharp identifies as dynamic symbol tables.

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
connorcashmsft: I might be misunderstanding the purpose of failedList here. I was basing my usage off of `analyzeDwarf`, which appears to simply store any command line args that are not stack-protector related in `failedList`. So, I'm simply storing any symbols that are not stack check related in my failedList. This might not actually be helpful for anything down the line, I just wanted to make this check as similar as possible to the existing one. I would expect there will be items in failedList for all binaries.

connorcashmsft: Ah ok. Yeah I don't think we really need that list of symbols. I'll remove it.

shaopeng-gh:
Looking...

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
shaopeng-gh: Not sure about history why we had that style, but the current status may need further improvement, 1. VS will suggest static if possible 2. should add modifier:
image


What I think normal style should be:
image

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
```diff
@@ -114,20 +126,49 @@ static bool analyze(IDwarfBinary binary, out List<DwarfCompileCommandLineInfo> f
...
-                    // Modules did not meet the criteria: {1}
-                    context.Logger.Log(this,
-                        RuleUtilities.BuildResult(FailureLevel.Error, context, null,
-                            nameof(RuleResources.BA3003_Error),
-                            context.TargetUri.GetFileName(),
-                            DwarfUtility.GetDistinctNames(failedList, context.TargetUri.GetFileName())));
-                    return;
+                    // Analysis using DWARF info failed.
+                    // This could be because the binary simply doesn't have DWARF info.
+                    // So we'll fall back to a symbol search similar to checksec, just to be sure.
```
shaopeng-gh, line 149: I have question about this.
A DWARF check founds `-fno-stack-protectors` and returns false. It should be different with no DWARF and returns false. Do we still to check by symbols.

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
```diff
@@ -64,70 +71,98 @@ public override AnalysisApplicability CanAnalyzeDwarf(IDwarfBinary target, Sarif
...
+        {
+            foreach (ISection section in binary.ELF.Sections)
+            {
+                if (section.Type == SectionType.DynamicSymbolTable)
+                {
+                    var symbols = section as SymbolTable<ulong>;
+                    foreach (SymbolEntry<ulong> symbol in symbols.Entries)
                     {
-                        failedList.Add(info);
+                        if (symbol.Name == "__stack_chk_fail" || symbol.Name == "stack_chk_guard" || symbol.Name == "__intel_security_cookie")
```
shaopeng-gh, line 128: >if (symbol.Name == "__stack_chk_fail" || symbol.Name == "stack_chk_guard" || symbol.Name == "__intel_security_cookie") [](http://example.com/codeflow?start=24&length=118)

Sorry not very familiar, can you provide webpage/doc about these 3 are the ones that is should be at least one present and also can indicates Stack Protector enabled with proper level.

connorcashmsft: I'm not aware of any documentation that concretely describes how stack cookies are implemented at this level, unfortunately. This is based off checksec's implementation here: https://github.com/slimm609/checksec.sh/blob/main/checksec#L679. From my manual analysis, I've always seen __stack_chk_fail from libc being used to trigger a stack check failure. This method alone cannot detect the stack protector level used, as we're effectively just checking for the presence of *some* stack check in the entire binary. That's why I'm treating this as a fallback. I did just notice that I typo'd stack_chk_guard; it needs to be __stack_chk_guard. I'll fix that.

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
shaopeng-gh: just throwing thoughts for discussion, maybe
check if have DWARF info
1. if have, only check by DWARF
2. if not, only check by symbol

shaopeng-gh:
@connorcashmsft Thanks for contributing! 
I have many questions myself with the details, I will get your branch and do some experiments. :slightly_smiling_face:

shaopeng-gh:
>                         string[] paramToCheck = { "--param=ssp-buffer-size=" };

(Note please don't rush to change according to my comments are for discussion only need Michael's decision)

Users can control how/when GCC inserts stack protectors with --param=ssp-buffer-size=N. If N is set to a very large number, then effectively stack protectors are not used at all and should be 4 or less. 

For the `+` in the test file it means binary has 2 files one is using 8 and one is using 4 it should fail.









---
Refers to: src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs:104 in 49025fa. [](commit_id = 49025fa1b4803a153d682425da10c1d883d3887c, deletion_comment = False)

shaopeng-gh (changes requested):
:clock1:

connorcashmsft:
> For new test binaries adding to the repo please add details in https://github.com/microsoft/binskim/blob/main/docs/FunctionalTestBuildScripts.md by alphabetical order. (Unfortunately we start this process only recently so don't have all existing ones.)

I've added docs for the two binaries I changed. `gcc.example2.fstackprotectorall+fnostackprotector` will need to be changed now that you've explained the intention behind the naming convention. It currently is incorrect. I also removed `gcc.example2.fstackprotectorstrongsspbuffersize8+fstackprotectorstrongsspbuffersize4`, so we'll have to replace that one and document it.

I think we can use the original binaries as-is if I adjust the fallback mechanism to not use the symbol check at all if DWARF is present. I'm not 100% certain what compiler flags were used for those, but I can guess based on the names. I don't have the original source for those either. The source I provided was a quick attempt at reverse engineering it from the binaries. I'm not sure how it was split up into two files for the gcc.example2.fstackprotectorall+fnostackprotector and gcc.example2.fstackprotectorstrongsspbuffersize8+fstackprotectorstrongsspbuffersize4. We could also just start from scratch, I suppose.

shaopeng-gh:
> 

Question, the fallback mechanism use the symbol check,
1. if a binary contains 10 .C file and only 1 .C file is compiled with fstackprotectorall and all others with fnostackprotector, will your symbol check work with this case
2. Do you know or tried if this symbol check method will work in the case of:
Users can control how/when GCC inserts stack protectors with --param=ssp-buffer-size=N. If N is set to a very large number, then effectively stack protectors are not used at all

connorcashmsft:
> 
> Question, the fallback mechanism use the symbol check,
> 
> 1. if a binary contains 10 .C file and only 1 .C file is compiled with fstackprotectorall and all others with fnostackprotector, will your symbol check work with this case
> 2. Do you know or tried if this symbol check method will work in the case of:
>    Users can control how/when GCC inserts stack protectors with --param=ssp-buffer-size=N. If N is set to a very large number, then effectively stack protectors are not used at all

For case 1 it will report a positive result. We can't distinguish where canaries are being used with this method, only that they are used in one or more places, essentially. The trade-off with this method is that it should work on any binary regardless of debug information, but it does not provide much fidelity.

For case 2, how large of a buffer size do you need for it to revert to no protection, and does this failure mode depend on any other parameters? I just tested with sizes of 4, 8, 16, and 1024. All of them appear to have canaries according to this method. It could very well be missing some nuance though. Do you have more information on how this parameter will break it, or an example binary? I can look into the assembly by hand if there's a good example somewhere. I simply compiled with `gcc --param=ssp-buffer-size=1024`, so it should be using the default stack protector settings for any other parameters.

shaopeng-gh:
> > Question, the fallback mechanism use the symbol check,
> > 
> > 1. if a binary contains 10 .C file and only 1 .C file is compiled with fstackprotectorall and all others with fnostackprotector, will your symbol check work with this case
> > 2. Do you know or tried if this symbol check method will work in the case of:
> >    Users can control how/when GCC inserts stack protectors with --param=ssp-buffer-size=N. If N is set to a very large number, then effectively stack protectors are not used at all
> 
> For case 1 it will report a positive result. We can't distinguish where canaries are being used with this method, only that they are used in one or more places, essentially. So if any single C file or object file uses canaries, it will report no errors for the whole thing. The trade-off with this method is that it should work on any binary regardless of debug information, but it does not provide much fidelity.
> 
> For case 2, how large of a buffer size do you need for it to revert to no protection, and does this failure mode depend on any other parameters? I just tested with sizes of 4, 8, 16, and 1024. All of them appear to have canaries according to this method. It could very well be missing some nuance though. Do you have more information on how this parameter will break it, or an example binary? I can look into the assembly by hand if there's a good example somewhere. I simply compiled with `gcc --param=ssp-buffer-size=1024`, so it should be using the default stack protector settings for any other parameters.

1. thanks, just to make sure in a same page. 
2. Can you try `-fstack-protector --param=ssp-buffer-size=1024 -g -gdwarf-5`
per https://gcc.gnu.org/onlinedocs/gcc-12.2.0/gcc/Optimize-Options.html#Optimize-Options
`The minimum size of buffers (i.e. arrays) that receive stack smashing protection when -fstack-protector is used.`
I understand we getting the symbol but I am not sure if it is getting effective at all in this case.

(My questions are primary to make sure in a same page )

connorcashmsft:
Ah, my test program is using a buffer of size 0x1000. So with ssp-buffer-size=1024, canaries were still being added since the buffer was larger than the threshold. When I reduce the buffer size down to 0x10, canaries go away in my test program. But, I would imagine canaries will still be detected as long as there's at least one buffer bigger than the threshold.

shaopeng-gh:
> Ah, my test program is using a buffer of size 0x1000. So with ssp-buffer-size=1024, canaries were still being added since the buffer was larger than the threshold. When I reduce the buffer size down to 0x10, canaries go away in my test program. But, I would imagine canaries will still be detected as long as there's at least one buffer bigger than the threshold.

Great I believe we are in the same page for above 2,
@michaelcfanning I have reviewed, the input from my side is the symbol test will be inaccurate for above 2 reasons,
to add this I would suggest to:
1. change the current fallback logic to only do it only when there is no DWARF command line info to check. When with proper DWARF info the check returns fail for any reason, should not do a second symbol check that may result in pass.
2. With the logic fixed, I believe the above 2 scenario old test cases still works and we should keep them as they are good tests to spot 2 issues. :)  
And add new ones as needed.

shaopeng-gh:
@connorcashmsft `only do it only when there is no DWARF command line info to check`
fyi my thought about this is:

```CSharp
 var validGccCommandLineInfos = new List<DwarfCompileCommandLineInfo>();
  foreach (DwarfCompileCommandLineInfo info in binary.CommandLineInfos)
  {
      if (ElfUtility.GetDwarfCommandLineType(info.CommandLine) != DwarfCommandLineType.Gcc)
      {
          continue;
      }
      validGccCommandLineInfos.Add(info);
  }
  if (validGccCommandLineInfos.Count > 1)
  {
      // Check by Dwarf, refractor existing method and pass in validGccCommandLineInfos
  }
  else
  {
      // Check by symbol, your method
  }
```

connorcashmsft:
> @connorcashmsft `only do it only when there is no DWARF command line info to check` fyi my thought about this is:
> 
> ```cs
>  var validGccCommandLineInfos = new List<DwarfCompileCommandLineInfo>();
>   foreach (DwarfCompileCommandLineInfo info in binary.CommandLineInfos)
>   {
>       if (ElfUtility.GetDwarfCommandLineType(info.CommandLine) != DwarfCommandLineType.Gcc)
>       {
>           continue;
>       }
>       validGccCommandLineInfos.Add(info);
>   }
>   if (validGccCommandLineInfos.Count > 0)
>   {
>       // Check by Dwarf, refractor existing method and pass in validGccCommandLineInfos
>   }
>   else
>   {
>       // Check by symbol, your method
>   }
> ```

Alright, I've switched it to use this logic, and updated AnalyzeDwarf to take an optional list of DwarfCompileCommandLineInfo objects. Updated test baselines too.

I'll revert those test binaries I changed to their old versions now, and see if they still work as expected.

docs/FunctionalTestBuildScripts.md
```diff
@@ -70,6 +70,36 @@ A simple hello world program, compiled with `clang 14.0.0` that generates a .exe
 `clang-cl -o clangcl.14.pe.c.codeview.pdbpagesize_default.exe -fuse-ld=lld-link helloc.c -m32 -Z7 -MTd /link /CETCOMPAT /guard:cf /PDB:clangcl.14.pe.c.codeview.pdbpagesize_default.exe.pdb`  
 `clang-cl -o clangcl.14.pe.c.codeview.pdbpagesize_8192_pdbmissing.exe -fuse-ld=lld-link helloc.c -m32 -Z7 -MTd /link /CETCOMPAT /guard:cf /PdbPageSize:8192 /PDB:clangcl.14.pe.c.codeview.pdbpagesize_8192_pdbmissing.exe.pdb (then delete the pdb)`
 
+## gcc.example2.fnostackprotector
```
shaopeng-gh, line 73: >example2 [](http://example.com/codeflow?start=7&length=8)

nit: use name with exmaple1 since this is only 1 file, too keep with existing pattern

docs/FunctionalTestBuildScripts.md
```diff
@@ -70,6 +70,36 @@ A simple hello world program, compiled with `clang 14.0.0` that generates a .exe
...
+Compiled with gcc 9.4.0 using:
+`gcc -g -fno-stack-protector example2.c -o gcc.example2.fnostackprotector`
+
+## gcc.example2.fstackprotectorall+fnostackprotector
+
+This uses the same source as gcc.example2.fnostackprotector.
+
+Compiled with gcc 9.4.0 using:
+`gcc -g -fstack-protector-all -fno-stack-protector binskim_test.c -o gcc.example2.fstackprotectorall+fnostackprotector`
+
```
shaopeng-gh, line 102: in you latest checkin, 
1. I see the old file is deleted, can you see if you can keep it?
2. I don't see a new file added, I think we should delete this info in the md file , if you add file please use new name. For old test files not added by you no need to add info in md file.
3. fyi `-fstack-protector-all -fno-stack-protector` together was not the way, it was 2 file one with `-fstack-protector-all` and the other with `-fno-stack-protector`.

shaopeng-gh: @connorcashmsft please wait, let me work on the tests for you.

shaopeng-gh: I will create a PR to your branch. Hold on.

docs/FunctionalTestBuildScripts.md
shaopeng-gh: pleas wait my PR.

docs/FunctionalTestBuildScripts.md
shaopeng-gh: since you added two new code path :
no dwarf -> pass
no dwarf -> fail
, the test file you added is still to the old path that has dwarf , I am working on that as well. I will create a PR to your branch today.

shaopeng-gh: Created a PR to your branch with my thoughts.
also I fixed a from 2018 old bug that is related to our tests.  :)
[https://github.com/connorcashmsft/binskim/pull/1](https://github.com/connorcashmsft/binskim/pull/1)

connorcashmsft: I just merged in your changes.

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
```diff
@@ -43,9 +48,11 @@
~        {
~            CanAnalyzeDwarfResult result = default;
```
shaopeng-gh: >elf [](http://example.com/codeflow?start=36&length=3)

Please fix LGTM alert:
1. elf is not needed assign since now it is not used.

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
```diff
@@ -64,70 +71,123 @@
...
+            }
+
+            return !failedList.Any();
+        }
+
+        private static bool AnalyzeSymbols(ElfBinary binary)
+        {
+            foreach (ISection section in binary.ELF.Sections)
+            {
+                if (section.Type == SectionType.DynamicSymbolTable)
```
shaopeng-gh, line 128: >var symbols = section as SymbolTable<ulong>; [](http://example.com/codeflow?start=20&length=44)

LGTM alret 2:
Variable `symbols` may be null, when used below. Some why of check and deal with null is needed.

shaopeng-gh:
Thanks for looking into my two main concerns of
1. only check by symbol when there is no dwarf commandline info
2. tests
Since build may not run is PR is from fork, I get and build locally `.\BuildAndTest.cmd` test passed.
Please fix those 2 LGTM alerts,
and any open comments from Michael if any.

shaopeng-gh (approved):
:shipit:

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
connorcashmsft: Fixed

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
connorcashmsft: fixed

connorcashmsft:
LGTM issues should be fixed. We'll see once it finishes this latest run. I believe all of Michael's concerns are addressed, but please let me know if I missed one or if something new comes up.

PR #736, opened 2022-11-13
* BUGFIX: Fix command line parameter in documents: `-Wl,z,relro` with `-Wl,-z,relro`, and `-Wl,z,now` with `-Wl,-z,now`.

shaopeng-gh:
More info:
https://wiki.debian.org/Hardening
`via gcc with -Wl,-z,now`
`via gcc with -Wl,-z,relro`
This change is the only change in this pr.

### DIFF
--- a/docs/BinSkimRules.md
+++ b/docs/BinSkimRules.md
@@ -156,7 +156,7 @@ The non-executable stack is not enabled for this binary, so '{0}' can have a vul
 
 ### Description
 
-This check ensures that some relocation data is marked as read only after the executable is loaded, and moved below the '.data' section in memory. This prevents them from being overwritten, which can redirect control flow. Use the compiler flags '-Wl,z,relro' to enable this.
+This check ensures that some relocation data is marked as read only after the executable is loaded, and moved below the '.data' section in memory. This prevents them from being overwritten, which can redirect control flow. Use the compiler flags '-Wl,-z,relro' to enable this.
 
 ### Messages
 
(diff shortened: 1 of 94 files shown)


docs/BinSkimRules.md
```diff
@@ -156,7 +156,7 @@ The non-executable stack is not enabled for this binary, so '{0}' can have a vul
 
 ### Description
 
-This check ensures that some relocation data is marked as read only after the executable is loaded, and moved below the '.data' section in memory. This prevents them from being overwritten, which can redirect control flow. Use the compiler flags '-Wl,z,relro' to enable this.
+This check ensures that some relocation data is marked as read only after the executable is loaded, and moved below the '.data' section in memory. This prevents them from being overwritten, which can redirect control flow. Use the compiler flags '-Wl,-z,relro' to enable this.
```
michaelcfanning, line 159: >check [](http://example.com/codeflow?start=5&length=5)

OK, so we clearly need to correct this typo, agreed and I've approved this change. 

Isn't it true, though, that the -Wl convention is for passing linker args specifically from the Clang compiler? And that other compilers support directly passing things like the noexecstack flag?

michaelcfanning (approved):
:shipit:

docs/BinSkimRules.md
shaopeng-gh: approved --- Thanks!

I think -Wl convention is for both:

https://gcc.gnu.org/onlinedocs/gcc-12.2.0/gcc/Link-Options.html#Link-Options
> -Wl,option
> Pass option as an option to the linker. If option contains commas, it is split into multiple options at the commas. You can use this syntax to pass an argument to the option. For example, -Wl,-Map,output.map passes -Map output.map to the linker. When using the GNU linker, you can also get the same effect with -Wl,-Map=output.map.

https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-wl-arg-arg2
> -Wl,`<arg>`,`<arg2>`...
> Pass the comma separated arguments in <arg> to the linker

I looked noexecstack rule, it currently use `-z noexecstack`, it will work, but `-Wl,-z,noexecstack` is said to the standard way and should be used instead. I think to create a PR to change it, let me know if sounds good @michaelcfanning 
Also, the this noexecstack not only can be enabled by linker but also assembler `-Wa,--noexecstack`, let me know if you like me to test it out and include in the doc, "fix by ... or ..."

PR #740, opened 2022-11-15
Fix Linux test issue

shaopeng-gh:
Linux test has been randomly success and fail, only starting these few days.
We have no related changes, wondering if there is new VM image testing.
The error is the test is randomly looking for AnyCpu and x64 folder of the dll and can not find it when it thinks AnyCpu is the one to look for.
(note: dotnet test ignores /p:platform so that is not the way to fix this)
The fix simply: Build as x64 and test x64 dll by specify path.
Test: have run the branch build 10 times all success. while main mostly fails.

### DIFF
--- a/BuildAndTest.sh
+++ b/BuildAndTest.sh
@@ -6,9 +6,9 @@ if [[  "$(uname)" == "Linux" || "$(uname)" == "Darwin" ]]; then
   sed 's#\\#/#g' src/BinSkim.sln > src/BinSkimUnix.sln
 fi
 
-dotnet build src/BinSkimUnix.sln --configuration Release /p:Platform="x64"
+# dotnet build src/BinSkimUnix.sln --configuration Release /p:Platform="x64"
 
-dotnet test --no-build
+dotnet test src/BinSkimUnix.sln Release /p:Platform="x64"
 
 # dotnet test src/Test.FunctionalTests.BinSkim.Driver/Test.FunctionalTests.BinSkim.Driver.csproj --no-build --configuration Release /p:Platform="x64"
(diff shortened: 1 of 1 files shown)


BuildAndTest.sh
```diff
@@ -6,9 +6,9 @@ if [[  "$(uname)" == "Linux" || "$(uname)" == "Darwin" ]]; then
...
 
-dotnet build src/BinSkimUnix.sln --configuration Release
+dotnet build src/BinSkimUnix.sln --configuration Release /p:Platform="x64"
 
-dotnet test src/Test.FunctionalTests.BinSkim.Driver/Test.FunctionalTests.BinSkim.Driver.csproj --no-build --configuration Release /p:Platform="x64"
-dotnet test src/Test.FunctionalTests.BinSkim.Rules/Test.FunctionalTests.BinSkim.Rules.csproj --no-build --configuration Release /p:Platform="x64"
-dotnet test src/Test.UnitTests.BinaryParsers/Test.UnitTests.BinaryParsers.csproj --no-build --configuration Release /p:Platform="x64"
-dotnet test src/Test.UnitTests.BinSkim.Rules/Test.UnitTests.BinSkim.Rules.csproj --no-build --configuration Release /p:Platform="x64"
\ No newline at end of file
+dotnet test bld/bin/x64_Release/netcoreapp3.1/Test.FunctionalTests.BinSkim.Driver.dll
```
shaopeng-gh, line 11: >x64_Release [](http://example.com/codeflow?start=20&length=11)

please see the pr description for details.

michaelcfanning (approved):

PR #744, opened 2022-11-23
Fix BA3003.EnableStackProtector false positive for static linked file

shaopeng-gh:
During the testing of GoLang binaries I noticed static linked file fails the BA3003 rule.
I think the check may need to check both DynamicSymbolTable and SymbolTable. @connorcashmsft let me know if you think this is correct.
Created test files for C as well.
related pr https://github.com/microsoft/binskim/pull/720.


michaelcfanning (approved):
:shipit:

PR #760, opened 2022-12-08
BUGFIX: Prevent exit with error `ERR997.ExceptionLoadingAnalysisTarget` when there is exception parsing Dwarf debugging info.

shaopeng-gh:
1. Prevent exit with error `ERR997.ExceptionLoadingAnalysisTarget` when there is exception parsing Dwarf debugging info.
2. Disable related check when it happens.
3. I need to figure out how to create test files that can trigger this exception.

### DIFF
--- a/src/BinaryParsers/ElfBinary/ElfBinary.cs
+++ b/src/BinaryParsers/ElfBinary/ElfBinary.cs
@@ -56,7 +56,15 @@ namespace Microsoft.CodeAnalysis.BinaryParsers
                 PublicSymbols = publicSymbols;
                 SectionRegions = ELF.Sections.Where(s => s.LoadAddress > 0).OrderBy(s => s.LoadAddress).ToArray();
 
-                CompilationUnits = DwarfSymbolProvider.ParseAllCompilationUnits(this, DebugData, DebugDataDescription, DebugDataStrings, NormalizeAddress);
+                try
+                {
+                    CompilationUnits = DwarfSymbolProvider.ParseAllCompilationUnits(this, DebugData, DebugDataDescription, DebugDataStrings, NormalizeAddress);
+                }
+                catch (ArgumentException)
+                {
+                    ErrorParsingCompilationUnits = true;
+                }
+
                 commandLineInfos = new Lazy<List<DwarfCompileCommandLineInfo>>(()
                     => DwarfSymbolProvider.ParseAllCommandLineInfos(CompilationUnits));
                 LineNumberPrograms = DwarfSymbolProvider.ParseLineNumberPrograms(DebugLine, NormalizeAddress);
@@ -238,6 +246,11 @@ namespace Microsoft.CodeAnalysis.BinaryParsers
         /// Gets or sets the CompilationUnits.
         /// </summary>
         public List<DwarfCompilationUnit> CompilationUnits { get; set; } = new List<DwarfCompilationUnit>();
+
+        /// <summary>
+        /// If there is error parsing Compilation Units.
+        /// </summary>
+        public bool ErrorParsingCompilationUnits { get; set; } = false;
 
         /// <summary>
         /// Gets or sets the CommandLineInfos.
--- a/src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
+++ b/src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
@@ -48,6 +48,12 @@ namespace Microsoft.CodeAnalysis.IL.Rules
         {
             CanAnalyzeDwarfResult result = default;
 
+            if (target.ErrorParsingCompilationUnits)
+            {
+                reasonForNotAnalyzing = MetadataConditions.ErrorParsingCompilationUnits;
+                return AnalysisApplicability.NotApplicableToSpecifiedTarget;
+            }
+
             if (target is ElfBinary)
             {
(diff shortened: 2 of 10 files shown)


src/BinaryParsers/ElfBinary/ElfBinary.cs
```diff
@@ -56,7 +56,15 @@ public ElfBinary(Uri uri, string localSymbolDirectories = null) : base(uri)
                 PublicSymbols = publicSymbols;
                 SectionRegions = ELF.Sections.Where(s => s.LoadAddress > 0).OrderBy(s => s.LoadAddress).ToArray();
 
-                CompilationUnits = DwarfSymbolProvider.ParseAllCompilationUnits(this, DebugData, DebugDataDescription, DebugDataStrings, NormalizeAddress);
+                try
```
michaelcfanning, line 59: >try [](http://example.com/codeflow?start=16&length=3)

There is a simpler change, you should use the lazy list, like on line 69. This way, when we disable relevant rules, we won't hit the problematic code path.

Note that in general, we might consider deferring all parsing unless/until it's necessary. To be consistent and prevent unnecessary upfront work/

michaelcfanning (changes requested):
:clock1:

src/BinaryParsers/ElfBinary/ElfBinary.cs
shaopeng-gh: @michaelcfanning my current code is:
when we can't not read it without exception, we programmatically disable 3003 and 3005.

Do you mean: remove the try catch and also remove the code programmatically disable 3003 and 3005, 
just ask user to use config file to manually disable rule 003 and 3005? And we only make one change: lazy loading here.

michaelcfanning: Yes! The problem here is that we're raising an exception right on binary load, because we aggressively parse data in the binary. 

If we performed a lazy initialization, then any rule that attempted to parse (and raised an exception) would be disabled and analysis would continue. This is a better behavior.

Next, as you observe, users could disable specific rules in order to make progress.

Now, unfortunately 1.9.5 doesn't permit this because of our aggressive parsing. :) That's ok, let's make the lazy load change and keep moving. We can disable 3003 and 3005 in the 2.0 release if can't figure out what's going on with the linkage name issue...

shaopeng-gh: @michaelcfanning 
I changed to Lazying loading locally, it works but still get triggered and loaded a few lines below:
`LoadDebug(localSymbolDirectories);`

This method depends on CompilationUnits to load the actually file by `Dwo`, or by `DebugLink`

yongyan-gh: looks like when `CanAnalyzeDwarf` is called, the `IDwarfBinary` is already initialized and populated the `CompilationUnits` value.

when the elf binary has satellite dwo files, it parses all the dwo files during the constructor methods.

shaopeng-gh: Looking at the code related to another issue, it appears that BinSkim aggressively loads and inspects a lot of data for every ELF binary.
---- Now that I understand your concern, we should think about the current code and refactor.
I will look into it.

michaelcfanning: OK! But note there is a one line change you could make here and drop all the other code, simply use the Lazy<T> pattern that's already in place:

CompilationUnit  = new Lazy<List<WHATEVERTYPEISRETURNED>>(()
                    => DwarfSymbolProvider.ParseAllCompilationUnits(this, DebugData, DebugDataDescription, DebugDataStrings, NormalizeAddress);

shaopeng-gh: No this will not be enough, 
I think the goal is: if we use a config file that disable all rule and leave one PE rule, the code that trigger the error will not be executed. I have tested this will still be triggered.

Lazy loading only delay the loading to where the property is read, so we need to find all the places the property is read, if there is one that is in the code path in will get triggered.

shaopeng-gh: @michaelcfanning 

I created a draft to read the compilation unit only when need:
https://github.com/microsoft/binskim/pull/762
The logic is not changed and tests pass,
There is still problem with this, the compilation unit read still gets triggered always because of this common code:

```CSharp
public sealed override AnalysisApplicability CanAnalyze(BinaryAnalyzerContext context, out string reasonForNotAnalyzing)
        {
            if (context.IsELF())
            {
                ElfBinary target = context.ElfBinary();

                if (target.IsDebugOnlyFile)
                {
                    reasonForNotAnalyzing = MetadataConditions.ImageIsDebugOnly;
                    return AnalysisApplicability.NotApplicableToSpecifiedTarget;
                }

                return this.CanAnalyzeElf(target, context.Policy, out reasonForNotAnalyzing);
            }
            else
            {
```

the target.IsDebugOnlyFile will need to read compilation unit to know.

Is it possible that we are over-reading? ---- back to this question I think compilation unit is important info to read.

shaopeng-gh: Change to another approach https://github.com/microsoft/binskim/pull/769
closing this.

PR #761, opened 2022-12-08
Ba2024 reporting improvements.

michaelcfanning:
Cite the <SpectreMitigation> build property and also mention installing spectre-mitigated C runtime libs from the VS installer.


src/BinaryParsers/ElfBinary/Dwarf/DwarfCompilationUnit.cs
```diff
@@ -194,6 +195,12 @@ private void ReadData(IDwarfBinary dwarfBinary, DwarfMemoryReader debugData, Dwa
                             attributeValue.Value = debugStrings.ReadString(offsetStrp);
                             break;
 
+                        case DwarfFormat.StrpSup:
```
michaelcfanning, line 198: >StrpSup [](http://example.com/codeflow?start=41&length=7)

This case wasn't handled previously. It wasn't causing our crash, as it happens, but I fixed it anyway.

src/BinaryParsers/ElfBinary/Dwarf/DwarfCompilationUnit.cs
```diff
@@ -250,6 +257,7 @@ private void ReadData(IDwarfBinary dwarfBinary, DwarfMemoryReader debugData, Dwa
                             break;
 
                         case DwarfFormat.ImplicitConst:
+                            attributeValue.Value = descriptionAttribute.Value;
```
michaelcfanning, line 260: >Value [](http://example.com/codeflow?start=72&length=5)

Now that we properly obtain the implicit const value from the debug abbreviations table, we can populate it here.

src/BinaryParsers/ElfBinary/Dwarf/DwarfCompilationUnit.cs
```diff
@@ -270,6 +278,7 @@ private void ReadData(IDwarfBinary dwarfBinary, DwarfMemoryReader debugData, Dwa
                             break;
 
                         default:
+                            Console.WriteLine($"No case for format {format}");
```
michaelcfanning, line 281: >Console [](http://example.com/codeflow?start=28&length=7)

This debug output was helpful as it reports out on unknown format types. We should remove it though, it shouldn't be in production.

src/BinaryParsers/ElfBinary/Dwarf/DwarfEnums.cs
```diff
@@ -183,17 +183,23 @@ public enum DwarfAttribute
         Sibling = 0x01,
         Location = 0x02,
         Name = 0x03,
+        Reserved0 = 0x4,
```
michaelcfanning, line 186: >Reserved0 [](http://example.com/codeflow?start=8&length=9)

I populated more values here taken from the DWARF5 spec. None of this work proved to be essential to the actual bug

PR #763, opened 2022-12-12
BugFix assertion failed with no clue when TargetFileSpecifiers is null or empty for BinSkim analyze

LingZhou-gh:
Before this change, when no Target file is specified for BinSkim analyze, BinSkim will fail with no clue to users (as shown below)
image

The change is to check whether Target file is specified at the very beginning of running the analyze command. If target file is not specified, throw an exception with clear message.

After the change, when no target file is specified for BinSkim analyze, BinSkim will fail with clear clue (as shown below).
image

Add/Modify the test cases.

### DIFF
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -128,11 +128,10 @@ namespace Microsoft.CodeAnalysis.IL
         }
 
         public override int Run(AnalyzeOptions analyzeOptions)
-        {
-            // Check whether target file is specified for BinSkim Analyze. If not, exit with code 1 (failure).
+        {            
             if (analyzeOptions.TargetFileSpecifiers == null || !analyzeOptions.TargetFileSpecifiers.Any())
             {
-                throw new ArgumentNullException("Please specify a targe file after BinSkim analyze", (Exception)null);
+                throw new ArgumentNullException("Please specify one or more files, directories, or filter patterns for BinSkim analyze.", (Exception)null);
             }
 
             if (!Environment.GetCommandLineArgs().
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -128,7 +128,7 @@ namespace Microsoft.CodeAnalysis.IL
         }
 
         public override int Run(AnalyzeOptions analyzeOptions)
-        {            
+        {
             if (analyzeOptions.TargetFileSpecifiers == null || !analyzeOptions.TargetFileSpecifiers.Any())
             {
                 throw new ArgumentNullException("Please specify one or more files, directories, or filter patterns for BinSkim analyze.", (Exception)null);
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -130,10 +130,9 @@ namespace Microsoft.CodeAnalysis.IL
         public override int Run(AnalyzeOptions analyzeOptions)
         {
             // Check whether target file is specified for BinSkim Analyze. If not, exit with code 1 (failure).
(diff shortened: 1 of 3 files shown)


src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -129,6 +129,12 @@ internal static string ReturnCommonPathRootFromTargetSpecifiersIfOneExists(IEnum
 
         public override int Run(AnalyzeOptions analyzeOptions)
         {
+            // Check whether target file is specified for BinSkim Analyze. If not, exit with code 1 (failure).
+            if (analyzeOptions.TargetFileSpecifiers == null || !analyzeOptions.TargetFileSpecifiers.Any())
+            {
+                throw new ArgumentNullException("Please specify a targe file after BinSkim analyze", (Exception)null);
```
shaopeng-gh, line 135: >targe [](http://example.com/codeflow?start=66&length=5)

1. spelling targe -> target
2. I think this does not need to be a file, please check what can be in `TargetFileSpecifiers`
3. If sentence, end with `.`
4. I think, change `after` to `for`

yongyan-gh:
You can update the title to a more readable title

shaopeng-gh:
fix format:
image

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -129,6 +129,11 @@ internal static string ReturnCommonPathRootFromTargetSpecifiersIfOneExists(IEnum
 
         public override int Run(AnalyzeOptions analyzeOptions)
         {
+            if (analyzeOptions.TargetFileSpecifiers == null || !analyzeOptions.TargetFileSpecifiers.Any())
+            {
+                throw new ArgumentNullException("Please specify one or more files, directories, or filter patterns for BinSkim analyze.", (Exception)null);
```
michaelcfanning, line 134: >null [](http://example.com/codeflow?start=149&length=4)

You don't need to pass this argument, null is the default value for inner exception when it isn't provided.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -129,6 +129,11 @@ internal static string ReturnCommonPathRootFromTargetSpecifiersIfOneExists(IEnum
 
         public override int Run(AnalyzeOptions analyzeOptions)
         {
+            if (analyzeOptions.TargetFileSpecifiers == null || !analyzeOptions.TargetFileSpecifiers.Any())
```
michaelcfanning, line 132: >analyzeOptions.TargetFileSpecifiers == null || !analyzeOptions.TargetFileSpecifiers.Any()) [](http://example.com/codeflow?start=16&length=90)

You can write this line of code more concisely as:

if (!analyzeOptions.TargetFileSpecifiers?.Any() == true)

shaopeng-gh: please merge as Michael's input.
Maye need to move `!` since can not use `!` on `null`
`if (analyzeOptions.TargetFileSpecifiers?.Any() != true)`

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
LingZhou-gh: If only passing one argument, it will be treated as the parameter name.
image
Then the result will be as follows:
image

michaelcfanning: Oh, right, this is one of those exception ctors that's inconsistent in .NET.

So do this:

throw new ArgumentNullException(nameof(analyzeOptions.TargetFileSpecifiers));

LingZhou-gh: Ok. 
Now the result is as follows:
image

michaelcfanning (approved):
:shipit:

shaopeng-gh (approved):
:shipit:

PR #764, opened 2022-12-14
Implicit const break.

michaelcfanning:
An implicit const attribute actually has its value data in the debug abbreviations table. This is different from nearly all other types, that read values from the debug data stream itself.

This failure to retrieve the additional value from the abbreviations data led to corrupt attribute specifications for everything following this value.

This condition is pretty easy to detect within code, we start getting unrecognized attribute/format values + combinations.

Looking further into the code/spec, it is clear we're not handling lots of other new DWARF5 things that we should. I did a bit of work/analysis and realized we need to separate some of this into a follow-on change (as it requires populating sections we don't populate today, a non-trivial amount of new code).

@shaopeng-gh @suvamM @LingZhou-gh @yongyan-gh


SetCurrentVersion.cmd
```diff
@@ -1,9 +1,9 @@
...
+set MAJOR_PREVIOUS=2
+set MINOR_PREVIOUS=0
+set PATCH_PREVIOUS=0
+set PRERELEASE_PREVIOUS=-rc1
 
 set MAJOR=2
 set MINOR=0
 set PATCH=0
-set PRERELEASE=-rc1
+set PRERELEASE=-rc2
```
michaelcfanning, line 9: >rc2 [](http://example.com/codeflow?start=16&length=3)

This change prepares us to release rc2.

src/BinaryParsers/ElfBinary/Dwarf/DwarfCompilationUnit.cs
```diff
@@ -263,7 +264,27 @@ private void ReadData(IDwarfBinary dwarfBinary, DwarfMemoryReader debugData, Dwa
                         case DwarfFormat.Strx:
                         case DwarfFormat.GNUStrIndex:
                             attributeValue.Type = DwarfAttributeValueType.String;
-                            attributeValue.Value = debugData.LEB128() + (ulong)beginPosition;
+                            debugData.LEB128();
+                            break;
+
+                        case DwarfFormat.Strx1:
```
michaelcfanning, line 270: >Strx1 [](http://example.com/codeflow?start=41&length=5)

These format types currently exist in some test binaries but we're not handling them properly. For now, I just seek forward into the debug data streams so that follow on attributes don't get corrupted. In a next change, we will retrieve the actual values of these strings.

src/BinaryParsers/ElfBinary/Dwarf/DwarfCompilationUnit.cs
```diff
@@ -263,7 +264,27 @@
...
+                        case DwarfFormat.Strx2:
+                            attributeValue.Type = DwarfAttributeValueType.String;
+                            debugData.ReadUshort();
+                            break;
+
+                        case DwarfFormat.Strx3:
+                            attributeValue.Type = DwarfAttributeValueType.String;
+                            debugStrings.ReadThreeBytes();
+                            break;
+
```
michaelcfanning, line 284: >Addrx [](http://example.com/codeflow?start=41&length=5)

We aren't handling several other address types we should....

src/BinaryParsers/ElfBinary/Dwarf/DwarfCompilationUnit.cs
```diff
@@ -190,12 +190,13 @@
                             break;
 
                         case DwarfFormat.Strp:
+                        case DwarfFormat.StrpSup:
```
shaopeng-gh, line 193: >case DwarfFormat.StrpSup: [](http://example.com/codeflow?start=24&length=25)

there are 2 `case DwarfFormat.StrpSup:`
another one about 5 lines below

src/BinaryParsers/ElfBinary/Dwarf/DwarfCompilationUnit.cs
```diff
@@ -515,6 +539,7 @@
                     {
                         DwarfAttribute attribute = (DwarfAttribute)debugDataDescription.LEB128();
                         DwarfFormat format = (DwarfFormat)debugDataDescription.LEB128();
+                        object value = null;
 
                         while (format == DwarfFormat.Indirect)
```
michaelcfanning, line 543: >ImplicitConst [](http://example.com/codeflow?start=50&length=13)

This is the code path that caused our break. The failure to retrieve this value and move the stream position pointer forward caused subsequent corruption in parsing.

suvamM (approved):

michaelcfanning:
@suvamM, we can cut an RC2. i am also looking into other parsing gaps, did we get a good binary drop to tests against from someone??

suvamM:
> @suvamM, we can cut an RC2. i am also looking into other parsing gaps, did we get a good binary drop to tests against from someone??

@michaelcfanning Nothing yet unfortunately. I am tracking several threads, however, and hopefully will have a good drop(s) soon.

PR #765, opened 2022-12-16
Debug string offsets.

michaelcfanning:
PR to propose parsing a more comprehensive set of DWARF5 symbol constructs.

In a few cases, such as retrieving strings from the lines strings section, we actually populate debug data attribute values where we previously did not.

In other cases, I've done sufficient work to properly move the stream reader pointer forward, to prevent follow-on problems parsing invalid values from the data stream. We don't actually resolve the attribute value for all these cases. These include values that need to be retrieved from supplemental objects. We also do not properly resolve all addresses.


src/BinaryParsers/ElfBinary/Dwarf/DwarfAttributeValue.cs
```diff
@@ -77,6 +77,10 @@ public class DwarfAttributeValue
         /// </summary>
         public object Value { get; set; }
 
+        // Gets or sets a property that temporarily stores an offset that 
+        // will beused to retrieve the actual attribute value.
+        public ulong? Offset { get; set; }
```
michaelcfanning, line 82: >Offset [](http://example.com/codeflow?start=22&length=6)

So there's a core design issue, that when parsing certain attribute specifications, there may be information elsewhere in the current set of specifications that are required to resolve an actual attribute value.

For example, a string attribute may include an offset in the literal attribute data stream. This offset can't be interpreted unless/until a different attribute spec provides a base location for that offset.

With the current code, what we do is to populate this 'Offset' property with the interim attribute spec data. We need to figure out later how best to handle post-processing to resolve all these offsets into actual values.

It's not clear that resolving all these is strictly required for our binary analysis. It would be for any comprehensive DWARD5 parsing component, of course.

src/BinaryParsers/ElfBinary/Dwarf/DwarfCompilationUnit.cs
```diff
@@ -29,10 +31,23 @@ public class DwarfCompilationUnit
...
         /// <param name="debugStrings">The debug strings.</param>
+        /// <param name="debugStrings">The debug string offsets.</param>
         /// <param name="addressNormalizer">Normalize address delegate (<see cref="NormalizeAddressDelegate"/>)</param>
-        public DwarfCompilationUnit(IDwarfBinary dwarfBinary, DwarfMemoryReader debugData, DwarfMemoryReader debugDataDescription, DwarfMemoryReader debugStrings, NormalizeAddressDelegate addressNormalizer)
+        public DwarfCompilationUnit(IDwarfBinary dwarfBinary,
+                                    DwarfMemoryReader debugData,
+                                    DwarfMemoryReader debugDataDescription,
+                                    DwarfMemoryReader debugStrings,
+                                    DwarfMemoryReader debugLineStrings,
+                                    IList<int> debugStringOffsets,
```
michaelcfanning, line 41: >debugStringOffsets [](http://example.com/codeflow?start=47&length=18)

So this implementation today processes and resolves all debug string offsets. That's because parsing/interpreting this data seems to be different than what the current design can be cleanly extended to include. I'll keep looking at it.

The way this works is that there's a single debug string offsets section. Each element in this section comprises a string location. The attribute values include an index into this locations array. Each compilation unit, however, has an index offset (which marks where in the overall section the compilation unit started contributing strings).

As a convenience, we rip through the section completely to make it easier to resolve things. The remainder of the code works in a sort of forward only capacity (often caching data that its already parsed).

src/BinaryParsers/MachOBinary/SingleMachOBinary.cs
```diff
@@ -16,11 +16,33 @@ public class SingleMachOBinary : BinaryBase, IDwarfBinary
         private const string SECTIONNAME_TEXT = "__text";
         private const string SECTIONNAME_DATA = "__data";
         private const string SECTIONNAME_EH_FRAME = "__eh_frame";
+        private const string SECTIONNAME_DEBUG_LOC = "__debug_loc";
```
michaelcfanning, line 19: >__debug_loc [](http://example.com/codeflow?start=54&length=11)

I've aggressively added a number of section names we don't currently parse in any way, simply so that others don't need to look for them in the future. These sections names don't match ELF section names precisely and docs seems incrementally harder to locate.

PR #766, opened 2022-12-16
Build BinSkim against .net 6.0

LingZhou-gh:
Since .NET Core 3.1 reached End of Support on December 13, 2022, BinSkim needs to build against .net6.0.

The changes so far only build **BinSkim.Driver.csproj** against .net6.0, as based on the following article, the referenced projects don't seem to need any changes.
[Example of migrating to the latest .NET | Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/architecture/modernize-desktop/example-migration)
image

The output folders will look like as follows:
image


src/BinSkim.Driver/BinSkim.Driver.csproj
```diff
@@ -68,6 +68,7 @@
   <ItemGroup>
     <!-- May need to be changed to x86 on a 32 bit machine. -->
     <MsDiaLibs Include="..\..\refs\x64\*" Condition="'$(TargetFramework)'=='$(NetCoreVersion)' And '$(RuntimeIdentifier)'==''" />
```
yongyan-gh, line 70: >$(TargetFramework) [](http://example.com/codeflow?start=54&length=18)

This macro doesn't exist anymore.
Do not need the additional <MsDiaLibs> since both `NetCoreVersion` and `NetSixVersion` need this tag. Just remove the TargetFramework condition.

yongyan-gh:
> The changes so far only build **BinSkim.Driver.csproj** against .net6.0, as based on the following article, the referenced projects don't seem to need any changes. [Example of migrating to the latest .NET | Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/architecture/modernize-desktop/example-migration) image
 
This is about migrating to new SDK project file format.

Since the dependent projects of Binskim.Driver are built on .net core 3.1 and .net standard 2.1 it should be fine, but prefer updating all projects to net 6.0

src/build.netcore.props
```diff
@@ -3,7 +3,8 @@
   <Import Project="$(MSBuildThisFileDirectory)build.common.props" />
 
   <PropertyGroup>
-    <NetCoreVersion>netcoreapp3.1</NetCoreVersion>
```
yongyan-gh, line 6: >netcoreapp3.1 [](http://example.com/codeflow?start=20&length=13)

do we want to keep .net core 3.1?
[https://devblogs.microsoft.com/dotnet/net-core-3-1-will-reach-end-of-support-on-december-13-2022/](https://devblogs.microsoft.com/dotnet/net-core-3-1-will-reach-end-of-support-on-december-13-2022/)

shaopeng-gh: I think we need to keep all existing and add 6.0

yongyan-gh:
You can update all test projects adding `net6.0` so dotnet will run tests using .net 6.0. Existing framework can be removed if we don't want to run tests for existing framework (.netcoreapp3.1)

src/BinSkim.Rules/RulesExtensionMethods.cs
```diff
@@ -35,7 +35,7 @@ public static string CreateOutputCoalescedByCompiler(this IList<ObjectModuleDeta
 
                 if (!libraries.TryGetValue(modulesKey, out SortedSet<string> modules))
                 {
-                    modules = libraries[modulesKey] = new SortedSet<string>();
+                    modules = libraries[modulesKey] = new SortedSet<string>(Comparer<string>.Create((a, b) => Path.GetFileNameWithoutExtension(a).CompareTo(Path.GetFileNameWithoutExtension(b))));
```
LingZhou-gh, line 38: Without this code change, under .net6.0 testing, the expected .sarif file and the actual .sarif file only differs in file **order**, as shown below:
image
It turns out that net6.0 changed the output of **String.Compare("dll_dllmain_stub.obj", "dll_dllmain.obj")**  compared with netcore3.1. The statement returns 1 in netcore3.1 while it returns -1 in net6.0. 

The code change only compares the filename **without extension**, so that the actual file can be consistent across freamworks.

src/BinSkim.Rules/RulesExtensionMethods.cs
```diff
@@ -92,7 +92,7 @@ public static string CreateOutputCoalescedByLibrary(this IList<ObjectModuleDetai
                 sb.Append(key);
 
                 List<string> objectModules = librariesToObjectModulesMap[key];
-                objectModules.Sort();
+                objectModules.Sort((a, b) => Path.GetFileNameWithoutExtension(a).CompareTo(Path.GetFileNameWithoutExtension(b)));
                 string[] oms = objectModules.ToArray();
```
LingZhou-gh, line 96: Same as above.

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/MixedMode_x64_VS2013_Default.dll.sarif
```diff
@@ -882,14 +882,14 @@
...
+                "text": "Application code should be compiled with the Spectre mitigations switch (/Qspectre cl.exe command-line argument or <SpectreMitigation>Spectre</SpectreMitigation> build property). Spectre attacks can compromise hardware-based isolation, allowing non-privileged users to retrieve potentially sensitive data from the CPU cache. To resolve this issue, provide the /Qspectre switch on the compiler command-line (or specify <SpectreMitigation>Spectre</SpectreMitigation> in build properties), or pass /d2guardspecload in cases where your compiler supports this switch and it is not possible to update to a toolset that supports /Qspectre. This warning should be addressed for code that operates on data that crosses a trust boundary and that can affect execution, such as parsing untrusted file inputs or processing query strings of a web request. You may need to install the 'C++ spectre-mitigated libs' component from the Visual Studio installer if you observe violations against C runtime libraries such as libcmt.lib, libvcruntime.lib, etc."
```
LingZhou-gh, line 885: All the file changes under **BaselineTestData/Expected** are generated by the tool: **UpdateBaselines.cmd.**
The change should go with commit (https://github.com/microsoft/binskim/commit/5f9aca8676f6c39aaf230f637b57fa414c0c3f27) . It won't cause test failure under netcore3.1 but will cause test failure in net6.0.

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
```diff
@@ -788,7 +788,7 @@ public static string GetExampleSarifPath(Sarif.SarifVersion sarifVersion)
 
         internal static string GetTestDirectory(string relativeDirectory)
         {
-            var codeBaseUrl = new Uri(Assembly.GetExecutingAssembly().CodeBase);
+            var codeBaseUrl = new Uri(Assembly.GetExecutingAssembly().Location);
             string codeBasePath = Uri.UnescapeDataString(codeBaseUrl.AbsolutePath);
```
LingZhou-gh, line 792: Assembly.CodeBase Property is obsolete, so change it to Assembly.Location.

src/ReleaseHistory.md
```diff
@@ -4,6 +4,7 @@
 * BUGFIX: Fix assertion failed with no clue when TargetFileSpecifiers is null or empty for BinSkim analyze.[763](https://github.com/microsoft/binskim/pull/763)
 * BUGFIX: fix `ERR997.ExceptionLoadingAnalysisTarget : Could not load analysis target` errors analyzing *nix binary resulting from failure to properly parse DWARF debug information.
 * Upgrade `Newtonsoft.JSON` package to 13.0.2 to resolve security alert.
+* Upgrade BinSkim to .net6.0 as .net core 3.1 reached end of support on 12/13/2022.
```
LingZhou-gh, line 7: v2.0.0-rc2 is released, will we have v2.0.0-rc3?

michaelcfanning: So, rc2 was released so you need to update the release history to reflect this and start the rc3 (Unreleased) section.

Just follow the pattern as you see for rc1

src/build.netcore.props
```diff
@@ -3,7 +3,8 @@
...
+    <NetSixVersion>net6.0</NetSixVersion>
+    <NetCoreVersion>netcoreapp3.1</NetCoreVersion>
```
michaelcfanning, line 7: >NetCoreVersion [](http://example.com/codeflow?start=5&length=14)

I don't think you need a new property, just set the following value to NetCoreVersion?

<NetCoreVersion>netcoreapp3.1;net6.0</NetCoreVersion>

src/BinSkim.Rules/RulesExtensionMethods.cs
michaelcfanning: Can you please comment this code?

src/BinSkim.Rules/RulesExtensionMethods.cs
michaelcfanning: Please comment this code, it's a non-obvious change. Did you find any online documentation (or PR) on this change?

src/Test.FunctionalTests.BinSkim.Driver/Test.FunctionalTests.BinSkim.Driver.csproj
```diff
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory).., build.netcore.props))\build.netcore.props" />
   <PropertyGroup>
-    <TargetFrameworks>$(NetCoreVersion);</TargetFrameworks>
+	<TargetFrameworks>$(NetCoreVersion);$(NetSixVersion)</TargetFrameworks>
```
michaelcfanning, line 5: >NetSixVersion [](http://example.com/codeflow?start=39&length=13)

Suggest just updating the value of NetCoreVersion rather than creating a new property.

src/Test.UnitTests.BinaryParsers/Elf/ElfBinaryTests.cs
```diff
@@ -20,7 +20,7 @@ public class ElfBinaryTests
 
         internal static string GetTestDirectory(string relativeDirectory)
         {
-            var codeBaseUrl = new Uri(Assembly.GetExecutingAssembly().CodeBase);
+            var codeBaseUrl = new Uri(Assembly.GetExecutingAssembly().Location);
```
michaelcfanning, line 23: >Location [](http://example.com/codeflow?start=70&length=8)

Explanation for change? Location looks right, but why did this work previously?

src/Test.UnitTests.BinSkim.Driver/CompilerDataLoggerTests.cs
michaelcfanning: OK, here's the info I asked for. :)

michaelcfanning (changes requested):
:clock1:

src/BinSkim.Rules/RulesExtensionMethods.cs
LingZhou-gh: Behavior changes when comparing strings in .NET 5 and later (official reprint) (iditect.com)
https://iditect.com/article/behavior-changes-when-comparing-strings-in-net-5-and-later-official-reprint.html

Add the following item in the project file to revert back using NLS.
image

src/BinSkim.Rules/RulesExtensionMethods.cs
LingZhou-gh: Behavior changes when comparing strings in .NET 5 and later (official reprint) (iditect.com)
https://iditect.com/article/behavior-changes-when-comparing-strings-in-net-5-and-later-official-reprint.html

Add the following item in the project file to revert back using NLS.
image

suvamM (approved):

BuildAndTest.cmd
```diff
@@ -59,6 +59,9 @@ echo Creating Platform Specific BinSkim 'Publish' Packages
 call :CreatePublishPackage netcoreapp3.1 win-x64 || goto :ExitFailed
```
michaelcfanning, line 59: >CreatePublishPackage [](http://example.com/codeflow?start=6&length=20)

Is this right? Why don't we have a single CreatePublishPackage step for each platform?

We don't create a netcoreapp3.1 specific package and also a net6.0 specific package, we should be creating a single package with both frameworks.

It's possible our script isn't set up to do this.

Let's get this change in and take a further look at my question.

michaelcfanning (approved):
:shipit:

PR #770, opened 2022-12-22
Lazily load DWARF data.

michaelcfanning:
Lazily load various DWARF things. This will push parsing exceptions into the various rules that first access (and therefore trigger) the parsing problem.

This also prevents over-eager loading of data that rules don't even inspect currently (as was happening with debug line information). This will reduce analysis time and memory pressure.

@shaopeng-gh, my apologies, I didn't see your PR yesterday. Can we review this and check it in and then you can reapply your debug symbols improvement (which looks necessary for complete lazy loading of that code path).

@suvamM


shaopeng-gh:
Can we review this and check it in and then you can reapply your debug symbols improvement (which looks necessary for complete lazy loading of that code path). ----- sounds good, I think part of my changes are also needed, will base on your change and look again.

PR #774, opened 2022-12-23
* BUGFIX: Fix over-aggressive parsing of DWARF compilation units even when all related rules are disabled.

shaopeng-gh:
When user disable all related rules, we are still reading DWARF compilation units,
when there is bug reading DWARF compilation units, there is no way user can get pass it by disable related rules.
This PR fix this issue.


src/BinaryParsers/ElfBinary/ElfBinary.cs
```diff
@@ -57,13 +57,13 @@ public ElfBinary(Uri uri, string localSymbolDirectories = null, bool forceCompre
                 SectionRegions = ELF.Sections.Where(s => s.LoadAddress > 0).OrderBy(s => s.LoadAddress).ToArray();
 
                 CompilationUnits = new Lazy<List<DwarfCompilationUnit>>(()
-                    => DwarfSymbolProvider.ParseAllCompilationUnits(this,
+                    => LoadDebug(DwarfSymbolProvider.ParseAllCompilationUnits(this,
```
michaelcfanning, line 60: >ParseAllCompilationUnits [](http://example.com/codeflow?start=53&length=24)

You have broken the formatting for these parameters, each of which should have its own line, each of which should be left-aligned with the first parameter.

src/BinaryParsers/ElfBinary/ElfBinary.cs
```diff
@@ -57,13 +57,13 @@ public ElfBinary(Uri uri, string localSymbolDirectories = null, bool forceCompre
...
                                                                     DebugData,
                                                                     DebugDataDescription,
                                                                     DebugDataStrings,
                                                                     DebugLineString,
                                                                     DebugStringOffsets,
-                                                                    NormalizeAddress));
+                                                                    NormalizeAddress), localSymbolDirectories));
```
michaelcfanning, line 66: >localSymbolDirectories [](http://example.com/codeflow?start=87&length=22)

this should be on its own line.

src/BinaryParsers/ElfBinary/ElfBinary.cs
```diff
@@ -57,13 +57,13 @@ public ElfBinary(Uri uri, string localSymbolDirectories = null, bool forceCompre
                 SectionRegions = ELF.Sections.Where(s => s.LoadAddress > 0).OrderBy(s => s.LoadAddress).ToArray();
 
                 CompilationUnits = new Lazy<List<DwarfCompilationUnit>>(()
-                    => DwarfSymbolProvider.ParseAllCompilationUnits(this,
+                    => LoadDebug(DwarfSymbolProvider.ParseAllCompilationUnits(this,
                                                                     DebugData,
```
michaelcfanning, line 61: >DebugData [](http://example.com/codeflow?start=68&length=9)

this argment name (and all others) should be left-aligned with 'this' (the first argument)

src/BinaryParsers/ElfBinary/ElfBinary.cs
```diff
@@ -448,21 +469,16 @@ private void LoadDebug(string localSymbolDirectories = null)
 
                         if (dwoBinary != null && dwoBinary.CompilationUnits.Value.Count > 0)
```
michaelcfanning, line 470: >dwoBinary.CompilationUnits.Value [](http://example.com/codeflow?start=49&length=32)

are we supposed to be referring to this CompilationUnits member here?

src/ReleaseHistory.md
```diff
@@ -2,6 +2,7 @@
 
 ## **v2.0.0-rc3** (Unreleased)
 * BUGFIX: Eliminate `BA3003.EnableStackProtector` false positives when the target is statically linked. [744](https://github.com/microsoft/binskim/pull/744)
+* BUGFIX: Fix over-aggressive parsing of DWARF compilation units even when all related rules are disabled. [774](https://github.com/microsoft/binskim/pull/774)
```
michaelcfanning, line 5: >aggressive [](http://example.com/codeflow?start=19&length=10)

I note that we don't actually have a test that ensures we don't get into this data. Should we create one?

A simple test that initializes an Elf binary and then ensures no values are populated may be sufficient.

michaelcfanning (changes requested):
:clock1:

michaelcfanning (approved):
:shipit:

PR #775, opened 2022-12-26
Dwarf5 line number programs.

michaelcfanning:
Our DWARF parser doesn't properly create line number programs for v5.

This PR is a start on adding that support. I noticed this when we were aggressively parsing debug line number information (which we no longer do). As a result, you have to explicitly trigger parsing this information using BinSkim configuration, as below (and pass this XML via the -c argument).

This PR as it stands doesn't comprehensively retrieve all possible string values, nor does it yet handle DWO splitting properly.
~~~
<?xml version="1.0" encoding="utf-8"?>
<Properties Type="PropertiesDictionary">
  <Properties Key="BinaryParsers.Options" Type="PropertiesDictionary">
    <!--Set this value to 'true' to aggressively fault in all binary data on scan target load. This is useful to flush out exceptions and other issues in various binary parsers.-->
    <Property Key="ComprehensiveBinaryParsing" Value="True" Type="System.Boolean" />
  </Properties>
</Properties>
~~~


src/BinaryParsers/ElfBinary/Dwarf/DwarfCommonInformationEntry.cs
```diff
@@ -173,9 +173,9 @@
                     AddressSize = defaultAddressSize;
                     SegmentSelectorSize = 0;
                 }
-                CodeAlignmentFactor = data.LEB128();
-                DataAlignmentFactor = data.LEB128();
-                ReturnAddressRegister = data.LEB128();
+                CodeAlignmentFactor = data.ULEB128();
```
michaelcfanning, line 176: >ULEB128 [](http://example.com/codeflow?start=43&length=7)

I changed this name as the DWARF spec itself is constantly referring to ULEB so I found this clearer.

src/BinaryParsers/ElfBinary/Dwarf/DwarfLineNumberProgram.cs
```diff
@@ -173,62 +178,212 @@
...
+                byte addressSize = debugLine.ReadByte();
+                byte segmentSelectorSize = debugLine.ReadByte();
+            }
+
+            int headerLength = debugLine.ReadOffset(is64bit);
             byte minimumInstructionLength = debugLine.ReadByte();
-            bool defaultIsStatement = debugLine.ReadByte() != 0;
+            byte maximumOperationsPerInstruction = debugLine.ReadByte();
+
+            bool defaultIsStatement = debugLine.ReadByte() == 1;
```
shaopeng-gh, line 224: these 2 line change we now `debugLine.ReadByte()` one more time than before, was it a bug before?

michaelcfanning: Nice catch! This data was introduced in V4, updated code.

shaopeng-gh (approved):
:shipit:

Issue #785, opened 2023-01-06
Telemetry enhancements


pharring:
Based on an email thread with @michaelcfanning and others, these are the new telemetry points we'd like:

- Binskim version
- Command line
- Operating system information
- XML configuration
- Available rules and their enablement state; include whether rules are not applicable because we're running on the wrong platform.
- Which rules are firing

pharring:
Closed via #798

PR #786, opened 2023-01-06
Telemetry enhancements

pharring:
Addresses #785

Move App Insights telemetry out of CompilerDataLogger into its own Telemetry class Add support for connection string via BinskimAppInsightsConnectionString environment variable. Put the session ID in the global context instead of individual events. Use a more compact session ID format.
Log OS and tool version.
Log rule summary data.

### DIFF
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -23,6 +23,15 @@ namespace Microsoft.CodeAnalysis.IL
             new string[] { ".dll", ".exe", ".sys" }
             );
 
+        /// <summary>
+        /// Construct a new <see cref="MultithreadedAnalyzeCommand"/> instance.
+        /// </summary>
+        /// <param name="telemetry">The Application Insights telemetry instance. May be null.</param>
+        public MultithreadedAnalyzeCommand(Sdk.Telemetry telemetry = null)
+        {
+            this.Telemetry = telemetry;
+        }
+
         public override IEnumerable<Assembly> DefaultPluginAssemblies
         {
             get => new Assembly[] { typeof(MarkImageAsNXCompatible).Assembly };
@@ -31,6 +40,11 @@ namespace Microsoft.CodeAnalysis.IL
 
         protected override BinaryAnalyzerContext CreateContext(AnalyzeOptions options, IAnalysisLogger logger, RuntimeConditions runtimeErrors, PropertiesDictionary policy = null, string filePath = null)
         {
+            if (logger is AggregatingLogger aggregatingLogger && this.Telemetry?.TelemetryClient != null)
+            {
+                aggregatingLogger.Loggers.Add(new RuleTelemetryLogger(this.Telemetry.TelemetryClient));
+            }
+
             BinaryAnalyzerContext binaryAnalyzerContext = base.CreateContext(options, logger, runtimeErrors, policy, filePath);
 
             // Update context object based on command-line parameters.
@@ -59,7 +73,7 @@ namespace Microsoft.CodeAnalysis.IL
             // Command-line provided policy is now initialized. Update context 
             // based on any possible configuration provided in this way.
 
-            context.CompilerDataLogger = new CompilerDataLogger(options.OutputFilePath, options.SarifOutputVersion, context, this.FileSystem);
+            context.CompilerDataLogger = new CompilerDataLogger(options.OutputFilePath, options.SarifOutputVersion, context, this.FileSystem, this.Telemetry);
 
             // If the user has hard-coded a non-deterministic file path root to elide from telemetry,
             // we will honor that. If it has not been specified, and if all file target specifiers
@@ -110,13 +124,13 @@ namespace Microsoft.CodeAnalysis.IL
                 }
             }
 
-            int smallestPathLegth = smallestPath.Length;
-            if (smallestPathLegth == 0)
+            int smallestPathLength = smallestPath.Length;
+            if (smallestPathLength == 0)
             {
                 return string.Empty;
             }
 
-            if (smallestPath[smallestPathLegth - 1] != '\\')
+            if (smallestPath[smallestPathLength - 1] != '\\')
             {
                 // In this case, 'smallestPath' is equal to 'c:\path1\partial-path' (this is an incomplete path).
                 // Once we execute, our 'smallestPath' will be transformed into 'c:\path1\'.
@@ -196,5 +210,7 @@ namespace Microsoft.CodeAnalysis.IL
         }
 
         internal Sarif.SarifVersion UnitTestOutputVersion { get; set; }
+
+        private Sdk.Telemetry Telemetry { get; }
     }
 }
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -47,7 +47,7 @@ namespace Microsoft.CodeAnalysis.IL
 
         private bool IsValidScanTarget(string file)
         {
-            Uri uri = new Uri(file);
+            var uri = new Uri(file);
 
             return PEBinary.CanLoadBinary(uri) ||
                    ElfBinary.CanLoadBinary(uri) ||
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -8,6 +8,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 
+using Microsoft.CodeAnalysis.BinaryParsers;
 using Microsoft.CodeAnalysis.IL.Rules;
 using Microsoft.CodeAnalysis.IL.Sdk;
 using Microsoft.CodeAnalysis.Sarif;
@@ -38,6 +39,21 @@ namespace Microsoft.CodeAnalysis.IL
             set => throw new InvalidOperationException();
         }
 
+        protected override bool ShouldComputeHashes(string file, BinaryAnalyzerContext context)
+        {
+            return base.ShouldComputeHashes(file, context)
+                && IsValidScanTarget(file);
+        }
+
+        private bool IsValidScanTarget(string file)
+        {
+            Uri uri = new Uri(file);
+
+            return PEBinary.CanLoadBinary(uri) ||
+                   ElfBinary.CanLoadBinary(uri) ||
+                   MachOBinary.CanLoadBinary(uri);
+        }
+
         protected override BinaryAnalyzerContext CreateContext(AnalyzeOptions options, IAnalysisLogger logger, RuntimeConditions runtimeErrors, PropertiesDictionary policy = null, string filePath = null)
         {
             if (logger is AggregatingLogger aggregatingLogger && this.Telemetry?.TelemetryClient != null)
@@ -53,7 +69,11 @@ namespace Microsoft.CodeAnalysis.IL
             binaryAnalyzerContext.IgnorePdbLoadError = options.IgnorePdbLoadError;
             binaryAnalyzerContext.LocalSymbolDirectories = options.LocalSymbolDirectories;
             binaryAnalyzerContext.TracePdbLoads = options.Traces.Contains(nameof(Traces.PdbLoad));
-            binaryAnalyzerContext.MaxFileSizeInKilobytes = options.MaxFileSizeInKilobytes > 1024 ? options.MaxFileSizeInKilobytes : int.MaxValue;
+
+            binaryAnalyzerContext.MaxFileSizeInKilobytes =
+                options.MaxFileSizeInKilobytes >= 0
+                ? options.MaxFileSizeInKilobytes
+                : BinaryAnalyzerContext.MaxFileSizeInKilobytesDefaultValue;
 
 #pragma warning disable CS0618 // Type or member is obsolete
             if (options.Verbose && ShouldWarnVerbose)
@@ -143,6 +163,13 @@ namespace Microsoft.CodeAnalysis.IL
 
         public override int Run(AnalyzeOptions analyzeOptions)
         {
+            Stopwatch stopwatch = null;
+
+            if (analyzeOptions.Traces.Where(s => s == "ScanTime").Any())
+            {
+                stopwatch = Stopwatch.StartNew();
+            }
+
             if (analyzeOptions.TargetFileSpecifiers?.Any() != true)
             {
                 throw new ArgumentNullException(nameof(analyzeOptions.TargetFileSpecifiers), "Please specify one or more files, directories, or filter patterns for BinSkim analyze.");
@@ -197,6 +224,13 @@ namespace Microsoft.CodeAnalysis.IL
             catch (Exception e)
             {
                 Console.WriteLine(e);
+            }
+            finally
+            {
+                if (stopwatch != null)
+                {
+                    Console.WriteLine($"Scan time: {stopwatch.Elapsed}");
+                }
             }
 
             // In BinSkim, no rule is ever applicable to every target type. For example,
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -58,7 +58,12 @@ namespace Microsoft.CodeAnalysis.IL
         {
             if (logger is AggregatingLogger aggregatingLogger && this.Telemetry?.TelemetryClient != null)
             {
-                aggregatingLogger.Loggers.Add(new RuleTelemetryLogger(this.Telemetry.TelemetryClient));
+                var ruleTelemetryLogger = new RuleTelemetryLogger(this.Telemetry.TelemetryClient);
+
+                // Analysis has already started, so let the new logger know.
+                ruleTelemetryLogger.AnalysisStarted();
+
+                aggregatingLogger.Loggers.Add(ruleTelemetryLogger);
             }
 
             BinaryAnalyzerContext binaryAnalyzerContext = base.CreateContext(options, logger, runtimeErrors, policy, filePath);
--- a/src/BinSkim.Driver/RuleTelemetryLogger.cs
+++ b/src/BinSkim.Driver/RuleTelemetryLogger.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 
 using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.CodeAnalysis.Sarif;
 
 namespace Microsoft.CodeAnalysis.IL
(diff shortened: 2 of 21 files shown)


src/BinSkim.Sdk/Telemetry.cs
```diff
@@ -0,0 +1,183 @@
...
+        public void Dispose()
+        {
+            TelemetryClient? telemetryClient = TelemetryClient;
+            if (telemetryClient is null)
+            {
+                return;
+            }
+
+            // Flush and close AppInsights client.
+            telemetryClient.Flush();
```
yongyan-gh, line 90: >telemetryClient.Flush(); [](http://example.com/codeflow?start=12&length=24)

Looks like there is an async version `[FlushAsync]([https://github.com/microsoft/ApplicationInsights-dotnet/pull/2202)`](https://github.com/microsoft/ApplicationInsights-dotnet/pull/2202)`) merged in April 2021. Not sure if its released

pharring: There is, but we wouldn't want it here. We want a **synchronous** flush (a method that blocks until all buffered events are transmitted). Otherwise, this would have to be an async method and IAsyncDisposable.

yongyan-gh: ok do you mean even [FlushAsync] may return before all telemetries are sent?

I saw below comment at this closed [issue](https://github.com/MicrosoftDocs/azure-docs/issues/98256)

>>>Any user who wants to avoid the Sleep should shut down their application using FlushAsync instead.

But the Azure Doc page still says sleep is required.

pharring: Because it simplifies the code (and "hides" the problem), I'll switch to FlushAsync.

src/BinSkim.Sdk/Telemetry.cs
```diff
@@ -0,0 +1,183 @@
...
+
+            // Flush is not always synchronous. See:
+            // https://github.com/microsoft/ApplicationInsights-dotnet/issues/407
+            // However, the default InMemoryChannel _does_ have a synchronous
+            // Flush and we can avoid an inconvenient delay on shutdown.
+            if (!(this.TelemetryConfiguration!.TelemetryChannel is InMemoryChannel))
+            {
+                Task.Delay(5000).Wait();
```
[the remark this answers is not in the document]
pharring: The latest version uses FlushAsync, so no sleep is necessary.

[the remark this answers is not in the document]
marmegh: If we don't need a sleep, why are we using `Task.Delay` here? Should this be removed?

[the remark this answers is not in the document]
pharring: @marmegh you commented on an outdated change. In the latest version, there is neither Sleep nor Delay.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -47,6 +56,11 @@ private bool IsValidScanTarget(string file)
 
         protected override BinaryAnalyzerContext CreateContext(AnalyzeOptions options, IAnalysisLogger logger, RuntimeConditions runtimeErrors, PropertiesDictionary policy = null, string filePath = null)
         {
+            if (logger is AggregatingLogger aggregatingLogger && this.Telemetry?.TelemetryClient != null)
```
michaelcfanning, line 59: >AggregatingLogger [](http://example.com/codeflow?start=26&length=17)

So, if the input logger is _not_ an aggregating logger, should we create one, add the original sole logger to it and then add the telemetry logger?

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -130,13 +144,13 @@ internal static string ReturnCommonPathRootFromTargetSpecifiersIfOneExists(IEnum
...
-            int smallestPathLegth = smallestPath.Length;
-            if (smallestPathLegth == 0)
+            int smallestPathLength = smallestPath.Length;
+            if (smallestPathLength == 0)
             {
                 return string.Empty;
             }
 
-            if (smallestPath[smallestPathLegth - 1] != '\\')
+            if (smallestPath[smallestPathLength - 1] != '\\')
```
michaelcfanning, line 153: >smallestPathLength [](http://example.com/codeflow?start=29&length=18)

You are the hero we've been looking for.

src/BinSkim.Driver/RuleTelemetryLogger.cs
```diff
@@ -0,0 +1,179 @@
...
+    /// </summary>
+    internal sealed class RuleTelemetryLogger : IAnalysisLogger
+    {
+        // Application Insights event names
+        internal const string RuleSummaryEventName = "RuleSummary";
+        internal const string AnalysisStoppedEventName = "AnalysisStopped";
+
+        // Application Insights property names
+        internal const string RuleIdPropertyName = "RuleId";
+        internal const string RuntimeConditionsPropertyName = "Flags";
```
michaelcfanning, line 25: >Flags [](http://example.com/codeflow?start=63&length=5)

Is this the right name here? Will keep an eye open for references. These runtime conditions specify various things we observed at runtime, like exceptions, whether an error was produced. Did you prefer 'Flags' over something like 'RuntimeConditions' from a general spirit of keeping terms more concise?

src/BinSkim.Driver/RuleTelemetryLogger.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// <summary>
+        /// Construct a new <see cref="RuleTelemetryLogger"/>.
+        /// </summary>
+        /// <param name="telemetryClient">The Application Insights telemetry client.</param>
+        public RuleTelemetryLogger(TelemetryClient telemetryClient)
+        {
+            this.telemetryClient = telemetryClient ?? throw new ArgumentNullException(nameof(telemetryClient));
+        }
+
+        public void AnalysisStarted()
```
michaelcfanning, line 54: >AnalysisStarted [](http://example.com/codeflow?start=20&length=15)

So, no start event? I can appreciate you have sketched out some (very useful) things here and we may want to defer building out additional data not immediately relevant to our immediate goal to measure occurrences for specific rules. Eventually, I think perhaps a timestamp at start and analysis stopped would be useful, to measure how long binskim takes to analyze in various contexts.

src/BinSkim.Driver/RuleTelemetryLogger.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+        }
+
+        public void AnalysisStopped(RuntimeConditions runtimeConditions)
+        {
+            EventTelemetry stopEvent = new EventTelemetry(AnalysisStoppedEventName);
```
michaelcfanning, line 60: >EventTelemetry [](http://example.com/codeflow?start=12&length=14)

generally we prefer to use 'var' when assigning object construction with clearly specified type. we're supposed to be configured to produce warnings in the IDE on this, might not be properly enabled/flowing to you (or maybe you're using vs code not vs).

src/BinSkim.Driver/RuleTelemetryLogger.cs
```diff
@@ -0,0 +1,179 @@
...
+                default:
+                    break;
+            }
+        }
+
+        public void LogConfigurationNotification(Sarif.Notification notification)
+        {
+        }
+
+        public void LogToolNotification(Sarif.Notification notification)
```
michaelcfanning, line 131: >LogToolNotification [](http://example.com/codeflow?start=20&length=19)

We don't need to block this first PR, but we will want to add some events on notifications we receive, which will include unhandled exceptions in the engine and individual rules. We will have a special interest in enabling AI telemetry to tell us if any specific rule crashed and was disabled. Doing this is blocked on a change we just made to the SARIF SDK (to add the associated rule descriptor to this method), so let's defer. Maybe I can drop this in and run past you for review, so I can continue to learn about what you've done. I think it'd provide a nice verification on behavior for your new rule (and others).

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -29,12 +27,6 @@ namespace Microsoft.CodeAnalysis.IL.Sdk
     /// </summary>
     public class CompilerDataLogger : IDisposable
     {
-        [ThreadStatic]
-        internal static TelemetryClient s_injectedTelemetryClient;
```
michaelcfanning, line 33: >s_injectedTelemetryClient [](http://example.com/codeflow?start=40&length=25)

Thanks for cleaning up the type cohesion/encapsulation problems in the class.

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -263,8 +228,7 @@ public void Write(BinaryAnalyzerContext context, CompilerData compilerData)
                 { "dialect", compilerData.Dialect },
                 { "moduleName", compilerData.ModuleName ?? string.Empty },
                 { "moduleLibrary", (compilerData.ModuleName == compilerData.ModuleLibrary ? string.Empty : compilerData.ModuleLibrary ?? string.Empty) },
-                { "sessionId", this.sessionId },
```
michaelcfanning, line 266: >sessionId [](http://example.com/codeflow?start=19&length=9)

It's good to prefer AI's actually handling for session id, thanks.

I have wondered, however, if we should have previously had an additional id that would help us sort telemetry within the data store.

For example. think about telemetry that we send to the working system that strictly comprises test information: should we have a constant guid/id here that identifies that data comain?

Or is there a more natural AI mechanism to keep data sets isolated? Any suggestions? We have pre-release test data, produced by dev tests, nightly functionals. We will have data produced by release candidates (that will execute in production automation systems) that could be useful to identify. eg.

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -387,7 +348,6 @@ internal void SendChunkedContent(string eventName, string contentId, string cont
 
                 this.telemetryClient.TrackEvent(eventName, properties: new Dictionary<string, string>
                 {
-                    { "sessionId", this.sessionId },
```
michaelcfanning, line 390: >sessionId [](http://example.com/codeflow?start=23&length=9)

Roughly related to previous comment, how should we be tracking version data in our telemetry? If BinSkim 1.0 has a bug in the source link rule, and we ship a fix in BinSkim 1.1, we'll want to assess behavior of the most current release (and also measure how many teams need to upgrade from 1.0 to the fixed version). How best to enable this?

src/BinSkim.Sdk/Telemetry.cs
```diff
@@ -0,0 +1,166 @@
...
+        {
+            // Flush and close AppInsights client.
+            this.TelemetryClient?.FlushAsync(CancellationToken.None).GetAwaiter().GetResult();
+            this.TelemetryConfiguration?.Dispose();
+        }
+
+        private static void ConfigureTelemetryContext(TelemetryContext context)
+        {
+            context.Session.Id = CreateRandomSessionId();
+            context.Component.Version = Assembly.GetCallingAssembly().GetCustomAttribute<AssemblyFileVersionAttribute>().Version;
```
michaelcfanning, line 90: >Version [](http://example.com/codeflow?start=30&length=7)

Oh, here's the version data. :)

src/BinSkim.Sdk/Telemetry.cs
```diff
@@ -0,0 +1,166 @@
...
+            context.Device.OperatingSystem = RuntimeInformation.OSDescription;
+        }
+
+        private static TelemetryConfiguration? CreateTelemetryConfigurationFromEnvironment()
+        {
+            string? appInsightsConnectionString = CompilerDataLogger.RetrieveEnvironmentVariable(AppInsightsConnectionStringEnvVar);
+            if (!string.IsNullOrEmpty(appInsightsConnectionString))
+            {
+                return new TelemetryConfiguration
```
michaelcfanning, line 99: >TelemetryConfiguration [](http://example.com/codeflow?start=27&length=22)

What happens if we pass a bad key here? Exception? Did you observe behavior for this case? What should our behavior be? On the one hand, silently failing to generate telemetry would be bad. On the other, anyone setting a bad key could bring down every binskim process that subsequently runs.

I think I'd favor failing immediately, ensuring sufficiently informative exception details on the console.

Thoughts?

pharring: I'll check and add a unit test, if necessary.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
pharring: Interesting. I tried something like that originally by creating a custom logger that wraps the passed-in logger, and passing the custom logger to base.CreateContext. It failed in CreateContext because there's an unconditional cast to AggregatingLogger. However, if I use another AggregatingLogger, that should work.

src/BinSkim.Driver/RuleTelemetryLogger.cs
pharring: Raises an interesting point. In App Insights, something with a duration is known as an Operation. There are two types of Operation: Requests and Dependencies. It would be natural to model the operation of "Analysis" as a "Request" with a start/stop. The nice thing about that, is that you emit just a single event when the operation ends with the duration. It'll be logged as a "RequestTelemetry" with a timestamp (the stopping time) and a duration from which you can compute the start time, if necessary.

src/BinSkim.Driver/RuleTelemetryLogger.cs
pharring: Yes, I was trying to be terse. Easy enough to change, of course.

src/BinSkim.Sdk/CompilerDataLogger.cs
pharring: You found where we set the version in the global context so it's attached to all telemetry items.

src/BinSkim.Sdk/CompilerDataLogger.cs
pharring: Probably "role name" is closest, but I'm not sure I understand what you mean by "data domain".

src/BinSkim.Driver/RuleTelemetryLogger.cs
pharring: Happy to help out. Yes, if "exception in rule x" goes through here, then it would be easy to log it. Although, I think the CompilerDataLogger already does something like that.

michaelcfanning:
>         // require that the scan is configured to produce a disk-based report.

Boo. This is what needs to be fixed and which is fixable by implementing a telemetry collecting logger, precisely the pattern you created.

---
Refers to: src/BinSkim.Sdk/CompilerDataLogger.cs:73 in 1a7ac0e. [](commit_id = 1a7ac0e0969c51c842f9571f8aa88df06f62ac79, deletion_comment = False)

src/BinSkim.Driver/RuleTelemetryLogger.cs
michaelcfanning: Right, of course! The compiler data logger has a very undesirable behavior: on analysis completion it actually reparses the output file to collect metrics for a summary.  Not good. All of that logic should be replaced with precisely the pattern you've started here: a logger should collect relevant metrics during analysis and then blat them out in the telemetry channel on analysis completion.

We definitely need to get an updated to the SDK that I just checked in to finish cleaning this up. If you're still investing in this PR, feel free to merge from main and get that breaking change (adds a reporting descriptor for the associated rule to the Log(Notification) method. I also think it'd be just fine to polish this up very slightly, merge, and then take the next incremental change.

src/BinSkim.Sdk/CompilerDataLogger.cs
michaelcfanning: RIght, not super clear. What I mean is do we need a way of organizing data in the general AI telemetry channel in logical sets.  Perhaps by allocating a specific GUID that groups all telemetry that's ever collected which publishes the guid. So, we have one GUID that all of our testing/pre-release versions send with telemetry. We use another GUID for all official releases. All telemetry is persisted to the same AI sink using the same AI connection strings (because we can't easily reconfigure these when running pre-release vs. production, say in OneBranch or some other ADO pipeline).

src/BinSkim.Driver/RuleTelemetryLogger.cs
michaelcfanning: Nice! It makes sense that the AI interaction itself can generate good timestamps for kick-off; there's no need for explicit activity here in the `AnalysisStarted` callback...

michaelcfanning:
I'm very comfortable accepting this change once you've completed the items you've marked 'pending'. Let's get that in. I will then update to the new IAnalyzeContext contract. We can then eliminate the log file parsing in the EO scenario and also ensure we're collecting rule exception data.

michaelcfanning (approved):
:shipit:

src/BinSkim.Sdk/CompilerDataLogger.cs
pharring: OK. RoleName is the closest to that. The AnalysisSummaryExtractor (called from CompilerDataLogger) already queries a few interesting environment variables such as the OrganizationName, ProjectName and RepositoryId. Ideal for when running in an AzDO pipeline. We could compose a RoleName from one or more of those environment variables.

src/BinSkim.Sdk/Telemetry.cs
pharring: There is validation in the ConnectionString setter. It throws an ArgumentException if it can't be parsed as "key=value" pairs with semicolon separators and also if InstrumentationKey is missing. I'm happy with that. It'll fail immediately with a meaningful exception.

src/BinSkim.Driver/RuleTelemetryLogger.cs
pharring: In the latest iteration, I'm using a RequestTelemetry with a name of "Analysis"

src/BinSkim.Driver/RuleTelemetryLogger.cs
pharring: OK. I implemented that in the latest iteration (11).
One little wrinkle: I had to call AnalysisStarted manually because it's called before we get a chance to add this logger. Might want to look at deferring the AnalysisStarted call in MultithreadedAnalyzeCommandBase.Analyze until after CreateContext. Or have another callback mechanism to let derived classes customize the logger.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
pharring: I looked into this and decided to leave it the way it is for now. The problem with handling the "not AggregatingLogger" case is that this method (CreateContext) gets called again for each file with a CachedLogger and, at least for now, we don't want to add a RuleTelemetryLogger to those.

I think CreateContext is the wrong place to add this code, but it's the only customization point we have at the moment.

I think we should have a virtual "CreateLogger" or "ConfigureLogger" method that is called just once from the base class -- at the point where we currently create the AggregatingLogger. But that requires a Sarif SDK change.

pharring:
@michaelcfanning, @marmegh, just as a reminder, I don't have permission to merge this myself. I think I've addressed all the comments so far. If you're happy with it, please merge for me.

pharring:
Policies won't allow merging this from my private fork. Recreated this PR in #798 from the Microsoft fork

PR #790, opened 2023-01-11
A working prototype for BinSkim dotnet tool

LingZhou-gh:
This is a working prototype for BinSkim dotnet tool. Create this PR for future use.

It doesn't use the traditional way as documented in https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools-how-to-create, as BinSkim need to reference native .dll and etc. 
This solution uses .nuspec and builds a publish folder especially for dotnet tool and then copy it into the inner folder of the package and then pack it into a dotnet tool.

The name has to be different from the normal NuGet package we already built, otherwise error will occur during installation.

Install the BinSkim dotnet tool according to https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools-how-to-use.

### DIFF
--- a/BuildAndTest.cmd
+++ b/BuildAndTest.cmd
@@ -63,6 +63,11 @@ call :CreatePublishPackage net6.0 win-x64 || goto :ExitFailed
 call :CreatePublishPackage net6.0 linux-x64 || goto :ExitFailed
 call :CreatePublishPackage net6.0 osx-x64 || goto :ExitFailed
 
+::Create the BinSkim dotnet tool publish packages
+echo Creating BinSkim dotnet tool publish packages
+dotnet publish %~dp0src\BinSkim.Driver\BinSkim.Driver.csproj --no-restore -c %Configuration% -f net6.0 -p:PublishDir=%~dp0bld\bin\%Platform%_%Configuration%\DotNetToolPublish\net6.0\any
+dotnet publish %~dp0src\BinSkim.Driver\BinSkim.Driver.csproj --no-restore -c %Configuration% -f netcoreapp3.1 -p:PublishDir=%~dp0bld\bin\%Platform%_%Configuration%\DotNetToolPublish\netcoreapp3.1\any
+
 ::Build NuGet package
 echo BuildPackages.cmd
 call BuildPackages.cmd || goto :ExitFailed
--- a/src/BinSkim.Driver/BinSkim.Driver.csproj
+++ b/src/BinSkim.Driver/BinSkim.Driver.csproj
@@ -20,7 +20,7 @@
     <RepositoryUrl>https://github.com/Microsoft/binskim</RepositoryUrl>
   </PropertyGroup>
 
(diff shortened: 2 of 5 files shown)


src/BinSkim.Driver/BinSkim.Driver.csproj
```diff
@@ -74,7 +74,7 @@
     <!-- Required DLLs for Microsoft Visual C++ runtime. -->
     <MsRuntime Include="..\..\refs\runtime\*" />
   </ItemGroup>
-
+
```
yongyan-gh, line 77: nit: remove this tab

src/BinSkim.Driver/BinSkim.Driver.csproj
```diff
@@ -20,7 +20,7 @@
     <RepositoryUrl>https://github.com/Microsoft/binskim</RepositoryUrl>
   </PropertyGroup>
 
-  <ItemGroup>
+  <ItemGroup>
```
yongyan-gh, line 23: >     [](http://example.com/codeflow?start=13&length=4)

nit: remove extra whitepsaces

BuildAndTest.cmd
```diff
@@ -63,6 +63,11 @@ call :CreatePublishPackage net6.0 win-x64 || goto :ExitFailed
 call :CreatePublishPackage net6.0 linux-x64 || goto :ExitFailed
 call :CreatePublishPackage net6.0 osx-x64 || goto :ExitFailed
 
+::Create the BinSkim dotnet tool publish packages
+echo Creating BinSkim dotnet tool publish packages
+dotnet publish %~dp0src\BinSkim.Driver\BinSkim.Driver.csproj --no-restore -c %Configuration% -f net6.0 -p:PublishDir=%~dp0bld\bin\%Platform%_%Configuration%\DotNetToolPublish\net6.0\any
```
yongyan-gh, line 68: >dotnet [](http://example.com/codeflow?start=0&length=6)

this can be done in `:CreatePublishPackage` by adding `PublishDir` as a new parameter

LingZhou-gh: The reason why :CreatePublishPackage is not used is that there is no runtimeID (win-64/linux-64/osx-64) to provide. The files are published into %~dp0bld\bin\%Platform%_%Configuration%\DotNetToolPublish\net6.0\**any** folder

src/Nuget/BinSkimDotNetTool.nuspec
```diff
@@ -0,0 +1,23 @@
+﻿<?xml version="1.0"?>
+<package >
+  <metadata>
+    <id>Microsoft.CodeAnalysis.BinSkim.DotNetTool</id>
+    <version>1.0.0</version>
```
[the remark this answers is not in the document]
LingZhou-gh: Tried and haven't made it work for some reason.  Maybe nuget.exe and dotnet pack has different format for input. Will keep trying.

PR #792, opened 2023-01-12
Fix for BuildAndTest.sh for public BinSkim

LingZhou-gh:
This is for the public BinSkim, where the existing Linux script does not check out the submodules.

PS. There are already similar scripts in BuildAndTest.cmd to check out submodules. The logic is same.

### DIFF
--- a/BuildAndTest.sh
+++ b/BuildAndTest.sh
@@ -6,6 +6,11 @@ if [[  "$(uname)" == "Linux" || "$(uname)" == "Darwin" ]]; then
   sed 's#\\#/#g' src/BinSkim.sln > src/BinSkimUnix.sln
 fi
 
+if [ ! -f src/sarif-sdk/src/Sarif.Sdk.sln ]; then
+  echo "Get submodule..."
+  git submodule update --init --recursive
+fi
+
 dotnet build src/BinSkimUnix.sln --configuration Release /p:Platform="x64"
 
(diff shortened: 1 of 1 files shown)


BuildAndTest.sh
```diff
@@ -6,6 +6,11 @@ if [[  "$(uname)" == "Linux" || "$(uname)" == "Darwin" ]]; then
   sed 's#\\#/#g' src/BinSkim.sln > src/BinSkimUnix.sln
 fi
 
+if [ ! -f src/sarif-sdk/src/Sarif.Sdk.sln ]; then
```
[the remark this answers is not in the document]
LingZhou-gh: The "!" in the condition did the trick.

BuildAndTest.sh
```diff
@@ -6,6 +6,11 @@ if [[  "$(uname)" == "Linux" || "$(uname)" == "Darwin" ]]; then
...
+  echo "Get submodule..."
+  git submodule update --init --recursive
+fi
+
```
shaopeng-gh, line 13: looks good.
just throwing a open question if this new added code need to be inside of the 
`if [[  "$(uname)" == "Linux" || "$(uname)" == "Darwin" ]]; then .... fi`
code block or outside.

shaopeng-gh (approved):
:shipit:

BuildAndTest.sh
suvamM: @LingZhou-gh Just to check, please test the changes on a fresh enlistment from this branch.

PR #794, opened 2023-01-14
Update SARIF SDK to 235394a8537f462dfe86f456d6b9e87cc8e22e9a

michaelcfanning:


michaelcfanning:
@pharring, this change is for you, I've merged the SDK change that provides an `associatedRule` reporting descriptor to the `IAnalsisLogger.Log(Notification)` helper that can document exceptions in your (and all other) rules.

PR #798, opened 2023-01-20
Telemetry enhancements

pharring:
Addresses #785

Move App Insights telemetry out of CompilerDataLogger into its own Telemetry class Add support for connection string via BinskimAppInsightsConnectionString environment variable. Put the session ID in the global context instead of individual events. Use a more compact session ID format.
Log OS and tool version.
Log rule summary data.


michaelcfanning (approved):
:shipit:

PR #805, opened 2023-02-03
Create reference point for v4.0.0 rc3

shaopeng-gh:
We are not planning to release v4.0.0 rc3
but still create reference point in case needed.


src/ReleaseHistory.md
```diff
@@ -13,7 +13,9 @@
 - UER => eliminate unhandled exceptions in rules
 - UEE => eliminate unhandled exceptions in engine
 
-## **v4.0.0-rc3** UNRELEASED
+## **v4.0.0-rc4** UNRELEASED
```
suvamM, line 16: @shaopeng-gh Why add `rc4` here? We should just tag `rc3` as `Unreleased`.

suvamM (changes requested):

src/ReleaseHistory.md
shaopeng-gh: people will be adding to the last one, since rc3 is freezed, adding a new one rc4 above to avoid that.

suvamM (approved):

PR #821, opened 2023-02-28
* BUG: Fix `PdbLoad` not working for `--trace`.

shaopeng-gh:
This is a Hotfix without the need to change sdk code.
The issue is currently `PdbLoad` is broken with exception:
image

In sdk it is trying to convert `PdbLoad` into `DefaultTraces` enum and it will throw exception because that enum will not have that value, that value is for Binskim only.

The fix is to "hide" the property and so that in sdk it will get an empty list so no error will be throw.
The way we hide the property, in Binskim we will still be able to get the values passed from the command line so this line will still work:
`binaryAnalyzerContext.TracePdbLoads = options.Traces.Contains(nameof(Traces.PdbLoad));`
in path:
`\src\BinSkim.Driver\MultithreadedAnalyzeCommand.cs`

### DIFF
--- a/src/BinSkim.Driver/AnalyzeOptions.cs
+++ b/src/BinSkim.Driver/AnalyzeOptions.cs
@@ -16,11 +16,11 @@ namespace Microsoft.CodeAnalysis.IL
         [Option(
             "trace",
             Separator = ';',
-            Default = null,
+            Default = new string[] { },
             HelpText = "Execution traces, expressed as a semicolon-delimited list, that " +
                        "should be emitted to the console and log file (if appropriate). " +
                        "Valid values: PdbLoad.")]
-        public override IEnumerable<string> Traces { get; set; }
+        public new IEnumerable<string> Traces { get; set; } = Array.Empty<string>();
 
         [Option(
             "sympath",
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -13,7 +13,8 @@
 - UER => eliminate unhandled exceptions in rules
 - UEE => eliminate unhandled exceptions in engine
 
-## **v4.0.0-rc5** UNRELEASED
+## **v4.0.0** UNRELEASED
+* BUG: Fix `PdbLoad` not working for `--trace`. [821](https://github.com/microsoft/binskim/pull/821)
 
 ## **v4.0.0-rc4**
 * DEP: Update Sarif.Sdk submodule from [235394a to 120fae35](https://github.com/microsoft/sarif-sdk/compare/235394a...120fae35). Full [SARIF SDK Release History](https://github.com/microsoft/sarif-sdk/blob/120fae35/src/ReleaseHistory.md).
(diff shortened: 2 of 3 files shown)


src/BinSkim.Driver/AnalyzeOptions.cs
```diff
@@ -16,11 +16,11 @@ public class AnalyzeOptions : AnalyzeOptionsBase
         [Option(
             "trace",
             Separator = ';',
-            Default = null,
+            Default = new string[] { },
```
shaopeng-gh, line 19: >new string[] { } [](http://example.com/codeflow?start=22&length=16)

fixed btw

src/BinSkim.Driver/AnalyzeOptions.cs
```diff
@@ -16,11 +16,11 @@ public class AnalyzeOptions : AnalyzeOptionsBase
...
             HelpText = "Execution traces, expressed as a semicolon-delimited list, that " +
                        "should be emitted to the console and log file (if appropriate). " +
                        "Valid values: PdbLoad.")]
-        public override IEnumerable<string> Traces { get; set; }
+        public new IEnumerable<string> Traces { get; set; } = Array.Empty<string>();
```
shaopeng-gh, line 23: > = Array.Empty<string>(); [](http://example.com/codeflow?start=59&length=25)

This is also needed for test cases.

shaopeng-gh, line 23: >new [](http://example.com/codeflow?start=15&length=3)

this is the only main change needed, everything works.

src/Test.FunctionalTests.BinSkim.Driver/AnalyzeCommandTests.cs
```diff
@@ -73,6 +73,25 @@ public void AnalyzeCommand_ShouldThrowWithVersionOne()
             Assert.Throws<InvalidOperationException>(() => command.Run(options));
         }
 
+        [Fact]
+        public void AnalyzeCommand_ShouldNotThrowWithPdbLoadTrace()
```
shaopeng-gh, line 77: before fix this test will throw.

michaelcfanning: well done. :)

michaelcfanning (approved):
:shipit:

src/ReleaseHistory.md
```diff
@@ -13,7 +13,8 @@
 - UER => eliminate unhandled exceptions in rules
 - UEE => eliminate unhandled exceptions in engine
 
-## **v4.0.0-rc5** UNRELEASED
+## **v4.0.0** UNRELEASED
+* BUG: Fix `PdbLoad` not working for `--trace`. [821](https://github.com/microsoft/binskim/pull/821)
```
michaelcfanning, line 17: >BUG [](http://example.com/codeflow?start=2&length=3)

Remember in the release notes to put yourself in the perspective of the customer. What evidence of this problem would they have seen? They will look for that evidence here. 

For exceptions, the typical thing to search on is the type of the exception and the top of the stack where it occurs. Notice how much more useful information is encoded in the line below. Please do think about this particular code review note, it is a comment that I have to frequently make. 

Great fix! And a nice job with this change overall, thanks very much

Fix unhandled `ArgumentException` in `Enum.TryParse` on passing `PdbLoad` value to `--trace` command-line argument.

shaopeng-gh: yes, fixed.

PR #828, opened 2023-03-04
Fix pdb trace bugs

shaopeng-gh:
1.	We have the trace object attached to the pdb object which is not right.
in each location finding the pdb will be a new object and we lost every trace before the last try.
I have moved the trace object to under the PEBinary so all try and be retained in the object model we can do as we need. (Logging for now)

2.	The logging is missing at a few places, ideally it should show every place it tried and the result of each try, by order.
I have added.

3.	I am able to create this error, it is nothing about computer have no memory, the code executed in a blink without any memory usage spike. 
This can be reproduced by using a .txt file and renamed to .pdb.
This is the behavior of the library. and there is a value for it `DiaHresult.E_OUTOFMEMORY`
 
image

4.	This code block should not be in the else block, it should always be run, even load side-by-side failed. 
Lost opportunity to load from Symbol Dir.
I have moved it outside.

image


src/BinaryParsers/PEBinary/PEBinary.cs
```diff
@@ -45,6 +46,12 @@ public class PEBinary : BinaryBase
             this.symbolPath = symbolPath;
             this.Valid = this.PE.IsPEFile;
             this.tracePdbLoad = tracePdbLoad;
+
+            if (this.tracePdbLoad)
```
shaopeng-gh, line 50: >if (this.tracePdbLoad) [](http://example.com/codeflow?start=12&length=22)

Please first see the description of the PR in browser, 
I notice screenshot does not show in CodeFlow, thanks.

src/Test.UnitTests.BinaryParsers/PEBinary/PEBinaryTests.cs
```diff
@@ -137,6 +139,88 @@ public void PEBinary_TryLoadPdbFromSymbolFolder()
...
+
+        private static void PdbLoadTraceTest_Helper(
+            string folderForPE,
+            string folderForSymbol,
+            bool tracePdbLoad,
+            bool expectPdbLoadSuccess,
+            int expectPdbLoadTraceLineCount
+            )
+        {
+            PEBinary.ClearLocalSymbolDirectoriesCache();
```
shaopeng-gh, line 185: >PEBinary.ClearLocalSymbolDirectoriesCache(); [](http://example.com/codeflow?start=12&length=44)

this is needed because the cache is static.

src/BinaryParsers/PEBinary/PEBinary.cs
```diff
@@ -121,25 +141,64 @@ private Pdb LoadPdb()
...
                 peOrPdbPath = peOrPdbPath.Replace(extension, pdbExtension, StringComparison.OrdinalIgnoreCase);
 
                 // If Pdb exists side-by-side with exe, let's try to read
                 if (File.Exists(peOrPdbPath))
                 {
                     TryLoadPdb(peOrPdbPath, pdbExtension, this.symbolPath, this.localSymbolDirectories, this.tracePdbLoad, out pdb);
+
+                    if (pdb == null && !string.IsNullOrWhiteSpace(this.PdbParseException?.LoadTrace))
+                    {
+                        this.PdbLoadTrace?.Append(this.PdbParseException.LoadTrace);
```
michaelcfanning, line 158: create a little helper to remove duplicated code. DRY it out.

src/BinaryParsers/PEBinary/ProgramDatabase/Pdb.cs
```diff
@@ -512,9 +506,40 @@ private void WindowsNativeLoadPdbUsingDia(string pdbPath)
...
+                };
+            }
+            catch (COMException ce)
+            {
+                this.loadTrace?.AppendLine($"  Examined PDB path: '{pdbPath}'. HResult: {ce.HResult}.");
+                if (!string.IsNullOrEmpty(ce.Message) && ce.Message.StartsWith("Error HRESULT E_FAIL"))
+                {
+                    throw new PdbException(DiaHresult.E_FAIL, ce)
+                    {
+                        LoadTrace = this.loadTrace?.ToString()
```
michaelcfanning, line 529: >this.loadTrace?.ToString() [](http://example.com/codeflow?start=36&length=26)

$"{this.loadTrace}"  == 
 this.loadTrace?.ToString()

src/ReleaseHistory.md
```diff
@@ -14,6 +14,7 @@
 - UEE => eliminate unhandled exceptions in engine
 
 ## **v4.1.0** UNRELEASED
+* BUG: Fix PDB probing details incompleteness on passing `PdbLoad` value to `--trace` command-line argument. [828](https://github.com/microsoft/binskim/pull/828)
```
michaelcfanning, line 17: >BUG: Fix PDB probing details incompleteness on passing `Pdb [](http://example.com/codeflow?start=2&length=59)

* BUG: Probe local symbols directory for PDBs in all code paths. [828](https://github.com/microsoft/binskim/pull/828)
* BUG: Add missing output in PDB load tracing (enabled by `--trace PdbLoad`. [828](https://github.com/microsoft/binskim/pull/828)

michaelcfanning (approved):
:shipit:

PR #853, opened 2023-04-11
Sarif sdk pipeline update

michaelcfanning:
This PR updates to latest SARIF SDK with some significant implications.

- Our threading model has changed. We no longer provide caching by file hash, that feature is disabled. This has implications for binskim (that feature was actually developed because BinSkim benefits from per-file caching for builds that make many copies of files, as many builds do). We need to clearly document this and also run comparison tests (in time-to-analyze) for some at-scale builds.
- The BinSkim command-line is significantly updated. We've dropped obsolete args like `--verbose`, `--hashes`, `--pretty-print`. `--force` & `pretty-print` are deprecated in favor of `--log ForceOverwrite;PrettyPrint` (this example combines the two in a single arg, which is the benefit of the change, to reduce cmd-line arg surface area). `--insert Hashes` replaces `--hashes`. We need to update the tool docs. This will break Guardian.
- Most churn in change comes from conversion of context `TargetUri` property to `CurrentTarget` (which has a `Uri` property)
- The PR itself is highly annotated with other observations and open things to chase down. I believe this change can be accepted into v4.1.0, our release branch, to unblock the Coyote testing. I will continue to chase down loose threads from the PR and pursue some SDK fixes as well.


src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -243,7 +226,7 @@ public override int Run(AnalyzeOptions analyzeOptions)
             // Because of this, the return code bit for RuleNotApplicableToTarget is not
             // interesting (it will always be set).
 
-            return analyzeOptions.RichReturnCode
+            return analyzeOptions.RichReturnCode != null && analyzeOptions.RichReturnCode.Value
```
michaelcfanning, line 229: >RichReturnCode [](http://example.com/codeflow?start=34&length=14)

Many SARIF SDK options are now nullable to allow them to be specified in a config file and then overridden on the command-line. This corner-case execution scenario will also require us to make a similar change to BinSkim. If we have time in 4.1 I'll make those updates in a follow-on PR. This PR is going in to enable the multithread analysis testing we need.

src/BinSkim.Driver/RuleTelemetryLogger.cs
```diff
@@ -48,13 +48,19 @@ internal sealed class RuleTelemetryLogger : IAnalysisLogger
...
+        }
+
         /// <summary>
         /// Construct a new <see cref="RuleTelemetryLogger"/>.
         /// </summary>
         /// <param name="telemetryClient">The Application Insights telemetry client.</param>
         public RuleTelemetryLogger(TelemetryClient telemetryClient)
         {
             this.telemetryClient = telemetryClient ?? throw new ArgumentNullException(nameof(telemetryClient));
+            this.FileRegionsCache = new FileRegionsCache();
```
michaelcfanning, line 63: >FileRegionsCache [](http://example.com/codeflow?start=17&length=16)

this is required to allow the #nullable feature in this file. I will clean this up soon.

src/BinSkim.Rules/DwarfRules/BA3003.EnableStackProtector.cs
```diff
@@ -171,8 +171,8 @@
                         context.Logger.Log(this,
                             RuleUtilities.BuildResult(FailureLevel.Error, context, null,
                                 nameof(RuleResources.BA3003_Error),
-                                context.TargetUri.GetFileName(),
-                                DwarfUtility.GetDistinctNames(failedList, context.TargetUri.GetFileName())));
+                                context.CurrentTarget.Uri.GetFileName(),
```
michaelcfanning, line 174: >CurrentTarget [](http://example.com/codeflow?start=40&length=13)

This is the most intrusive change. The `TargetUri` property is now an enumerated artifact, i.e., a URI + contents (if loaded).

src/BinSkim.Sdk/BinaryAnalyzerContext.cs
```diff
@@ -34,12 +33,6 @@ public IBinary Binary
             set => this.iBinary = value;
         }
 
-        public override Exception TargetLoadException
```
michaelcfanning, line 37: >TargetLoadException [](http://example.com/codeflow?start=34&length=19)

We now have runtime exceptions hanging off the context object. I think this change breaks target load exceptions! But no unit tests fail. I will go confirm this point.

src/BinSkim.Sdk/BinaryAnalyzerContext.cs
```diff
@@ -86,8 +66,6 @@ public override string MimeType
 
         public override bool AnalysisComplete { get; set; }
 
-        public override DefaultTraces Traces { get; set; }
```
michaelcfanning, line 89: >Traces [](http://example.com/codeflow?start=38&length=6)

This property and ForceOverwrite have been pushed lower into the stack, they are standard on the SDK base context classes.

src/Test.FunctionalTests.BinSkim.Driver/CommandLineTests.cs
```diff
@@ -18,37 +18,26 @@ namespace Microsoft.CodeAnalysis.BinSkim.Driver
...
-                this.InputBinSkimSarifVersion = inputBinSkimSarifVersion;
-                this.ArgsString = inputBinSkimSarifVersion == null ? argsStringBase : argsStringBase + inputBinSkimSarifVersion;
-                this.ExpectedSarifVersion = expectedSarifVersion;
+                this.ArgsString = argsStringBase;
                 this.ExpectedErrorParameter = expectedErrorParameter;
             }
         }
 
         [Fact]
-        public void CorrectlyParseGenerateJsonIntegerAs()
```
michaelcfanning, line 40: >CorrectlyParseGenerateJsonIntegerAs [](http://example.com/codeflow?start=20&length=35)

Huh? This test didn't test json integer at all. We no longer permit SARIF v1 to be generated so nearly everything useful is gone from these tests. Leaving the code as a stub to build out in a more useful way later.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -208,18 +203,6 @@ public override int Run(AnalyzeOptions analyzeOptions)
                 analyzeOptions.Level = new List<FailureLevel> { FailureLevel.Error, FailureLevel.Warning, FailureLevel.Note };
             }
```
schlaman-ms, line 204: Doesn't this go away as well?
Your notes say you have dropped obsolete args like --verbose.

src/ReleaseHistory.md
```diff
@@ -1,231 +0,0 @@
...
-- FPC => regex candidate reduction
-- FNC => regex candidate increase
-- FPS => FP reduction in static analysis
-- FNS => false negative reduction in static analysis
-- FPD => FP reduction in dynamic phase
-- FND => False negative reduction in dynamic phase
-- UER => eliminate unhandled exceptions in rules
-- UEE => eliminate unhandled exceptions in engine
-
-## **v4.1.0** UNRELEASED
```
shaopeng-gh, line 16: why do we delete release history

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
shaopeng-gh, line 229: >analyzeOptions.RichReturnCode != null && analyzeOptions.RichReturnCode.Value [](http://example.com/codeflow?start=19&length=76)

I think this line can be just as 
`analyzeOptions.RichReturnCode == true`

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -54,60 +46,63 @@ private bool IsValidScanTarget(string file)
                    MachOBinary.CanLoadBinary(uri);
         }
 
-        protected override BinaryAnalyzerContext CreateContext(AnalyzeOptions options, IAnalysisLogger logger, RuntimeConditions runtimeErrors, PropertiesDictionary policy = null, string filePath = null)
+        public override BinaryAnalyzerContext InitializeContextFromOptions(AnalyzeOptions options, ref BinaryAnalyzerContext context)
         {
-            if (logger is AggregatingLogger aggregatingLogger && this.Telemetry?.TelemetryClient != null)
+            if (this.Telemetry != null)
```
shaopeng-gh, line 51: >if (this.Telemetry != null) [](http://example.com/codeflow?start=12&length=27)

this should be changed back to at least check 
`this.Telemetry?.TelemetryClient != null`

suvamM: @shaopeng-gh But in that case, we would never end up creating an aggregating logger either. I see that was the situation before. For my understanding: if the TelemetryClient == null, we do not perform any logging? Is there a default logger we can pass to the aggregating logger if the telemetry client is null?

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
suvamM: @schlaman-ms Yes this needs to go away as well. We have a work item for removing it, but not sure if removing this would in scope for this PR.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
michaelcfanning: Yes, better. I am still resistant to direct comparisons to bool literals but it makes code more concise for nullable vars.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
michaelcfanning: It's a good question. In general, if we're breaking command-line already, it might be good to batch this up (and remove the verbose thing as well). Let me see how costly it is. Certainly these changes need to be more explicit in the release notes, too.

michaelcfanning: Turned out to be easy. Removed, updated release notes.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
michaelcfanning: The safest thing to do is put it back in the condition it was). I'm not sure we have a test that exercises this code directly from the options (our new pipeline model will allow us to write this). Also, we do not need to populate the logger if null here, the calling code will do that. The convention is, if the logger is null and you need to add one, creating an aggregating logger and add yours to it.

If the logger is non-null, in most cases it s/be an aggregating logger, just add yours.

There is one exception, there is a narrow window for console reporting where we don't yet know the reporting levels on the options/config.

suvamM (approved):

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
shaopeng-gh: When I ran the code, the `this.Telemetry` object is not null but the `this.Telemetry.TelemetryClient` property is null, the line below throw null ref error.

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/clang.stack_protector.a.sarif
```diff
@@ -42,12 +42,13 @@
...
           "location": {
-            "uri": "file:///Z:/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/clang.stack_protector.a"
+            "uri": "file:///Z:/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/clang.stack_protector.a",
+            "index": 0
           },
           "hashes": {
-            "md5": "2400EE99A6BB7B792A67A03A59385018",
-            "sha-1": "525CB4CBA8B8F80539BBC86C2D78A4275279D8F6",
-            "sha-256": "41BF336F8AB7D9A6246D37A6EC7210D4BE07893836890BAB0BFEDA98ACEDE87D"
+            "md5": "C60E01E6A2660A3359EACF5D00799FA0",
```
michaelcfanning, line 49: >C60E01E6A2660A3359EACF5D00799FA0 [](http://example.com/codeflow?start=20&length=32)

@shaopeng, can you explain why these file hashes changed when I rebaselined?

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/clangcl.pe.cpp.codeview.exe.sarif
```diff
@@ -3,983 +3,49 @@
...
+                  "physicalLocation": {
+                    "artifactLocation": {
+                      "uri": "file:///Z:/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/clangcl.pe.cpp.codeview.exe",
+                      "index": 0
+                    }
+                  }
                 }
+              ],
+              "message": {
+                "text": "'Z:\\src\\Test.FunctionalTests.BinSkim.Driver\\BaselineTestData\\clangcl.pe.cpp.codeview.exe' was not analyzed as its size (1178624 kilobytes) exceeds the currently configured threshold (1024 kilobytes)."
```
michaelcfanning, line 31: Oops! Need to fix this. BinSkim's default I assume should be int.MaxValue.

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/gcc.stack_protector.a.sarif
```diff
@@ -42,12 +42,13 @@
...
           "location": {
-            "uri": "file:///Z:/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/gcc.stack_protector.a"
+            "uri": "file:///Z:/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/gcc.stack_protector.a",
+            "index": 0
           },
           "hashes": {
-            "md5": "764B913022B393EAF22F96014401576F",
-            "sha-1": "4B49DA507C9212490EE09280A5A7AD1F1A389188",
-            "sha-256": "8548292CA34C14D090A1DE7327874010E3A7ACE5846B2E28D743CC0F35C069CB"
+            "md5": "B3AA4DD82C145C4B6E589953A2A2FAC4",
```
michaelcfanning, line 49: >B3AA4DD82C145C4B6E589953A2A2FAC4 [](http://example.com/codeflow?start=20&length=32)

Another one, Shaopeng. I'm guessing we recompiled these binaries. What this means is that changes in scan target hashes must not break our functional tests when that occurs?

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/clang.stack_protector.a.sarif
shaopeng-gh: a first glance these were created by Everett Maus not me,
looking...

shaopeng-gh: I checked both files in your branch. The correct hashes are the original ones...

shaopeng-gh (dismissed):
:clock1:

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/clang.stack_protector.a.sarif
shaopeng-gh: I suspect this is wrong
to get hash we now use:
`fileText = _fileSystem.FileReadAllText(path);`
and it was:
`Stream stream = fileSystem.FileOpenRead(fileName)`
a binary should not read by text?

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
shaopeng-gh: This is my only comment left

shaopeng-gh (approved):
:shipit:

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/gcc.stack_protector.a.sarif
shaopeng-gh: I'm guessing we recompiled these binaries. ---- no, this is a new bug in SDK, I will resolve this comment, please see my comment on the other.

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/clang.stack_protector.a.sarif
shaopeng-gh: @michaelcfanning @suvamM @schlaman-ms 
found the changeset that I mentioned at the end of meeting, this changeset have a few open comments from Michael, this particular one I looked and replied.
Do we think it is a issue,
does it need to be addressed within 4.1.0?

michaelcfanning: Thanks for recalling this issue!! Yes, we need to address this. We shouldn't be changing the hashing mechanism for binskim. Isn't this what we need?

  using (Stream stream = fileSystem.FileOpenRead(fileName))

shaopeng-gh: my understanding of the problem is, we have added a new cache `FileRegionsCache`, and that cache is used for something else, but since we have it, we also re-use it for hashing purpose, but because it is string, we can not it will be different value after hash.
So, should be fix be, change the cache to store binary, then the other place that we need string if we can read string from the binary then the cache is re-usable.

or, another way is just not reuse the string cache, when hashing read from the disk again.

walbourn:
Note this change broke ADO usage:

```
      - task: securedevelopmentteam.vss-secure-development-tools.build-task-binskim.BinSkim@4
        displayName: 'Run BinSkim'
        inputs:
          AnalyzeTargetBinskim: ''
          AnalyzeTargetGlob: +:file|out\bin\**\RelWithDebInfo\*.exe
          AnalyzeVerbose: true
```

Is there a newer task version that ignores "AnalyzeVerbse" or uses a newer switch set?

PR #867, opened 2023-04-20
Fix Spacing in README.md

lspsign2:
This is only fixing the format missing 2 blank lines per linter, safe to test merge in. 
Not visible change in the rendering of the md file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,9 @@ This repository contains the source code for BinSkim, a Portable Executable (PE)
 3. Find out more about the Static Analysis Results Interchange Format **([SARIF](https://github.com/sarif-standard/sarif-spec/))** used to output Binskim results
 
 ### How to extract the exe file from the nuget package
+
 If you only want to run the Binskim tool without installing anything, then you can 
+
 1. Download BinSkim from **[NuGet](https://www.nuget.org/packages/Microsoft.CodeAnalysis.BinSkim/)**
 2. Rename the file extension from .nupkg to .zip (ie. via commandline: `rename microsoft.codeanalysis.binskim.x.y.z.nupkg microsoft.codeanalysis.binskim.x.y.z.zip`)
 3. Unzip


shaopeng-gh (approved):
:shipit:

shaopeng-gh:
This PR from a fork, the changeset looks good to me, I have approved the PR.
The build will not run as we know, which bad for reviewer.
In the process we did asked the user to:
`3. Execute BuildAndTest.cmd at the root of the enlistment to validate before submitting a PR.`
so the process is flawed but makes sense overall.

michaelcfanning:
@shaopeng-gh, this is not the test I'm looking for. As noted here, this process is flawed and we're familiar with it. The flawed process is that although we can approve a PR, we can't merge it because the PR automation won't run.

What I'd like to do is run a different test: someone who is *not* a contributor in the repo should try to create a branch (*not fork*) and push the branch. Try to create a PR and see if that succeeds. With this test, we will also see whether automation runs.

What should happen, I think is that although the enlist will succeed, branch push will fail for these users or PR creation.

---
In reply to: [1517069883](https://github.com/microsoft/binskim/pull/867#issuecomment-1517069883) [](http://example.com/codeflow?ancestors=1517069883,1517069883)

shaopeng-gh:
> @shaopeng-gh, this is not the test I'm looking for. As noted here, this process is flawed and we're familiar with it. The flawed process is that although we can approve a PR, we can't merge it because the PR automation won't run.
> 
> What I'd like to do is run a different test: someone who is _not_ a contributor in the repo should try to create a branch (_not fork_) and push the branch. Try to create a PR and see if that succeeds. With this test, we will also see whether automation runs.
> 
> What should happen, I think is that although the enlist will succeed, branch push will fail for these users or PR creation.
> 
> In reply to: [1517069883](https://github.com/microsoft/binskim/pull/867#issuecomment-1517069883)

It is not possible to create branch if you are not a contributor. You can only filter, not create:

image

image

michaelcfanning:
Thanks, Shaopeng! I suspect I could have simply googled this, so thanks for the answer & verification. OK, so this leaves us again with your thorough analysis here. What I think we need to do is to update the docs to refine the process. Once a first-time contributor has produced a code change, they should submit a PR for the team to review. That PR cannot be checked in, as a fork cannot satisfy our automated validation process. Once approved, the BinSkim team will add the user as a contributor to the project. The user should then branch directly off main in a new PR (which will run test automation), after which it will be merged.

Let's talk about this more offline as well. 



---
In reply to: [1517099802](https://github.com/microsoft/binskim/pull/867#issuecomment-1517099802) [](http://example.com/codeflow?ancestors=1517099802,1517099802)

shaopeng-gh:
> Thanks, Shaopeng! I suspect I could have simply googled this, so thanks for the answer & verification. OK, so this leaves us again with your thorough analysis here. What I think we need to do is to update the docs to refine the process. Once a first-time contributor has produced a code change, they should submit a PR for the team to review. That PR cannot be checked in, as a fork cannot satisfy our automated validation process. Once approved, the BinSkim team will add the user as a contributor to the project. The user should then branch directly off main in a new PR (which will run test automation), after which it will be merged.
> 
> Let's talk about this more offline as well.
> 
> In reply to: [1517099802](https://github.com/microsoft/binskim/pull/867#issuecomment-1517099802)

@michaelcfanning 
I would suggest another way around this.
When we add a user to our repo, 
1. he can create the same PR, or create other code changes, or other branch and triggers the pipeline run. 
2. I do not think we actively remove a user once his PR is merged. Sometime later the user's account could be compromised and create changes and triggers the pipeline run.

The other way is to not add any user that is just a first time PR.
User will be forking as he would on any other repos, and do not grant any permissions to our repo.
when he submit the PR, the pipeline run will only trigger by us clicking the approve pipeline run.
So, what we do is to review the changes first, if looks good we trigger the pipeline run to verify the build and test etc.
And change the build to allow to run on a fork PR.
I think this is a better way if we are able to.

shaopeng-gh:
This PR was created for discussion only, see above.
The purpose is done, closing it.

PR #868, opened 2023-04-21
Fix `BA2004.EnableSecureSourceCodeHashing` to report the actual broken hash algorithm (rather than always reporting SHA-1).

schlaman-ms:
Bug Fix

In cases where a binary is encountered that is NOT SHA256, now the actual HASH type that was found will be displayed instead of the assumed SHA1.

Nothing in the logic changes at all. It is simply showing the actual HASH type in the output string.


ReleaseHistory.md
```diff
@@ -17,6 +17,7 @@
 * DEP: Update Sarif.Sdk submodule from [120fae3 to 2d52c53](https://github.com/microsoft/sarif-sdk/compare/120fae3...2d52c53). Full [SARIF SDK Release History](https://github.com/microsoft/sarif-sdk/blob/2d52c53/src/ReleaseHistory.md).
 * BRK: Remove `--verbose` command-line option (in favor of `--level` and `--kind`). [#853](https://github.com/microsoft/binskim/pull/853)
 * BRK: Remove `--hashes` command-line option (in favor of `--insert Hashes`). [#853](https://github.com/microsoft/binskim/pull/853)
+* BUG: Fix `BA2004.EnableSecureSourceCodeHashing` where SHA1 always displayed if not SHA256 [#868](https://github.com/microsoft/binskim/pull/868)
```
shaopeng-gh, line 20: >  [](http://example.com/codeflow?start=91&length=1)

add a "."

shaopeng-gh, line 20: >where SHA1 always displayed if not SHA256 [](http://example.com/codeflow?start=50&length=41)

(don't change according my feedback on this)
this sounds perfect for a dev but I think we can make it longer and more clear like:

incorrectly always report that `SHA1` is used if the binary is not hashed with `SHA256`, while it could also be `MD5`.

shaopeng-gh:
Build and test is failing, please try execute BuildAndTest.cmd at the root of the enlistment to validate.

ReleaseHistory.md
suvamM: How about: Fixed issue where default hash is SHA1, when binary isn't compiled with SHA256.

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -883,19 +882,64 @@
~
~        private ChecksumAlgorithmType ChecksumAlgorithmForFullPdb(Pdb pdb)
~        {
~            foreach (DisposableEnumerableView<Symbol> omView in pdb.CreateObjectModuleIterator())
~            {
~                Symbol om = omView.Value;
                 foreach (DisposableEnumerableView<SourceFile> sfView in pdb.CreateSourceFileIterator(om))
                 {
                     SourceFile sf = sfView.Value;
-                    return sf.HashType == HashType.SHA256 ? ChecksumAlgorithmType.Sha256 : ChecksumAlgorithmType.Sha1;
```
suvamM, line 892: consider switch instead of a bunch of if statements.

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -883,19 +882,64 @@
...
+            {
+                return ChecksumAlgorithmType.Sha256;
+            }
+
+            if (hashType == HashType.MD5)
+            {
+                return ChecksumAlgorithmType.Md5;
+            }
+
+            if (hashType == HashType.SHA1)
```
suvamM, line 910: >throw [](http://example.com/codeflow?start=12&length=5)

if switch is used, this would be default

src/BinSkim.Rules/RuleResources.Designer.cs
```diff
@@ -124,7 +124,7 @@ internal class RuleResources {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to &apos;{0}&apos; is a managed binary compiled with an insecure (SHA-1) source code hashing algorithm. SHA-1 is subject to collision attacks and its use can compromise supply chain integrity. Pass &apos;-checksumalgorithm:SHA256&apos; on the csc.exe command-line or populate the project &lt;ChecksumAlgorithm&gt; property with &apos;SHA256&apos; to enable secure source code hashing..
+        ///   Looks up a localized string similar to &apos;{0}&apos; is a managed binary compiled with an insecure ({1}) source code hashing algorithm. {1} is subject to collision attacks and its use can compromise supply chain integrity. Pass &apos;-checksumalgorithm:SHA256&apos; on the csc.exe command-line or populate the project &lt;ChecksumAlgorithm&gt; property with &apos;SHA256&apos; to enable secure source code hashing..
```
suvamM, line 127: Line to long. Consider splitting to multiline.

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -883,19 +882,64 @@
...
+        }
+
+        private ChecksumAlgorithmType GetChecksumAlgorithmType(Guid hashGuid)
+        {
+            if (hashGuid == Constant.Sha256Guid)
+            {
+                return ChecksumAlgorithmType.Sha256;
+            }
+
+            if (hashGuid == Constant.MD5Guid)
```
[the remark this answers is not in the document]
schlaman-ms: Since Constant.Sha256Guid  etc. are Static I cannot use them in a switch statement. Those are used elsewhere and I don't want to change them and risk a breaking change in other places those constants are used.
So... I am staying with if statements.

As for the InvalidOperation:

In the 1st method where the HashType enum is passed, then there are 4 known types (Sha256, Sha1, Md5, None).
Those obviously directly map to the 256, 1, md5 and Unknown enums of ChecksumAlgorithmType and the thinking of the InvalidOperationException is to catch the case where someone adds another HashType enum and passes that new value and this catches that this code was not updated to account for that new case.

In this method you have a Guid passed in:
The caller can pass ANY guid. This code can recognize Sha256, Md5, and Sha1. ANYTHING else will be seen as Unknown. Unknown is one of the 4 possible values of ChecksumAlgorithmType enum. It is logical that any other guid than the recognized 3 would be Unknown. It seems logical to map anything else but the known 3 guids to Unknown.
The other alternative would be to throw a InvalidOperationException instead of returning a ChecksumAlgorithmType of Unknown. 
In this case then Unknown would NEVER be returned from this method and Unknown would be an Unused Enum type.

schlaman-ms:
BuildAndTest is done and passing

---
In reply to: [1517352904](https://github.com/microsoft/binskim/pull/868#issuecomment-1517352904) [](http://example.com/codeflow?ancestors=1517352904,1517352904)

ReleaseHistory.md
```diff
@@ -17,6 +17,7 @@
 * DEP: Update Sarif.Sdk submodule from [120fae3 to 2d52c53](https://github.com/microsoft/sarif-sdk/compare/120fae3...2d52c53). Full [SARIF SDK Release History](https://github.com/microsoft/sarif-sdk/blob/2d52c53/src/ReleaseHistory.md).
 * BRK: Remove `--verbose` command-line option (in favor of `--level` and `--kind`). [#853](https://github.com/microsoft/binskim/pull/853)
 * BRK: Remove `--hashes` command-line option (in favor of `--insert Hashes`). [#853](https://github.com/microsoft/binskim/pull/853)
+* BUG: Fix `BA2004.EnableSecureSourceCodeHashing` where if hash is not SHA256, now the actual hash found is reported instead of a hardcoded SHA-1. [#868](https://github.com/microsoft/binskim/pull/868)
```
michaelcfanning, line 20: >* BUG: Fix `BA2004.EnableSecureSourceCodeHashing` where if hash is not SHA256, now the actual hash found is reported instead of a hardcoded SHA-1 [](http://example.com/codeflow?start=0&length=145)

* BUG: Fix `BA2004.EnableSecureSourceCodeHashing` where the reported broken hash was hard-coded to SHA-1 rather than reporting the actual algorithm (which, e.g., could be MD5).

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -889,13 +888,58 @@ private ChecksumAlgorithmType ChecksumAlgorithmForFullPdb(Pdb pdb)
...
+                    checksumAlgorithmType = ChecksumAlgorithmType.Sha256;
+                    break;
+                case HashType.MD5:
+                    checksumAlgorithmType = ChecksumAlgorithmType.Md5;
+                    break;
+                case HashType.SHA1:
+                    checksumAlgorithmType = ChecksumAlgorithmType.Sha1;
+                    break;
+                case HashType.None:
+                    checksumAlgorithmType = ChecksumAlgorithmType.Unknown;
```
michaelcfanning, line 914: >checksumAlgorithmType [](http://example.com/codeflow?start=20&length=21)

So what happens in this code path? Would we report an error? Is this an expected condition or is this another exceptional code path?

michaelcfanning (approved):
:shipit:

ReleaseHistory.md
schlaman-ms: That is a perfect title now. Somehow my way with words mode was offline :)
Updated.

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
schlaman-ms: The consumer of the code that gets the checksumAlgorithmType back will see that the type is NOT SHA256 therefore it will still be an error. It just affects what will be display. There is no further logic that tests the value (it is only if (hashFound != SHA256) then generate error and the message in this case would contain the word 'Unknown' for the the hash type found instead of 'Sha-1' or 'Md5'. So no code flow difference just a messaging change.

ReleaseHistory.md
michaelcfanning: Heh. :)

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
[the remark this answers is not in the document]
schlaman-ms: Hello @suvamM 

As for the closing of the PR. I actually did not have it as autocomplete and I did not close it. It was also coming up on the end of Friday and that was the goal to get 4.1 bugs completed as I understood.

--For the const issue and not using a switch for one of the methods:
You said its weird to have a bunch of if conditions for an enumerated type - I don't actually have if conditions for an enumerated type. I changed the method that has tests on an enumerated type to a switch.
For the other method that is a Guid, not an enumerated type, you mentioned about changing the static variables to const should have solved the problem but as I see it you cannot define a Guid as a const. -- ** Error CS0283 The type 'Guid' cannot be declared const.

Do you have a suggestion for how to do that? It is not like the existing code is unclear. Also, as I said, those constants are used in another method as well and I did not want to open that other method up to change and subsequent verification to make sure I didn't break something.

As for the divergence of behavior in the two methods.
I did not use the code to justify the logic as I determined the logic first and coded the two methods fresh accordingly.

You are absolutely correct that there is a divergence between the two methods. The reason for that is because there is a divergence in the set of possible values for input in the two methods.
When given an enum there are MD5, SHA1, SHA256  and Unknown. The idea behind throwing an exception is that if a new type comes along that was not known before, then the code wouldn't have accounted for that.
But in the case of a GUID being passed in, then if it is any value other than the 3 known hash type values, do you really want to throw an exception. The only thing that the value is used for is to state what the found Hash type was if it is not SHA256.
That would introduce into BinSkim a failure point where none existed before just because we didn't account for a given hash type to display. It is not used in logic flow, just in a message.
I guess the question is do we want to have these two methods be consistent in their behavior even though the available set of inputs is not consistent and then introduce a potential failure point where an exception is thrown instead of just saying an unknown Hash Type was found instead of SHA256.
That was my thinking on this.

@michaelcfanning, do you have any thoughts on this?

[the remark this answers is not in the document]
schlaman-ms: I guess one more quick thought is that it seems we don't want to risk having an exception thrown unless there is notable value.
In the case of finding an Enum that wasn't accounted for then that is an actual coding situation where things got out of step between adding an enum type and then the corresponding code not accounting for that new possible value. That seems like the perfect use for an exception and one that couldn't happen because of new values showing up in the wild.

With the Guid based function, then I guess we need to decide if we want to get rid of the sense of an 'Unknown' enum type and if so then we would want to get rid of Unknown in both HashType and ChecksumAlgorithmType (it's actually 'None' in this one). Then it would make sense to throw an exception if something was found other than SHA1, SHA256, MD5. That is a notable change. The original logic had no exception just Unknown for fall through. And this then would add a case where the code would throw and we would evaluate whether that is added value for a display only field.

It is an interesting issue with pro's and con's.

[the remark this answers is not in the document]
schlaman-ms: Awesome, I'm glad that helped clarify :)

PR #870, opened 2023-04-24
Add 2 tests for nightly

shaopeng-gh:
look and select typical PE files and typical non-PE files,
create two tests which plan to be warped in Coyote Test.
The two tests `AnalyzeCommand_TypicalPEFilesTest` and `AnalyzeCommand_TypicalNonPEFilesTest` can pass when running as a normal test.
tried to wrap in Coyote as the sample we have (please check if correct)
got an error I am not sure: 
`Bug Report: Deadlock detected. Op(0) is paused on a dependency, but no other controlled operations are enabled.`


suvamM:
Looking into this @shaopeng-gh

suvamM:
@shaopeng-gh I made a bunch of changes to enable Coyote testing on the tests you wrote. But the deadlock report persists. The tests pass by themselves, so will not block 4.1 release because of this, but will return to this item post release.

PR #871, opened 2023-04-26
Last Modified Date for PDBs

suvamM:
This PR makes the following changes:

1. It emits the latest modified date (UTC) for associated PDBs. This information is added as an additional payload in the properties bag of the `CompilerInformation` event type.

2. _Testing_: I tested the change using a bunch of test dll's and exe's in the BinSkim. The modified date was emitted successfully in AppInsights.


src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -286,6 +287,7 @@ public void WriteException(BinaryAnalyzerContext context, string errorMessage)
                 { "language", string.Empty },
                 { "debuggingFileName", string.Empty },
                 { "debuggingGuid", string.Empty },
+                 { "debuggingFileLastModifiedDateUTC", DateTime.MaxValue.ToString() },
```
suvamM, line 290: The reason I kept the default value to the maximum timestamp is as follows: the reason the last modified date was requested was to compare it with the date of the last build, and reject the dll if the pdb is stale. With a maximum value, we indicate the scenario that the date could not be determined, and the alert corresponding to the dll should be factored for mitigation.

src/BinSkim.Rules/PERules/BA4001.ReportPortableExecutableCompilerData.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Composition;
+using System.IO;
```
shaopeng-gh, line 8: I think this change is a new small feature,
we can add a `NEW: ` in ReleaseHistory.md

shaopeng-gh:
>         private void PrintHeader()

ah header

---
Refers to: src/BinSkim.Sdk/CompilerDataLogger.cs:183 in 6fffbf1. [](commit_id = 6fffbf14e1db214ed63da48187dbd6dd09154b48, deletion_comment = False)

src/BinSkim.Sdk/CompilerDataLogger.cs
shaopeng-gh: agree with the rationale here.

though I lean more towards:
here writes string.Empty, 
and above check if it is Datetime default then writes string.Empty
The user of the data do the checking if empty.

The benefit will be consistency, and easier to read.
anything empty is no data.
e.g.
`"C:\file.exe",,,,,,,,,`
compare to 
`"C:\file.exe",,,,"12/31/9999 11:59:59 PM",,,,,`

shaopeng-gh: ah I think of an possible issue, the class `CompilerData`and the `context.CompilerDataLogger.Write()` is used both in PE 4001 and non-PE 4002

I think we intent to change PE only, 
can you debug and see if 4002 is till working good?
I guess without any change, 4002 will write DateTime.Min because it is the default of not nullable DateTime type

shaopeng-gh (approved):
:shipit:

schlaman-ms (approved):
:shipit:

src/BinSkim.Sdk/CompilerDataLogger.cs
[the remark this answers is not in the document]
suvamM: Fixed.

suvamM: @shaopeng-gh OK that makes sense. Here's the change I made:
1. Switched the type of `DebuggingFileLastModifiedDateUTC` in `CompilerDate` from `DateTime` to `string`. 
2. Default value in `CompilerDataLogger` is now `string.Empty`
3. In `BA4001.ReportPortableExecutableCompilerData`, I put the read of the of the last modified date of the pdb in try-catch (since exceptions are possible). If no exceptions, then we get the actual modified date. If exceptions, we emit the empty string (so we let the user care about the interpretation of this).

Can you point me to test binaries on which I can test BA4002?

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -225,6 +225,7 @@ public void Write(BinaryAnalyzerContext context, CompilerData compilerData)
                 { "language", compilerData.Language },
                 { "debuggingFileName", compilerData.DebuggingFileName ?? string.Empty },
                 { "debuggingGuid", compilerData.DebuggingFileGuid ?? string.Empty },
+                { "debuggingFileLastModifiedDateUTC", compilerData.DebuggingFileLastModifiedDateUtc },
```
shaopeng-gh, line 228: >compilerData.DebuggingFileLastModifiedDateUtc [](http://example.com/codeflow?start=54&length=45)

the syntax here is different than the other strings
`compilerData.DebuggingFileGuid ?? string.Empty`
because the default of string is null not string empty
I think we should have this as well?

suvamM: Makes sense/ Fixed.

PR #884, opened 2023-05-04
Add test to repro error in Linux `ERR999.UnhandledEngineException` : `System.InvalidOperationException`: This operation is not supported for a relative URI.

shaopeng-gh:
This test will pass in Windows but fail in Linux and Mac, please see the Build result.


src/Test.FunctionalTests.BinSkim.Driver/AnalyzeCommandTests.cs
```diff
@@ -127,6 +127,34 @@ public void AnalyzeCommand_DeterminismTest()
             log.Runs[0].Invocations[0].ToolConfigurationNotifications.Count(t => t.Message.Text.Contains("E_PDB_NOT_FOUND")).Should().Be(1);
         }
 
+        [Fact]
+        public void AnalyzeCommand_ZeroByteTest()
```
shaopeng-gh, line 131: >public [](http://example.com/codeflow?start=8&length=6)

The test binaries are existing ones.
The reason I added one empty "WORKSPACE" file with 
one PE and one elf is,
if there is no valid file at all the output will be changed to no valid files instead, and we don't want that, so with them this test can run in both Windows and Linux.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -224,7 +224,7 @@ public override int Run(AnalyzeOptions analyzeOptions)
             // interesting (it will always be set).
 
             return analyzeOptions.RichReturnCode == true
-                ? (int)((uint)result & ~(uint)RuntimeConditions.RuleNotApplicableToTarget)
+                ? (int)((uint)result & ~(long)RuntimeConditions.RuleNotApplicableToTarget)
```
shaopeng-gh, line 227: >long [](http://example.com/codeflow?start=41&length=4)

this is needed by other change.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -224,7 +224,7 @@ public override int Run(AnalyzeOptions analyzeOptions)
...
                 : result;
```
schlaman-ms, line 228: It looks like result is an int and needs to use long. It looks like the long needs to be propagated back up, otherwise the new higher bits are truncated. Then there would not need to be any type casts here.

shaopeng-gh: This changeset 1. added a test 2. bringing in the latest SDK to get my fix for "This operation is not supported for a relative URI". 
This issue looks like comes from a checkin before my fix: https://github.com/microsoft/sarif-sdk/pull/2660
I have leave a message to get Michael's input.
@suvamM let me know how do we like to proceed from here,

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
suvamM: Looking at what the code is doing, we look at the bitvector of `RuleNotApplicableToTarget`, negate it (why?), and conjuct with `result`. Why not cast `result` to `long` as well? Otherwise the bit vectors are of unequal length. It's also strange that we finally cast it down to 32-bit signed `int`. The additional bits in `RuleNotApplicableToTarget` then make no difference?

@michaelcfanning we might need your feedback here.

src/Test.FunctionalTests.BinSkim.Driver/AnalyzeCommandTests.cs
suvamM: To make sure I understand this correctly, @shaopeng-gh you added 0-byte files, correct? Could we not do this by extending the `IFileSystem` mock, instead of adding additional files?

suvamM (changes requested):

src/Test.FunctionalTests.BinSkim.Driver/AnalyzeCommandTests.cs
```diff
@@ -127,6 +127,34 @@ public void AnalyzeCommand_DeterminismTest()
...
+            var options = new AnalyzeOptions
+            {
+                TargetFileSpecifiers = new string[] {
+                    pathDeterminismTest
+                },
+                Level = new[] { FailureLevel.Error, FailureLevel.Warning, FailureLevel.Note, FailureLevel.None },
+                Kind = new[] { ResultKind.Fail, ResultKind.Pass },
+                OutputFilePath = fileName,
+                OutputFileOptions = new[] { FilePersistenceOptions.ForceOverwrite },
+                Recurse = true,
```
suvamM, line 144: Do we need the `recurse` to be true here?

michaelcfanning (approved):
:shipit:

suvamM (approved):

Issue #718, opened 2022-10-18
Provide better reporting on encountering debug binaries (linked to msvcrtd.lib and friends


michaelcfanning:
Users should not ship the MS debug CRT: these aren't licensed for redistribution, neither are they hardened for security and performance. 

Users cannot satisfy most security compliance standards by analyzing non-shipping code such as debug builds.

BinSkim should therefore be thoughtful in its reporting if it detects an apparent debug build is being analyzed. Ideally, we would alert on this in general 'Do not scan debug binaries for security compliance purposes'. Minimally, we should make sure that rues such as our spectre analysis don't propose that users chase down and fix issues in these libraries that won't, by design, be fixed (because they are not candidates for shipping and therefore aren't comprehensively hardened for security).

- MSVCRTD.lib
- MSVCMRTD.LIB
- vcruntimed.lib

dmachaj:
The performance rules added some detection of a debug C runtime.  That logic might be useful to you.  See [https://github.com/microsoft/binskim/blob/main/src/BinaryParsers/PEBinary/ProgramDatabase/CompilerCommandLine.cs#L127](https://github.com/microsoft/binskim/blob/main/src/BinaryParsers/PEBinary/ProgramDatabase/CompilerCommandLine.cs#L127).

michaelcfanning:
Thanks! We will leverage this. My apologies for the egregious delay closing on your open performance rules, I will add my review by EOW.

PR #887, opened 2023-05-09
Fix false positive for rule `BA2024.EnableSpectreMitigations` incorrectly flags compilation units using debug runtime (which are not Spectre-mitigated by design).

shaopeng-gh:
BinSkim’s BA2024 rule warns on debug VC runtime libraries, e.g.:

```
'xxx.dll' was compiled with one or more modules that do not enable code generation mitigations for speculative execution side-channel attack (Spectre) vulnerabilities. Spectre attacks can compromise hardware-based isolation, allowing non-privileged users to retrieve potentially sensitive data from the CPU cache. To resolve the issue, provide the /Qspectre switch on the compiler command-line (or /d2guardspecload in cases where your compiler supports this switch and it is not possible to update to a toolset that supports /Qspectre). This warning should be addressed for code that operates on data that crosses a trust boundary and that can affect execution, such as parsing untrusted file inputs or processing query strings of a web request.
The following modules were compiled with a toolset that supports /Qspectre but the switch was not enabled on the command-line:
MSVCRTD.lib,cxx,19.33.31616.1 (…)
MSVCRTD.lib,c,19.33.31616.1 (…)
vcruntimed.lib,cxx,19.33.31616.1 (…)
```

Since those debug libraries are not Spectre-mitigated by design, I think we should either suppress warnings for those libraries globally.  For the record, there is BinSkim issue on better handling this case:
https://github.com/microsoft/binskim/issues/718


src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/clangcl.pe.cpp.codeview.exe.sarif
```diff
@@ -381,11 +381,12 @@
         {
           "ruleId": "BA2024",
           "ruleIndex": 16,
+          "kind": "pass",
```
shaopeng-gh, line 384: >pass [](http://example.com/codeflow?start=19&length=4)

In the baseline changes,
all built with debug (including this one with -Mtd)
was failing now will pass.

src/BinSkim.Rules/PERules/BA2024.EnableSpectreMitigations.cs
```diff
@@ -145,6 +145,9 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
                 Symbol om = omView.Value;
                 ObjectModuleDetails omDetails = om.GetObjectModuleDetails();
 
+                // Debug libraries are not Spectre-mitigated by design.
```
schlaman-ms, line 148: Small nit for fast readability and consistency:
I'd rather see:

if (.....)
{
    continue;
}

shaopeng-gh: fixed.

schlaman-ms (approved):
:shipit:

suvamM (approved):
:shipit:

PR #896, opened 2023-05-16
Fix 4.1.0 issue: `System.InvalidOperationException`: `Sequence contains more than one matching element` when `--trace` is provided

shaopeng-gh:
## Issue:
Found a bug In 4.1.0 , if `--trace` is provided will have unhandled exception.
The reason is new new pipeline change has SDK traces changed to trace, because we overwrite in BinSkim, we need to change the name ass well.

This is the full error:
image

### DIFF
--- a/src/BinSkim.Driver/AnalyzeOptions.cs
+++ b/src/BinSkim.Driver/AnalyzeOptions.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 
@@ -13,7 +14,7 @@ namespace Microsoft.CodeAnalysis.IL
     [Verb("analyze", HelpText = "Analyze one or more binary files for security and correctness issues.")]
     public class AnalyzeOptions : AnalyzeOptionsBase
     {
-        private IEnumerable<string> trace;
+        private IEnumerable<string> trace = Array.Empty<string>();
         [Option(
             "trace",
             Separator = ';',
--- a/src/BinSkim.Driver/AnalyzeOptions.cs
+++ b/src/BinSkim.Driver/AnalyzeOptions.cs
@@ -1,8 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Collections.Generic;
+using System.Linq;
 
 using CommandLine;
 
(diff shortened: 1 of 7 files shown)


src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -171,7 +171,7 @@ public override int Run(AnalyzeOptions analyzeOptions)
         {
             Stopwatch stopwatch = null;
 
-            if (analyzeOptions.Traces.Where(s => s == "ScanTime").Any())
+            if (analyzeOptions.Trace.Where(s => s == "ScanTime").Any())
```
michaelcfanning, line 174: You should be using the Traces property in the base class.

[https://github.com/microsoft/sarif-sdk/blob/140d730afd838e7e168985bfda38edd4ead7d297/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs#L86](https://github.com/microsoft/sarif-sdk/blob/140d730afd838e7e168985bfda38edd4ead7d297/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs#L86)

src/Test.FunctionalTests.BinSkim.Driver/BaselineTests.cs
```diff
@@ -438,7 +438,7 @@ private SarifLog RunRules(StringBuilder sb, string inputFileName, Sarif.SarifVer
                 ConfigurationFilePath = "default",
                 SarifOutputVersion = Sarif.SarifVersion.Current,
                 TargetFileSpecifiers = new string[] { inputFileName },
-                Traces = Array.Empty<string>(),
+                Trace = Array.Empty<string>(),
```
michaelcfanning, line 441: >Trace [](http://example.com/codeflow?start=16&length=5)

Why is this required? Can you remove it?

michaelcfanning (changes requested):
:clock1:

src/BinSkim.Driver/AnalyzeOptions.cs
```diff
@@ -21,7 +21,7 @@ public class AnalyzeOptions : AnalyzeOptionsBase
                        "(escape semicolon with backslash in Unix-like OS), that " +
                        "should be emitted to the console and log file (if appropriate). " +
                        "Valid values: PdbLoad.")]
```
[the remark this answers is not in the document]
shaopeng-gh: Yes! found a way and fixed!

src/Test.FunctionalTests.BinSkim.Driver/BaselineTests.cs
shaopeng-gh: removed.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
shaopeng-gh: Please see other comments first, here can either use the current one or base one since current one incudes all.
The base one will throw null exception where 
`this.Trace` is null, so I kept the current one I have.

``` CSharp
public DefaultTraces Traces
        {
            get
            {
                defaultTraces ??=
                    this.Trace.Any()
                        ? (DefaultTraces)Enum.Parse(typeof(DefaultTraces), string.Join(",", this.Trace))
                        : DefaultTraces.None;

                return this.defaultTraces.Value;
            }
        }
```

shaopeng-gh:
Converted to draft the new direction is to fix AnalyzeOptionsBase. It should declare ‘Traces’ as an Iset<string> not a DefaultTraces enum. Working on it.

shaopeng-gh:
> Converted to draft the new direction is to fix AnalyzeOptionsBase. It should declare ‘Traces’ as an Iset not a DefaultTraces enum. Working on it.

Checked the SDK, the `DefaultTraces Traces` is not used I have removed: https://github.com/microsoft/sarif-sdk/pull/2676
This PR is ready to review.

michaelcfanning (approved):
:shipit:

PR #900, opened 2023-05-17
[For Compare] Add Binary File Hash Test Without SDK update to the fix commit

shaopeng-gh:
The SDK change was approved and merged, but the apply it to BinSkim https://github.com/microsoft/binskim/pull/891 was not approved, so currently the fix is not "in" for BinSkim. So this test will FAIL.
This is just for compare with the other PR, which also update the SDK to the commit of my fix in SDK:
https://github.com/microsoft/binskim/pull/901


shaopeng-gh:
Closed, the PR was reverted.

PR #903, opened 2023-05-17
Update Microsoft.CodeAnalysis.NetAnalyzers 7.0.0 --> 7.0.1

schlaman-ms:
There is a new version of the Microsoft.CodeAnalysis.NetAnalyzers package available (7.0.1)

Warnings were being generated that a new version of the NetAnalyzers were availble.
This PR will update all of the package references to the new version 7.0.1

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -16,7 +16,7 @@
 - NEW => new feature 
 
 ## UNRELEASED
-* DEP: Update 'Microsoft.CodeAnalysis.NetAnalyzers' package from 7.0.0 to 7.0.1 to resolve build warnings. [#903](https://github.com/microsoft/binskim/pull/903)
+* DEP: Update `Microsoft.CodeAnalysis.NetAnalyzers` package from 7.0.0 to 7.0.1 to resolve build warnings. [#903](https://github.com/microsoft/binskim/pull/903)
 
 ## **v4.1.0**
 * NEW: `CompilerInformation` telemetry now emits the last modified date of the scan target. [#873](https://github.com/microsoft/binskim/pull/873)
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -16,7 +16,7 @@
 - NEW => new feature 
 
 ## UNRELEASED
-* DEP: Update 'Microsoft.CodeAnalysis.NetAnalyzers' package to 7.0.1 to resolve build warnings. [#903](https://github.com/microsoft/binskim/pull/903)
+* DEP: Update 'Microsoft.CodeAnalysis.NetAnalyzers' package from 7.0.0 to 7.0.1 to resolve build warnings. [#903](https://github.com/microsoft/binskim/pull/903)
 
 ## **v4.1.0**
 * NEW: `CompilerInformation` telemetry now emits the last modified date of the scan target. [#873](https://github.com/microsoft/binskim/pull/873)
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -16,7 +16,7 @@
 - NEW => new feature 
 
 ## UNRELEASED
-* DEP: Eliminate build warning - The .NET SDK has newer analyzers with version '7.0.1' than what version of '7.0.0' of 'Microsoft.CodeAnalysis.NetAnalyzers' package provides. Update or remove this package reference. [#903](https://github.com/microsoft/binskim/pull/903)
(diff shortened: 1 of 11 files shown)


src/BinaryParsers/BinaryParsers.csproj
```diff
@@ -44,7 +44,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Update="Microsoft.CodeAnalysis.NetAnalyzers" Version="7.0.0">
```
shaopeng-gh, line 47: >7.0.0 [](http://example.com/codeflow?start=76&length=5)

I also see we have 6.0.0
should we upgrade them as well

src/BinaryParsers/BinaryParsers.csproj
```diff
@@ -44,7 +44,7 @@
...
+    <PackageReference Update="Microsoft.CodeAnalysis.NetAnalyzers" Version="7.0.1">
```
shaopeng-gh, line 47: >7.0.1 [](http://example.com/codeflow?start=76&length=5)

dependency upgrade can use a Release History.
`DEP`

src/BinaryParsers/BinaryParsers.csproj
schlaman-ms: Yes, what made sense is to update ONLY the build.common.props and to remove ALL the other NetAnalyzer references as they are just overriding what is in build.common.props.

Now with build.common.props updated to 7.0.1 then everything will use that.

src/BinaryParsers/BinaryParsers.csproj
```diff
@@ -43,10 +43,4 @@
     </EmbeddedResource>
   </ItemGroup>
 
-  <ItemGroup>
```
shaopeng-gh, line 46: > <ItemGroup> [](http://example.com/codeflow?start=1&length=12)

nit: there is a empty line, but it does not matter.

shaopeng-gh (approved):
:shipit:

suvamM (changes requested):
There should be a change noted in ReleaseHistory as a DEP.

ReleaseHistory.md
```diff
@@ -12,9 +12,11 @@
 - FND => False negative reduction in dynamic phase
 - UER => eliminate unhandled exceptions in rules
 - UEE => eliminate unhandled exceptions in engine
+- DEP => upgrade dependency versions
 - NEW => new feature 
 
 ## UNRELEASED
+* DEP: Eliminate build warning - The .NET SDK has newer analyzers with version '7.0.1' than what version of '7.0.0' of 'Microsoft.CodeAnalysis.NetAnalyzers' package provides. Update or remove this package reference. [#903](https://github.com/microsoft/binskim/pull/903)
```
suvamM, line 19: The language is inconsistent with other DEP updates. We should just say NetAnalyzers updated from 7.0.0 to 7.0.1

suvamM (changes requested):

ReleaseHistory.md
schlaman-ms: Ok, changed to be consistent.

suvamM: @schlaman-ms Can you add "from 7.0.0" before "to 7.0.1". Most of the other DEPs follow this convention, and it makes sense because it makes the update self contained.

schlaman-ms: Ok, done. Yes like I had it originally but with other wording.

ReleaseHistory.md
```diff
@@ -12,9 +12,11 @@
 - FND => False negative reduction in dynamic phase
 - UER => eliminate unhandled exceptions in rules
 - UEE => eliminate unhandled exceptions in engine
+- DEP => upgrade dependency versions
 - NEW => new feature 
 
 ## UNRELEASED
+* DEP: Update 'Microsoft.CodeAnalysis.NetAnalyzers' package from 7.0.0 to 7.0.1 to resolve build warnings. [#903](https://github.com/microsoft/binskim/pull/903)
```
suvamM, line 19: @schlaman-ms I think there is an issue here: Microsoft.CodeAnalysis.NetAnalyzers should be within backticks, not single quotes. Can you check if it is rendering corectly?

schlaman-ms: Yup, missed that. Not so used to the backtick for markdown. But I won't miss that again though.

suvamM (approved):

PR #908, opened 2023-05-24
* NEW: BinSkim will now log warning `OneOrMoreSymbolServersProvidedInaccessible` if can not connect to the http symbol servers provided by `--sympath`

shaopeng-gh:
### Issue:

User provided:
`--sympath srv*https://username:password@artifacts.dev.azure.com/Org/_apis/Symbol/symsrv`
It does not work for some reason. It is running on server which is less easy to verify if issues unrelated to BinSkim itself:
1. network restrictions the url can not be reached
2. user name and password wrong

### Change:
Add the probe and when the user provided http symbol server can not be reached, we log a warning message.

### DIFF
--- a/src/BinSkim.Driver/DriverUtilities.cs
+++ b/src/BinSkim.Driver/DriverUtilities.cs
@@ -1,13 +1,17 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Composition.Convention;
 using System.Composition.Hosting;
+using System.Net;
+using System.Net.Http;
 using System.Reflection;
 
 using Microsoft.CodeAnalysis.IL.Rules;
+using Microsoft.CodeAnalysis.Sarif;
 
 namespace Microsoft.CodeAnalysis.IL
 {
@@ -44,6 +48,36 @@ namespace Microsoft.CodeAnalysis.IL
                 .Export<T>();
 
             return conventions;
+        }
+
+        public static IList<string> GetInaccessibleSymbolPaths(string symbolPathString)
+        {
+            IList<string> inaccessibleList = new List<string>();
+            if (string.IsNullOrWhiteSpace(symbolPathString)) { return inaccessibleList; }
+
+            string[] symbolServers = Array.FindAll(symbolPathString.Split('*'), path =>
+            path.StartsWith("http://", StringComparison.OrdinalIgnoreCase) ||
+            path.StartsWith("https://", StringComparison.OrdinalIgnoreCase));
+
+            foreach (string symbolServer in symbolServers)
+            {
+                try
+                {
+                    var httpClient = new HttpClientWrapper();
+                    HttpResponseMessage httpResponseMessage = httpClient.GetAsync(symbolServer).GetAwaiter().GetResult();
+
+                    if (httpResponseMessage.StatusCode != HttpStatusCode.OK)
+                    {
+                        inaccessibleList.Add(symbolServer);
+                    }
+                }
+                catch
+                {
+                    inaccessibleList.Add(symbolServer);
+                }
+            }
+
+            return inaccessibleList;
         }
     }
 }
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -69,6 +69,24 @@ namespace Microsoft.CodeAnalysis.IL
 
             // Update context object based on command-line parameters.
             context.SymbolPath = options.SymbolsPath;
+
+            IList<string> inaccessibleSymbolPaths = DriverUtilities.GetInaccessibleSymbolPaths(context.SymbolPath);
+
+            if (inaccessibleSymbolPaths.Count > 0)
+            {
+                // These symbol servers provided are inaccessible: {0}.
+                context.Logger.LogConfigurationNotification(
+                    Errors.CreateNotification(
+                        context.CurrentTarget?.Uri,
+                        "OneOrMoreSymbolServersProvidedInaccessible",
+                        ruleId: null,
+                        FailureLevel.Warning,
+                        exception: null,
+                        persistExceptionStack: false,
+                        messageFormat: RuleResources.OneOrMoreSymbolServersProvidedInaccessible,
+                        string.Join(";", inaccessibleSymbolPaths)));
+            }
+
             context.IgnorePdbLoadError = options.IgnorePdbLoadError;
             context.LocalSymbolDirectories = options.LocalSymbolDirectories;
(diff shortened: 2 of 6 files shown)


src/BinSkim.Driver/DriverUtilities.cs
```diff
@@ -45,5 +49,35 @@ private static ConventionBuilder GetConventions<T>()
...
+
+            string[] symbolServers = Array.FindAll(symbolPathString.Split('*'), path =>
+            path.StartsWith("http://", StringComparison.OrdinalIgnoreCase) ||
+            path.StartsWith("https://", StringComparison.OrdinalIgnoreCase));
+
+            foreach (string symbolServer in symbolServers)
+            {
+                try
+                {
+                    var httpClient = new HttpClientWrapper();
```
suvamM, line 66: @shaopeng-gh Wanted to cross-check that this invocation does not return a new `HttpClient` in each iteration. If so, we should create a single client and re-use that.

src/BinSkim.Driver/DriverUtilities.cs
```diff
@@ -45,5 +49,35 @@ private static ConventionBuilder GetConventions<T>()
...
+                    HttpResponseMessage httpResponseMessage = httpClient.GetAsync(symbolServer).GetAwaiter().GetResult();
```
suvamM, line 67: We are doing to an HTTP GET here: what is the authentication model? Binskim mostly runs in the context of Guardian---will that have sufficient privileges? Or will the majority of these checks result in an inaccessible error?

src/Test.UnitTests.BinSkim.Driver/DriverUtilitiesTests.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.CodeAnalysis.IL;
+
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.BinSkim.Rules
+{
+    public class DriverUtilitiesTests
+    {
+        [Fact]
+        public void GetInaccessibleSymbolPathsTests()
```
suvamM, line 17: @shaopeng-gh This isn't really a unit test right? Will these result in actual network calls? If yes, then you need to mock the `HttpClient` to make this a unit test. We need separate integration tests to ensure that network calls work properly, but again the authentication model is unclear to me.

src/BinSkim.Driver/DriverUtilities.cs
```diff
@@ -45,5 +49,35 @@
 
             return conventions;
         }
+
+        public static IList<string> GetInaccessibleSymbolPaths(string symbolPathString)
```
michaelcfanning, line 53: >public [](http://example.com/codeflow?start=8&length=6)

make this internal

src/BinSkim.Driver/DriverUtilities.cs
```diff
@@ -45,5 +49,35 @@
...
+        {
+            IList<string> inaccessibleList = new List<string>();
+            if (string.IsNullOrWhiteSpace(symbolPathString)) { return inaccessibleList; }
+
+            string[] symbolServers = Array.FindAll(symbolPathString.Split('*'), path =>
+            path.StartsWith("http://", StringComparison.OrdinalIgnoreCase) ||
```
michaelcfanning, line 59: >path [](http://example.com/codeflow?start=12&length=4)

this should be indented.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -69,6 +69,24 @@ public override BinaryAnalyzerContext InitializeGlobalContextFromOptions(Analyze
...
+                // These symbol servers provided are inaccessible: {0}.
+                context.Logger.LogConfigurationNotification(
+                    Errors.CreateNotification(
+                        context.CurrentTarget?.Uri,
+                        "OneOrMoreSymbolServersProvidedInaccessible",
+                        ruleId: null,
+                        FailureLevel.Warning,
+                        exception: null,
+                        persistExceptionStack: false,
+                        messageFormat: RuleResources.OneOrMoreSymbolServersProvidedInaccessible,
```
michaelcfanning, line 86: >OneOrMoreSymbolServersProvidedInaccessible [](http://example.com/codeflow?start=53&length=42)

OneOrMoreSymbolServersAreNotAccessible.

When you create user-facing strings, please provide a real example as surfaced by the tool.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -69,6 +69,24 @@ public override BinaryAnalyzerContext InitializeGlobalContextFromOptions(Analyze
 
             // Update context object based on command-line parameters.
             context.SymbolPath = options.SymbolsPath;
+
+            IList<string> inaccessibleSymbolPaths = DriverUtilities.GetInaccessibleSymbolPaths(context.SymbolPath);
+
+            if (inaccessibleSymbolPaths.Count > 0)
+            {
+                // These symbol servers provided are inaccessible: {0}.
```
michaelcfanning, line 77: > // These symbol servers provided are inaccessible: {0}. [](http://example.com/codeflow?start=15&length=56)

The following symbol servers were specified via the --sympath argument but can't be reached:

michaelcfanning: One or more servers specified in the --sympath argument could not be reached ([https://symweb](https://symweb), [https://blah](https://blah)).

src/Test.UnitTests.BinSkim.Driver/DriverUtilitiesTests.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            DriverUtilities.GetInaccessibleSymbolPaths(null)
+                .Should().BeEquivalentTo(Array.Empty<string>());
```
michaelcfanning, line 20: >()); [](http://example.com/codeflow?start=60&length=4)

Use xunit scopes here?

Test the output of your test  failures please!

src/Test.UnitTests.BinSkim.Driver/DriverUtilitiesTests.cs
```diff
@@ -0,0 +1,31 @@
...
+            DriverUtilities.GetInaccessibleSymbolPaths("")
+                .Should().BeEquivalentTo(Array.Empty<string>());
+            DriverUtilities.GetInaccessibleSymbolPaths("SRV*http://msdl.microsoft.com/download/symbols")
+                .Should().BeEquivalentTo(Array.Empty<string>());
+            DriverUtilities.GetInaccessibleSymbolPaths("SRV*c:\\symbols*http://msdl.microsoft.com/download/symbols")
+                .Should().BeEquivalentTo(Array.Empty<string>());
+            DriverUtilities.GetInaccessibleSymbolPaths("CACHE*c:\\symbols;SRV*http://notexists/SymStore*http://notexists2/SymStore")
```
michaelcfanning, line 27: >notexists [](http://example.com/codeflow?start=85&length=9)

[https://does.notexist.example.com](https://does.notexist.example.com)

PR #915, opened 2023-05-31
Fix build error after update the SDK submodule in BinSkim

shaopeng-gh:
Found an issue, when Binskim update SDK Submodule to latest will break the build. I narrow down to it is caused by this PR: https://github.com/microsoft/sarif-sdk/pull/2669
The added "Microsoft.Diagnostics.Tracing.TraceEvent" in SDK contains a Dia2Lib.dll which will be brought into BinSkim and conflict with our existing Dia2Lib.dll. Can't seem to remove it, and this version will not work with our code missing classes. Not sure how to resolve this. 
The Dia2Lib.dll in TraceEvent has a different public key than the existing one we have, and is version 2.0.0.0.
The existing one we have don't have version number.  

---- found a possible fix, to replace TraceEvent Dia2Lib.dll with our own. 

The build will pass after the other issue is fixed, so need to go after this PR:
https://github.com/microsoft/binskim/pull/896


NuGet.Config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <config>
+    <add key="globalPackagesFolder" value="./src/packages" />
```
shaopeng-gh, line 4: we have similar settings in SDK with 2 settings, based on building and testing only this one is needed.

ReleaseHistory.md
```diff
@@ -17,6 +17,7 @@
 
 ## UNRELEASED
 * DEP: Update `Microsoft.CodeAnalysis.NetAnalyzers` package from 7.0.0 to 7.0.1 to resolve build warnings. [#903](https://github.com/microsoft/binskim/pull/903)
+* DEP: Update Sarif.Sdk submodule from [bc8cb57 to dd8b7b8](https://github.com/microsoft/sarif-sdk/compare/bc8cb57...dd8b7b8). Full [SARIF SDK Release History](https://github.com/microsoft/sarif-sdk/blob/dd8b7b8/ReleaseHistory.md).
```
shaopeng-gh, line 20: this is the current latest SDK (when this PR created)

src/BinSkim.Driver/BinSkim.Driver.csproj
```diff
@@ -37,7 +37,7 @@
 
   <ItemGroup>
     <Reference Include="dia2lib">
-      <HintPath>..\..\refs\dia2lib.dll</HintPath>
+      <HintPath>..\..\refs\Dia2Lib.dll</HintPath>
```
shaopeng-gh, line 40: The name changed is needed because in Linux it is case sensitive.

src/BinaryParsers/BinaryParsers.csproj
```diff
@@ -43,4 +43,14 @@
     </EmbeddedResource>
   </ItemGroup>
 
-</Project>
+  <PropertyGroup>
+    <PreBuildEvent Condition="$(OS) == 'Windows_NT'">
+      echo Copying file from "$(MSBuildProjectDirectory)\..\..\refs\Dia2Lib.dll" to "$(MSBuildProjectDirectory)\..\..\src\packages\microsoft.diagnostics.tracing.traceevent\3.1.3\lib\netstandard2.0\"
+      copy "$(MSBuildProjectDirectory)\..\..\refs\Dia2Lib.dll" "$(MSBuildProjectDirectory)\..\..\src\packages\microsoft.diagnostics.tracing.traceevent\3.1.3\lib\netstandard2.0\"
```
shaopeng-gh, line 49: This is to replace with our version of Dia2Lib.dll that works with our code.

src/BinSkim.Driver/BinSkim.Driver.csproj
suvamM: @shaopeng-gh What exactly is the issue when the SARIF SDK `Dia2Lib.dll` is used?

suvamM (changes requested):

src/BinSkim.Driver/BinSkim.Driver.csproj
shaopeng-gh: misssing classes: IDiaEnumSegments, IDiaSectionContrib etc.
conflict classes: IDiaDataSource2, IDiaDataSource3
image

src/BinaryParsers/BinaryParsers.csproj
shaopeng-gh: This is good point if it will break the generation of events.
I am not familiar about it,
I see the code we are doing check like:
`if (!string.IsNullOrEmpty(globalContext.EventsFilePath))`
to enable generation of events, and the setting of `Context.EventsFilePath` is:
`context.EventsFilePath = Environment.GetEnvironmentVariable("SPMI_ETW") ?? options.EventsFilePath ?? context.EventsFilePath;`
Is this only enabled in SPMI?

michaelcfanning (approved):
:shipit:

shaopeng-gh:
Note: We will explore upgrading of the lib before revisit this.

ReleaseHistory.md
```diff
@@ -16,6 +16,7 @@
 - NEW => new feature 
 
 ## UNRELEASED
+* DEP: Update Sarif.Sdk submodule from [bc8cb57 to fd6e615](https://github.com/microsoft/sarif-sdk/compare/bc8cb57...fd6e615). Full [SARIF SDK Release History](https://github.com/microsoft/sarif-sdk/blob/fd6e615/ReleaseHistory.md).
```
shaopeng-gh, line 19: >* DEP [](http://example.com/codeflow?start=0&length=5)

Reopen this PR.
The was for: if update SARIF sub module, the Dia will not work. At that moment I created this PR but we did not have any urgent fix we need from SARIF SDK. 
Now there is fix that is needed, I reopen this PR and update to point to latest SDK.
There is no new test added, but without this fix many tests will fail.
Please see PR description for some info.

shaopeng-gh:
our code uses the old classes, if re-create the dll, our current code that uses the dll will need a major re-write.

---
In reply to: [1591721869](https://github.com/microsoft/binskim/pull/915#issuecomment-1591721869) [](http://example.com/codeflow?ancestors=1591721869)

stacywray (changes requested):
Left an inline suggestion for ReleaseHistory.md

src/BinSkim.Driver/BinSkim.Driver.csproj
scottoneil-ms: If there isn't a reason *not* to change the Reference itself (as opposed to the hint), can you please change it to agree with the hint?

scottoneil-ms: Same comment for lines below.

ReleaseHistory.md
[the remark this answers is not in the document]
shaopeng-gh: Fixed. I left the `full` word out of the link, let me know if you think we should include it.

scottoneil-ms (approved):
Thanks!

src/BinSkim.Driver/BinSkim.Driver.csproj
shaopeng-gh: updated the 
`<Reference Include="dia2lib">`
to
`<Reference Include="Dia2Lib">`.
I have searched the whole repo there is 0 lower case of `dia2lib`.

stacywray (approved):
No need to include "full" in the link. Looks good.

ReleaseHistory.md
```diff
@@ -16,6 +16,7 @@
 - NEW => new feature 
 
 ## UNRELEASED
+* DEP: Update `Sarif.Sdk` submodule from [bc8cb57 to fd6e615](https://github.com/microsoft/sarif-sdk/compare/bc8cb57...fd6e615). Reference full [SARIF SDK Release History](https://github.com/microsoft/sarif-sdk/blob/fd6e615/ReleaseHistory.md).
```
michaelcfanning, line 20: This line is inconsistent with other entries in this file: they all need to be consistent. I think Stacy recommended removing the word `Full`, as below:

if that is correct, please update the entire release history to bring all of these DEP entries into conformance. Also someone needs to make sure our release notes maintenance documentation has the correct standard listed there.

> * DEP: Update `Sarif.Sdk` submodule from [bc8cb57 to fd6e615](https://github.com/microsoft/sarif-sdk/compare/bc8cb57...fd6e615). [SARIF SDK Release History](https://github.com/microsoft/sarif-sdk/blob/fd6e615/ReleaseHistory.md).

michaelcfanning: Also, why exactly did you need to make this dependency update? It is a good practice to cite the specific incoming change that's critical to binskim.

stacywray (approved):
LGTM.

PR #926, opened 2023-06-20
BA2004.EnableSecureSourceCodeHashing - For native binaries, include actual Hash instead of just saying insecure hash.

schlaman-ms:
# The actual actual Hash will be shown now for the libraries instead of just that it was an insecure hash. 
## If no hash was specified then you will now be able to see that as well.

For Direct compilands, if there is an insecure hash then this is an Error.
For Direct compilands, if there is no hash then this is a warning.

For statically linked, if no hash or insecure hash then this is a warning.


ReleaseHistory.md
```diff
@@ -16,6 +16,7 @@
 - NEW => new feature 
 
 ## UNRELEASED
+* NEW: `BA2004.EnableSecureSourceCodeHashing` Provide actual hash in BA2004 reporting of native binaries instead of just saying insecure hash. [926](https://github.com/microsoft/binskim/pull/926)
```
michaelcfanning, line 19: >NEW [](http://example.com/codeflow?start=2&length=3)

This release note isn't a complete sentence.

`BA2004.EnableSecureSourceCodeHashing` now explicitly reports the insecure hash algorithm found (or reports no hash information was present in that circumstance). [926](https://github.com/microsoft/binskim/pull/926)
* NR :

ReleaseHistory.md
```diff
@@ -16,6 +16,7 @@
 - NEW => new feature 
 
 ## UNRELEASED
```
michaelcfanning, line 18: >UNRELEASED [](http://example.com/codeflow?start=3&length=10)

the notes in this section and in 4.1.0 are not ordered properly.

DEP always comes first for example. Please refer to the release notes maintenance guide to identify the proper ordering (and add that information to the doc if it isn't there.

See 4.0.0 for the correct ordering.

src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs
```diff
@@ -194,75 +200,153 @@ public void AnalyzeNativeBinaryAndPdb(BinaryAnalyzerContext context)
 
                     if (sf.HashType != HashType.SHA256)
                     {
+                        // Adding the `MD5` or `SHA1` etc, comment delimited by a backquote/backtick (`) to the library name
```
michaelcfanning, line 203: >Adding [](http://example.com/codeflow?start=27&length=6)

Is this a useful comment? Where is the period at the end of your sentence? Please make sure our coding standard documents this style requirement, we should do this for every NIT code review note reviewers have to offer.

A better use of a comment here is not to provide a human-readable description of what the code is doing but to instead provide a readable example of the output of the code. IMO.

src/BinSkim.Rules/RulesExtensionMethods.cs
```diff
@@ -108,6 +132,35 @@
             return sb.ToString();
         }
 
+        public static (string nameOnly, string comment) ExtractNameAndComment(string nameWithComment, bool includeCommentMarkers = false)
```
michaelcfanning, line 135: >ExtractNameAndComment [](http://example.com/codeflow?start=56&length=21)

This complexity looks non-obvious to me.

i.e., i'm not sure this addition to the design puts us in a cleaner/more capable condition but rather appears to make some relatively hard to understand code a little harder to understand.

src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs
```diff
@@ -104,8 +107,11 @@ public void AnalyzeNativeBinaryAndPdb(BinaryAnalyzerContext context)
                 return;
             }
 
-            var compilandsBinaryWithOneOrMoreInsecureFileHashes = new List<ObjectModuleDetails>();
             var compilandsLibraryWithOneOrMoreInsecureFileHashes = new List<ObjectModuleDetails>();
+            var compilandsLibraryWithOneOrMoreNoFileHashes = new List<ObjectModuleDetails>();
```
michaelcfanning, line 111: >compilandsLibraryWithOneOrMoreNoFileHashes [](http://example.com/codeflow?start=16&length=42)

wouldnt' a name like 'compilandsLibraryWithOneOrMoreMissingFileHashes' be a little clear?

Rather than a plurality of NoFileHashes?

src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs
```diff
@@ -194,75 +200,153 @@ public void AnalyzeNativeBinaryAndPdb(BinaryAnalyzerContext context)
...
+                            }
+                            else
+                            {
+                                compilandsLibraryWithOneOrMoreNoFileHashes.Add(objectModuleDetailsWithHashComment);
+                            }
                         }
                         else
                         {
-                            compilandsBinaryWithOneOrMoreInsecureFileHashes.Add(omDetails);
+                            if (sf.HashType != HashType.None)
```
michaelcfanning, line 220: >HashType [](http://example.com/codeflow?start=35&length=8)

You did not follow my guidance to organize this information by a dictionary keyed by HashType.

Let's hold on this one. I will take a closer look tomorrow to see if I can better understand why you chose the direction you did. If possible, I'd prefer a simpler solution with less new complexity.

I note that you didn't add any new unit tests to the tool, despite adding a new rules extension method.

michaelcfanning:
Pipeline tests aren't passing for this change.

ReleaseHistory.md
schlaman-ms: Reordered the notes in this and the section below

PR #927, opened 2023-06-21
BA2024.EnableSpectreMitigations - If compiler command line missing then new warning message output saying so.

schlaman-ms:
## Without this enhancement, if a compiler command line is missing, then the user would just get a message that /Qspectre was not enabled (even though they probably did enable it). This completely misleads the consumer.
  - Now the user will see that the command line is missing and the message will say it is unable to determine whether /Qspectre is enabled.

### If it is missing then there is no point in checking for /Qspectre flag being set. And this will then create a new BA2024 warning to the user of the tool where they can see that the command line is missing for the given binary and it is not an issue of /Qspectre not being specified but just that you can't tell. 
 - This was a support case where the user was misled into wondering why they had a /Qspectre not specified error when they in fact did specify it. They just didn't have /Z7 turned on to include the command line info. Now you will see that the command line info is missing and the user will know how to proceed.


src/BinSkim.Rules/PERules/BA2024.EnableSpectreMitigations.cs
```diff
@@ -241,6 +246,14 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
                     // mitigations. We do not report here. BA2006 will fire instead.
                     continue;
                 }
+
+                // If the Command Line is not embedded in the binary, then just add a warning for this condition, and there is no point in checking switch states
```
michaelcfanning, line 250: >Command Line [](http://example.com/codeflow?start=26&length=12)

what is this strange casing? if this is english text, the casing is wrong. if you referring to a property name like 'RawCommandLine' it should be enquoted and match what you are referring to.

michaelcfanning, line 250: >states [](http://example.com/codeflow?start=155&length=6)

you do not have a period at the end of your comment.
your line of code is too long, please break your comment into two lines and columnize it.

ReleaseHistory.md
```diff
@@ -16,6 +16,7 @@
 - NEW => new feature 
 
 ## UNRELEASED
+* NEW: `BA2024.EnableSpectreMitigations` Inform user when compiler command line is missing and inform that unable to determine if `/Qspectre` is specified.  [#927](https://github.com/microsoft/binskim/pull/927)
 * NR : `BA2029.EnableIntegrityCheck` ([Rule Request](https://github.com/microsoft/binskim/issues/909)) [#922](https://github.com/microsoft/binskim/pull/922)
```
michaelcfanning, line 20: Try this. It now reads as a complete sentence. I removed some redundant language. I refer precisely to the property name that comprises the missing data.

* NEW: `BA2024.EnableSpectreMitigations` now informs user when a compiland `RawCommandLine` value is missing and the rule is therefore not able to determine if `/Qspectre` is specified.  [#927](https://github.com/microsoft/binskim/pull/927)

src/BinSkim.Rules/PERules/BA2024.EnableSpectreMitigations.cs
```diff
@@ -308,10 +321,28 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
             string line;
             var sb = new StringBuilder();
 
+            bool mitigationMissingCommandLineUnknownOnly = false;
+
+            if (mitigationUnknownNoCommandLineRaw.Count > 0)
+            {
+                mitigationMissingCommandLineUnknownOnly = true;
+
+                // The following modules were compiled with a toolset that supports
```
michaelcfanning, line 330: >following [](http://example.com/codeflow?start=23&length=9)

We have not explained the likely cause of this situation, e.g., that the code was linked to a static library with no debug information.

src/BinSkim.Rules/PERules/BA2024.EnableSpectreMitigations.cs
```diff
@@ -308,10 +321,28 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
...
+                sb.AppendLine(line);
+            }
+
             if (mitigationExplicitlyDisabledModules.Count > 0)
             {
-                // The following modules were compiled with Spectre
-                // mitigations explicitly disabled: {0}
+                mitigationMissingCommandLineUnknownOnly = false;
+
+                // The following modules were compiled with a toolset that supports
```
michaelcfanning, line 343: >following [](http://example.com/codeflow?start=23&length=9)

Is this *really* the string we report? Because if so, it needs improvement, let me know I will go edit.

michaelcfanning (changes requested):
:clock1:

src/BinSkim.Rules/PERules/BA2024.EnableSpectreMitigations.cs
schlaman-ms: These were the existing messages.
This shows up as new because there were no comments above the messages so I put them there. That's true for multiple of these.

src/BinSkim.Rules/PERules/BA2024.EnableSpectreMitigations.cs
schlaman-ms: I added the explanation of the compiland command line missing and the likely cause being about the static library...

ReleaseHistory.md
schlaman-ms: I have updated as you stated above and yes that is better. Thank you :)

src/BinSkim.Rules/PERules/BA2024.EnableSpectreMitigations.cs
schlaman-ms: The line is broken and the period is at the end now.

src/BinSkim.Rules/PERules/BA2024.EnableSpectreMitigations.cs
```diff
@@ -308,10 +322,28 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
...
+                // /Qspectre but no compiler command line is present and therefore
+                // it cannot be determined if /Qspectre was specified: {0}
                 line = string.Format(
                         RuleResources.BA2024_Warning_SpectreMitigationExplicitlyDisabled,
```
shaopeng-gh, line 348: >BA2024_Warning_SpectreMitigationExplicitlyDisabled [](http://example.com/codeflow?start=38&length=50)

the comment here "but no compiler command line is present and therefore it cannot be determined if /Qspectre was specified"
does not match the code:
`BA2024_Warning_SpectreMitigationExplicitlyDisabled`

src/BinSkim.Rules/PERules/BA2024.EnableSpectreMitigations.cs
schlaman-ms: Normal casing now.

shaopeng-gh (approved):
:shipit:

src/BinSkim.Rules/PERules/BA2024.EnableSpectreMitigations.cs
schlaman-ms: Nice catch.
Fixed.

PR #933, opened 2023-06-28
BA2024.EnableSpectreMitigations, Handle case where `CommandLineRaw` value is missing

schlaman-ms:
## Without this enhancement, if a compiler command line is missing, then the user would just get a message that /Qspectre was not enabled (even though they probably did enable it). This completely misleads the consumer.
  - Now the user will see that the command line is missing and the message will say it is unable to determine whether /Qspectre is enabled.

### If it is missing then there is no point in checking for /Qspectre flag being set. And this will then create a new BA2024 warning to the user of the tool where they can see that the command line is missing for the given binary and it is not an issue of /Qspectre not being specified but just that you can't tell. 
 - This was a support case where the user was misled into wondering why they had a /Qspectre not specified error when they in fact did specify it. They just didn't have /Z7 turned on to include the command line info. Now you will see that the command line info is missing and the user will know how to proceed.


src/BinSkim.Rules/PERules/BA2024.EnableSpectreMitigations.cs
```diff
@@ -308,10 +322,25 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
...
+                sb.AppendLine(line);
+            }
+
             if (mitigationExplicitlyDisabledModules.Count > 0)
             {
-                // The following modules were compiled with Spectre
-                // mitigations explicitly disabled: {0}
+                IsOnlyMissingCommandLineWarningPresent = false;
+
+                // The following modules were compiled with Spectre mitigations explicitly disabled: {0}
```
schlaman-ms, line 343: @michaelcfanning
Do you have edit suggestions for this and the other messages?

docs/BinSkimRules.md
```diff
@@ -817,6 +817,14 @@ Application code should be compiled with the Spectre mitigations switch (/Qspect
...
+
+{0}' was compiled with one or more modules with a toolset that supports /Qspectre but a compiland `RawCommandLine` value is missing and the rule is therefore not able to determine if `/Qspectre` is specified. The likely cause is that the code was linked to a static library with no debug information.  It is not known whether code generation mitigations for speculative execution side-channel attack (Spectre) vulnerabilities was enabled. Spectre attacks can compromise hardware-based isolation, allowing non-privileged users to retrieve potentially sensitive data from the CPU cache. To resolve the issue, ensure that the compiler command line is present (provide the /Z7 switch) and provide the /Qspectre switch on the compiler command-line (or specify <SpectreMitigation>Spectre</SpectreMitigation> in build properties), or pass /d2guardspecload in cases where your compiler supports this switch and it is not possible to update to a toolset that supports /Qspectre. This warning should be addressed for code that operates on data that crosses a trust boundary and that can affect execution, such as parsing untrusted file inputs or processing query strings of a web request.
```
shaopeng-gh, line 822: >/Z7 [](http://example.com/codeflow?start=669&length=3)

question about the language here `/Z7`.

/Z7, /Zi, /ZI :
[https://learn.microsoft.com/en-us/cpp/build/reference/z7-zi-zi-debug-information-format?view=msvc-170](https://learn.microsoft.com/en-us/cpp/build/reference/z7-zi-zi-debug-information-format?view=msvc-170)

shaopeng-gh (approved):
:shipit:

michaelcfanning (approved):

PR #935, opened 2023-06-29
Fix BA2004 bug creating error with empty list

shaopeng-gh:
when all error are caused by no hash, the current code will also generate an error with no item in the list.
generate an error will cause pipeline to fail, so we should fix it.

### DIFF
--- a/src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs
+++ b/src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs
@@ -252,17 +252,20 @@ namespace Microsoft.CodeAnalysis.IL.Rules
                 compilandsWithOneOrMoreInsecureFileHashes.Remove(HashType.None);
             }
 
-            string[] messages = new string[compilandsWithOneOrMoreInsecureFileHashes.Count];
-
-            int hashTypeCount = 0;
-            foreach (HashType hashType in compilandsWithOneOrMoreInsecureFileHashes.Keys)
+            if (compilandsWithOneOrMoreInsecureFileHashes.Count > 0)
             {
-                objectModuleDetails = compilandsWithOneOrMoreInsecureFileHashes[hashType];
-                messages[hashTypeCount++] = objectModuleDetails.CreateOutputCoalescedByCompiler(hashType.ToString());
-            }
+                string[] messages = new string[compilandsWithOneOrMoreInsecureFileHashes.Count];
 
-            message = string.Join(Environment.NewLine, messages);
-            GenerateCompilandsAndLog(context, message, failureLevel);
+                int hashTypeCount = 0;
+                foreach (HashType hashType in compilandsWithOneOrMoreInsecureFileHashes.Keys)
+                {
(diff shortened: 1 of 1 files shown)


src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs
```diff
@@ -252,17 +252,20 @@ private void GenerateCompilandsAndLog(BinaryAnalyzerContext context, Dictionary<
                 compilandsWithOneOrMoreInsecureFileHashes.Remove(HashType.None);
             }
 
-            string[] messages = new string[compilandsWithOneOrMoreInsecureFileHashes.Count];
-
-            int hashTypeCount = 0;
-            foreach (HashType hashType in compilandsWithOneOrMoreInsecureFileHashes.Keys)
+            if (compilandsWithOneOrMoreInsecureFileHashes.Count > 0)
```
shaopeng-gh, line 255: >> 0 [](http://example.com/codeflow?start=64&length=3)

when all error are caused by no hash, the current code will also generate an error with no item in the list.
generate an error will cause pipeline to fail, so we should fix it.
No test was added because the fix is clear, let me know I can add tests.

suvamM: @shaopeng-gh The problem is unclear. Could you add an example to the PR description please?

src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs
```diff
@@ -252,17 +252,20 @@ private void GenerateCompilandsAndLog(BinaryAnalyzerContext context, Dictionary<
...
             {
-                objectModuleDetails = compilandsWithOneOrMoreInsecureFileHashes[hashType];
-                messages[hashTypeCount++] = objectModuleDetails.CreateOutputCoalescedByCompiler(hashType.ToString());
-            }
+                string[] messages = new string[compilandsWithOneOrMoreInsecureFileHashes.Count];
```
suvamM, line 257: @shaopeng-gh A unit test here would be useful. Also helps drive home what the PR is trying to achieve.

suvamM (changes requested):

src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs
shaopeng-gh: agree, will work on test.

src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs
shaopeng-gh: will work on test and update the PR description.

schlaman-ms, line 255: Ok, this looks like exactly what it needs.

schlaman-ms (approved):
:shipit:

suvamM (approved):

PR #946, opened 2023-07-27
* BUG: Fix `--quiet`, `--recurse`, `--rich-return-code`, `--ignorePdbLoadError` and `--environment` not working without explicitly adding `true`.

shaopeng-gh:
Issue found by Michael.
`--recurse` not working without explicitly adding true.
SPM already had similar fix, we just re-use it.


michaelcfanning (approved):
:shipit:

PR #964, opened 2023-09-29
[Editorial] Fix render of angle brackets in binskim rules doc

ThomsonTan:
Angle brackets are rendered as empty in markdown which makes the xml element like `<ChecksumAlgorithm>` invisible. One option is replacing '<' by "&lt;". Another option used in this PR is wrapping the xml element in code marker. The latter approach makes the element visible in the raw text format as well.


michaelcfanning:
Thanks for this!

PR #965, opened 2023-10-11
`BA2004.EnableSecureSourceCodeHashing` now will no longer generate false positives on precompiled headers

shaopeng-gh:
1. `BA2004.EnableSecureSourceCodeHashing` now will no longer generate false positives on precompiled headers, because they are always without hash. 
2. The old code has a bug as well, when `pchFileName != string.Empty` is true, the current file is always `continue;` skipped. This is wrong, we are trying to ignore .pch file name mismatch, but effectively we didn't even check if the current file is a .pch at all. This lead to e.g. a file name `c:\program files (x86)\reference assemblies\microsoft\framework\.netframework\v4.5.2\system.dll` is skipped when pchFileName is `stdafx.pch`.


src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/MixedMode_x64_VS2015_Default.exe.sarif
```diff
@@ -76,7 +76,7 @@
             "id": "Warning_NativeWithInsecureStaticLibraryCompilands",
             "arguments": [
               "MixedMode_x64_VS2015_Default.exe",
-              "Microsoft (R) Optimizing Compiler : cxx : 19.0.24215.1 : [directly linked] [No hash value present] (.NETFramework,Version=v4.5.2.AssemblyAttributes.obj)\r\n"
+              "Microsoft (R) Optimizing Compiler : cxx : 19.0.24215.1 : [directly linked] [No hash value present] (.NETFramework,Version=v4.5.2.AssemblyAttributes.obj,stdafx.obj)\r\n"
```
shaopeng-gh, line 79: >stdafx.obj [](http://example.com/codeflow?start=167&length=10)

This is due to the 2nd in the PR description.
the object is `stdafx.obj`
it has PCH `stdafx.pch`
the source file without hash in it is `c:\program files (x86)\reference assemblies\microsoft\framework\.netframework\v4.5.2\system.dll`
not due to `stdafx.pch`.

michaelcfanning (approved):
:shipit:

ReleaseHistory.md
```diff
@@ -16,6 +16,7 @@
 - NEW => new feature 
 
 ## UNRELEASED
```
michaelcfanning, line 18: >UNRELEASED [](http://example.com/codeflow?start=3&length=10)

can we go ahead and burn a 4.2.1 release next? Seems like this issue is blocking lots of people.

michaelcfanning (approved):

ReleaseHistory.md
shaopeng-gh: @suvamM if sounds good to you I can start this.

shaopeng-gh: Release is finished.

PR #969, opened 2023-11-01
BUG: Fix missing error code to error description mappings for `CryptoError`

shaopeng-gh:
BUG: Fix `ERR998.ExceptionInAnalyze`: `InvalidOperationException: Unrecognized crypto HRESULT: 0x80096011` for check `BA2022.SignSecurely` when the signature is malformed, by adding missing error code to error description mappings.

We have a mapping that is from the error code to description, `RulesExtensionMethods.BuildCryptoErrorDescriptions()`.
Whenever `CryptoError` is refreshed,
this mapping must be updated accordingly.


src/BinSkim.Rules/RulesExtensionMethods.cs
```diff
@@ -130,10 +130,11 @@ private static string CreateLibraryKey(ObjectModuleDetails objectModuleDetail)
 
         private static readonly Dictionary<CryptoError, string> s_cryptoErrorToDescriptionMap = BuildCryptoErrorDescriptions();
 
-        private static Dictionary<CryptoError, string> BuildCryptoErrorDescriptions()
+        public static Dictionary<CryptoError, string> BuildCryptoErrorDescriptions()
```
shaopeng-gh, line 133: >public [](http://example.com/codeflow?start=8&length=6)

`CryptoError` is public,
this is descriptions for `CryptoError`, I think it is ok to have it as public.

michaelcfanning (approved):
:shipit:

src/BinSkim.Rules/RulesExtensionMethods.cs
scottoneil-ms: I don't know how nitty we are about class visibility modifiers, but I think making it internal and using 'InternalsVisibleTo' to specify that the test assembly can access would be ideal.  If code signing is used, this requires sticking the public key in a class attribute.  If there's no code signing then it's trivial.

My take: if you haven't taken a crack at this yet, I'd encourage giving it a shot.  But not a hill to die on if you run into a hard time figuring out the signing stuff.

https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=net-7.0

src/BinSkim.Rules/CryptoErrors.cs
```diff
@@ -10,6 +10,8 @@ namespace Microsoft.CodeAnalysis.IL.Rules
     /// </summary>
     public enum CryptoError : uint
     {
+        // Whenever this file is refreshed,
```
scottoneil-ms, line 13: Nit: suggest "modified" instead of "refreshed."

src/Test.UnitTests.BinSkim.Rules/CryptoErrorEnumTests.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class CryptoErrorEnumTests
+    {
+        [Fact]
+        public void CryptoErrorEnumTest_NoMissingValues()
+        {
+            IEnumerable<CryptoError> enumValues = Enum.GetValues(typeof(CryptoError)).Cast<CryptoError>();
+            Dictionary<CryptoError, string> dictionaryValues = RulesExtensionMethods.BuildCryptoErrorDescriptions();
+            var missingValues = enumValues.Where(value => !dictionaryValues.ContainsKey(value)).ToList();
+            missingValues.Should().HaveCount(0, "BuildCryptoErrorDescriptions() should contain all cases from CryptoError enum.");
+        }
```
scottoneil-ms, line 23: Nice test.

src/BinSkim.Rules/RulesExtensionMethods.cs
```diff
@@ -153,15 +154,64 @@ private static string CreateLibraryKey(ObjectModuleDetails objectModuleDetail)
                 { CryptoError.CERT_E_UNTRUSTEDTESTROOT, "The certification path terminates with the test root which is not trusted with the current policy settings." },
                 { CryptoError.CERT_E_VALIDITYPERIODNESTING, "The validity periods of the certification chain do not nest correctly." },
                 { CryptoError.CERT_E_WRONG_USAGE, "The certificate is not valid for the requested usage." },
+                { CryptoError.CERTSRV_E_ADMIN_DENIED_REQUEST, "The request was denied by a certificate manager or CA administrator." },
+                { CryptoError.CERTSRV_E_ALIGNMENT_FAULT, "A memory reference caused a data alignment fault." },
+                { CryptoError.CERTSRV_E_ARCHIVED_KEY_REQUIRED, "The request is missing a required private key for archival by the server." },
+                { CryptoError.CERTSRV_E_ARCHIVED_KEY_UNEXPECTED, "The request includes a private key for archival by the server, but key archival is not enabled for the specified certificate template." },
```
scottoneil-ms, line 160: "for archiv**ing** by the server" scans a little better for me, but none of my public school English teachers were as excited about teaching grammar as discussing literature, so I couldn't explain exactly why. I might be in error. :)

src/BinSkim.Rules/RulesExtensionMethods.cs
```diff
@@ -153,15 +154,64 @@ private static string CreateLibraryKey(ObjectModuleDetails objectModuleDetail)
...
+                { CryptoError.CERTSRV_E_SIGNATURE_POLICY_REQUIRED, "The request is missing required signature policy information." },
+                { CryptoError.CERTSRV_E_SIGNATURE_REJECTED, "One or more signatures did not include the required application or issuance policies. The request is missing one or more required valid signatures." },
+                { CryptoError.CERTSRV_E_SMIME_REQUIRED, "The request is missing a required SMIME capabilities extension." },
+                { CryptoError.CERTSRV_E_SUBJECT_ALT_NAME_REQUIRED, "The request is missing a required Subject Alternate name extension." },
+                { CryptoError.CERTSRV_E_SUBJECT_DIRECTORY_GUID_REQUIRED, "The Active Directory GUID is unavailable and cannot be added to the Subject Alternate name." },
+                { CryptoError.CERTSRV_E_SUBJECT_DNS_REQUIRED, "The DNS name is unavailable and cannot be added to the Subject Alternate name." },
+                { CryptoError.CERTSRV_E_SUBJECT_EMAIL_REQUIRED, "The EMail name is unavailable and cannot be added to the Subject or Subject Alternate name." },
```
scottoneil-ms, line 206: Suggest: "E**-**Mail"

src/BinSkim.Rules/CryptoErrors.cs
shaopeng-gh: fixed

src/BinSkim.Rules/RulesExtensionMethods.cs
shaopeng-gh: Fixed, the text was from Windows SDK, both sounds good to me.

src/BinSkim.Rules/RulesExtensionMethods.cs
shaopeng-gh: fixed

src/BinSkim.Rules/RulesExtensionMethods.cs
shaopeng-gh: fixed

src/BinSkim.Rules/RulesExtensionMethods.cs
scottoneil-ms: Oh man, if I'd known you were using strings that were already public-facing I wouldn't have said anything.  Thanks though!

scottoneil-ms (approved):

shaopeng-gh:
@michaelcfanning I think ready to merge.

Issue #986, opened 2024-02-23
--ignorePdbLoadError behavior changed


shaopeng-gh:
if target PE missing pdb,
In 1.9.5 when the command line does not contain --ignorePdbLoadError the tool completes successfully.
In latest version when the command line does not contain --ignorePdbLoadError the tool exits with “Unexpected fatal runtime condition(s) observed: ExceptionLoadingPdb.

PR #988, opened 2024-02-29
Fix `BA2027.EnableSourceLink` cause `ExceptionLoadingPdb`

shaopeng-gh:
Fix `BA2027.EnableSourceLink` unexpectedly cause `ExceptionLoadingPdb` error when the PDB file is missing.

Please see details in https://github.com/microsoft/binskim/issues/986

The cause is this PR [[WIP] Add BA2027.EnableSourceLink](https://github.com/microsoft/binskim/pull/657)
missed setting `LogPdbLoadException` => false. (Pattern found from other rules)
I used the get history version to locate this exact PR changed the behavior.

Fix: add it, and add tests to prevent future new rules missing the same property set.

Additionally, I have question about other existing rules that have this `LogPdbLoadException` as true.
Are they bugs? What condition we want this true / false? This is a question about original design when we introduce this flag. 
Let me know I can create a bug if need fixes.


ReleaseHistory.md
```diff
@@ -18,6 +18,7 @@
 * DEP: Update `Sarif.Sdk` submodule from [bc8cb57 to fd6e615](https://github.com/microsoft/sarif-sdk/compare/bc8cb57...fd6e615). Reference [SARIF SDK Release History](https://github.com/microsoft/sarif-sdk/blob/fd6e615/ReleaseHistory.md).
 * NEW: Add `--disable-telemetry` argument to disable telemetry collection.
 * BUG: Fix `ERR998.ExceptionInAnalyze`: `InvalidOperationException: Unrecognized crypto HRESULT: 0x80096011` for check `BA2022.SignSecurely` when the signature is malformed, by adding missing error code to error description mappings. [969](https://github.com/microsoft/binskim/pull/969)
+* BUG: Fix `BA2027.EnableSourceLink` unexpectedly cause `ExceptionLoadingPdb` error when the PDB file is missing. [988](https://github.com/microsoft/binskim/pull/988)
```
stacywray, line 21: - Change "cause" to "causes"
- Add a period to the end of the line after the 988 link. The BUG above is also missing a period at the end after 969.

stacywray (changes requested):
Requesting a few minor changes:
- Change "cause" to "causes".
- Add period at end of line. The BUG immediately above also needs a period at the end of the line.

stacywray (approved):
Looks good.

src/Test.UnitTests.BinSkim.Rules/RulePropertyTests.cs
```diff
@@ -0,0 +1,66 @@
...
+
+using FluentAssertions;
+
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.IL.Rules
+{
+    public class RulePropertyTests
+    {
+        private static readonly string[] ExpectedLogPdbLoadExceptionRules = new string[]
```
shaopeng-gh, line 18: Please see PR desc.

src/Test.UnitTests.BinaryParsers/TestData/PE/Determinism/E_PDB_NOT_FOUND/Managed_x64_VS2022_CSharp_Net70_Default.dll
shaopeng-gh, line 1: The old test was based on BA2027 throw that error, now that is fixed, so need to change to another file triggered by other rules.

PR #994, opened 2024-05-24
Adding VC project properties info to 'how to resolve' suggestions.

olgaark:
Adding VC project properties info to suggestions
Making long strings multiline for easier reading.
Adding more info for BA2004 and clarifying some suggestions for compiler defaults.

### DIFF
--- a/src/BinSkim.Rules/RuleResources.resx
+++ b/src/BinSkim.Rules/RuleResources.resx
@@ -338,7 +338,7 @@ To resolve this issue, don't set /HIGHENTROPYVA:NO on link.exe command line and 
   </data>
   <data name="BA2015_Error_NoLargeAddressAware" xml:space="preserve">
     <value>'{0}' does not declare itself as high entropy ASLR compatible. High entropy makes Address Space Layout Randomization more effective in mitigating memory corruption vulnerabilities.
-To resolve this issue, don't set /LARGEADDRESSAWARE:NO on link.exe command line. (This image was determined to have been properly compiled as /HIGHENTROPYVA.)</value>
+To resolve this issue, don't set /LARGEADDRESSAWARE:NO on link.exe command line and allow it to be enabled by default. (This image was determined to have been properly compiled as /HIGHENTROPYVA.)</value>
   </data>
   <data name="BA2015_Pass" xml:space="preserve">
     <value>'{0}' is high entropy ASLR compatible, reducing an attacker's ability to exploit code in well-known locations.</value>
--- a/src/BinSkim.Rules/RuleResources.resx
+++ b/src/BinSkim.Rules/RuleResources.resx
@@ -118,10 +118,17 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="BA2001_Error" xml:space="preserve">
-    <value>'{0}' is a 64-bit image with a preferred base address below the 4GB boundary. Having a preferred base address below this boundary triggers a compatibility mode in Address Space Layout Randomization (ASLR) on recent versions of Windows that reduces the number of locations to which ASLR may relocate the binary. This reduces the effectiveness of ASLR at mitigating memory corruption vulnerabilities. To resolve this issue, either use the default preferred base address by removing any uses of /baseaddress from compiler command lines, or /BASE from linker command lines (recommended), or configure your program to start at a base address above 4GB when compiled for 64 bit platforms (by changing the constant passed to /baseaddress or /BASE). Note that if you choose to continue using a custom preferred base address, you will need to make this modification only for 64-bit builds, as base addresses above 4GB are not valid for 32-bit binaries.</value>
+    <value>'{0}' is a 64-bit image with a preferred base address below the 4GB boundary. 
+Having a preferred base address below this boundary triggers a compatibility mode in Address Space Layout Randomization (ASLR) on recent versions of Windows that reduces the number of locations to which ASLR may relocate the binary. This reduces the effectiveness of ASLR at mitigating memory corruption vulnerabilities.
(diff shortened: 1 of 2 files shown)


src/BinSkim.Rules/RuleResources.resx
```diff
@@ -266,43 +310,63 @@ Modules triggering this check were:
...
+To resolve this issue, don't set /HIGHENTROPYVA:NO on link.exe command line and allow it to be enabled by default. (This image was determined to have been properly compiled as /LARGEADDRESSAWARE.)</value>
   </data>
   <data name="BA2015_Error_NoLargeAddressAware" xml:space="preserve">
-    <value>'{0}' does not declare itself as high entropy ASLR compatible. High entropy makes Address Space Layout Randomization more effective in mitigating memory corruption vulnerabilities. To resolve this issue, configure your tools to mark the program high entropy compatible by supplying /LARGEADDRESSAWARE to the C or C++ linker command line. (This image was determined to have been properly compiled as /HIGHENTROPYVA.)</value>
+    <value>'{0}' does not declare itself as high entropy ASLR compatible. High entropy makes Address Space Layout Randomization more effective in mitigating memory corruption vulnerabilities.
+To resolve this issue, don't set /LARGEADDRESSAWARE:NO on link.exe command line. (This image was determined to have been properly compiled as /HIGHENTROPYVA.)</value>
```
shaopeng-gh, line 341: >. [](http://example.com/codeflow?start=79&length=1)

should we use same style as above with `and allow it to be enabled by default.`

src/BinSkim.Rules/RuleResources.resx
```diff
@@ -266,43 +310,63 @@ Modules triggering this check were:
...
   </data>
   <data name="BA2018_EnableSafeSEH_Description" xml:space="preserve">
-    <value>X86 binaries should enable the SafeSEH mitigation to minimize exploitable memory corruption issues. SafeSEH makes it more difficult to exploit vulnerabilities that permit overwriting SEH control blocks on the stack, by verifying that the location to which a thrown SEH exception would jump is indeed defined as an exception handler in the source program (and not shellcode). To resolve this issue, supply the /SafeSEH flag on the linker command line. Note that you will need to configure your build system to supply this flag for x86 builds only, as the /SafeSEH flag is invalid when linking for ARM and x64.</value>
+    <value>X86 binaries should enable the SafeSEH mitigation to minimize exploitable memory corruption issues. SafeSEH makes it more difficult to exploit vulnerabilities that permit overwriting SEH control blocks on the stack, by verifying that the location to which a thrown SEH exception would jump is indeed defined as an exception handler in the source program (and not shellcode).
+To resolve this issue, supply the /SafeSEH flag on link.exe command line.
```
shaopeng-gh, line 361: >link.exe command line [](http://example.com/codeflow?start=51&length=21)

should we change all instance of `link.exe` in the PR to just `linker`.
For user that is not using Microsoft toolchain.

src/BinSkim.Rules/RuleResources.resx
```diff
@@ -266,43 +310,63 @@ Modules triggering this check were:
...
+    <value>'{0}' is an x86 binary which {1}, indicating that it does not enable the SafeSEH mitigation. SafeSEH makes it more difficult to exploit memory corruption vulnerabilities that can overwrite SEH control blocks on the stack, by verifying that the location to which a thrown SEH exception would jump is indeed defined as an exception handler in the source program (and not shellcode).
+To resolve this issue, supply the /SafeSEH flag on link.exe command line.
+For VC projects use ItemDefinitionGroup - Link = ImageHasSafeExceptionHandlers property with 'true' value.
```
shaopeng-gh, line 368: >VC projects [](http://example.com/codeflow?start=4&length=11)

For user that is not using Microsoft toolchain. Do we think 
`VC projects` is clear for them or we want a longer name more clear,
I am not sure, an open question.

shaopeng-gh:
Thanks for submitting this PR!

src/BinSkim.Rules/RuleResources.resx
olgaark: Most of the switches are tool specific, clang/gcc have their own, not /SafeSEH.

olgaark: /SafeSEH is only applicable to link.exe, clang/GCC have different switches.

src/BinSkim.Rules/RuleResources.resx
olgaark: What would you suggest? "VC" is short name for "Visual Studio C++" and if you search for "VC Project" you get
https://learn.microsoft.com/en-us/cpp/build/creating-and-managing-visual-cpp-projects?view=msvc-170. 
Alternatively, I can say "for vcxproj" - would it be better?

src/BinSkim.Rules/RuleResources.resx
shaopeng-gh: Yes. just a note not related to this PR, imo we have assumptions in a lot places the user is using Microsoft toolchain, while BinSkim target is just PE files.

src/BinSkim.Rules/RuleResources.resx
shaopeng-gh: Just tested "VC Project" with non-Microsoft search engine with Incognito, the fist 3 results all are something else other than the topic. The Microsoft search engine did return all related topic. 
But "VC Project" is clear as well.
+ @suvamM if any preference.

shaopeng-gh (approved):
:shipit:

src/BinSkim.Rules/RuleResources.resx
olgaark: How about I leave it as "VC project" for now? As soon as people can quickly tell if this info is applicable to them or not, we are good here. And we can always search/replace it with something else if we get feedback.

shaopeng-gh: sounds good to me

michaelcfanning:
Our pipeline won't run against a code fork for security reasons. I've added you as a contributor, would you mind recreating your PR as a direct branch off main? :)

PR #1001, opened 2024-07-26
Add conditional check for `TryGetPortablePdbMetadataReader` when PDB file is missing

lukaskohl-msft:
`File.OpenRead` can throw if `pdb.PdbLocation` is a directory.

This can happen if (for some reason) the pdb file is not found -- In which case `PdbLocation` will return the working directory, causing `File.OpenRead` to throw.

There's perhaps a better place to put this check, or a better way to handle this. Suggestions welcome :)

It would be great to have this scenario logged in some way, as there are `E_PDB_NOT_FOUND` or `E_PDB_DBG_NOT_FOUND` HRESULT codes.
For example:
C:\somePath\sni.dll : error ERR997.ExceptionLoadingPdb : 'sni.dll' was not evaluated because its PDB could not be loaded (E_PDB_NOT_FOUND).
However, I was unable to find a good way to integrate that into existing functionality, let me know if that's easily implementable.


ReleaseHistory.md
```diff
@@ -16,6 +16,7 @@
 - NEW => new feature 
 
 ## UNRELEASED
+* BUG: Fix `TryGetPortablePdbMetadataReader` unexpectedly causes `UnauthorizedAccessException` error when the PDB file is missing. [1001](https://github.com/microsoft/binskim/pull/1001).
```
michaelcfanning, line 19: >TryGetPortablePdbMetadataReader [](http://example.com/codeflow?start=12&length=31)

`due to evaluating to Pdb.PdbLocation being populated with the load directory (not an existing location).`

michaelcfanning, line 19: >TryGetPortablePdbMetadataReader [](http://example.com/codeflow?start=12&length=31)

I'd suggest creating a repro binary to check-in.

michaelcfanning (approved):
:shipit:

PR #1002, opened 2024-07-29
Using consoleLogger in Telemetry client within MultithreadedAnlyzeCom…

AllDwarf:
This pull request includes changes to the `MultithreadedAnalyzeCommand` class in `src/BinSkim.Driver` to enhance logging capabilities by adding a new console logger and ensuring proper namespaces are included.

Enhancements to logging:

* [`src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs`](diffhunk://#diff-d24327e39f7ce09f6e1bcd7cf14bffbe07c6341a35c9dc6c0647b0f6262900efR54-R64): Added a `ConsoleLogger` to the `AggregatingLogger` to enable console logging during analysis.

Namespace inclusion:

* [`src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs`](diffhunk://#diff-d24327e39f7ce09f6e1bcd7cf14bffbe07c6341a35c9dc6c0647b0f6262900efR16): Included the `Microsoft.CodeAnalysis.Sarif.Writers` namespace to support the new logging functionality.

### DIFF
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -8,6 +8,8 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 
+using CommandLine;
+
 using Microsoft.CodeAnalysis.BinaryParsers;
 using Microsoft.CodeAnalysis.IL.Rules;
 using Microsoft.CodeAnalysis.IL.Sdk;
@@ -52,13 +54,21 @@ namespace Microsoft.CodeAnalysis.IL
 
             if (this.Telemetry?.TelemetryClient != null)
             {
+                // Create an aggregating logger that will combine all loggers into a single logger.
                 var aggregatingLogger = new AggregatingLogger();
+                if (context.Logger is AggregatingLogger)
+                {
+                    aggregatingLogger = context.Logger as AggregatingLogger;
+                }
+                else
+                {
+                    aggregatingLogger.Loggers.Add(context.Logger);
+                }
 
                 var ruleTelemetryLogger = new RuleTelemetryLogger(this.Telemetry.TelemetryClient);
                 ruleTelemetryLogger.AnalysisStarted();
 
                 // Combine rule telemetry with any other loggers that may be present.
-                aggregatingLogger.Loggers.Add(context.Logger);
                 aggregatingLogger.Loggers.Add(ruleTelemetryLogger);
                 context.Logger = aggregatingLogger;
             }
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -57,6 +57,7 @@ namespace Microsoft.CodeAnalysis.IL
                 var ruleTelemetryLogger = new RuleTelemetryLogger(this.Telemetry.TelemetryClient);
                 ruleTelemetryLogger.AnalysisStarted();
 
+                // Combine rule telemetry with any other loggers that may be present.
                 aggregatingLogger.Loggers.Add(context.Logger);
                 aggregatingLogger.Loggers.Add(ruleTelemetryLogger);
                 context.Logger = aggregatingLogger;
--- a/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
+++ b/src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
@@ -48,6 +48,8 @@ namespace Microsoft.CodeAnalysis.IL
 
         public override BinaryAnalyzerContext InitializeGlobalContextFromOptions(AnalyzeOptions options, ref BinaryAnalyzerContext context)
         {
+            base.InitializeGlobalContextFromOptions(options, ref context);
+
             if (this.Telemetry?.TelemetryClient != null)
             {
                 var aggregatingLogger = new AggregatingLogger();
@@ -55,6 +57,7 @@ namespace Microsoft.CodeAnalysis.IL
                 var ruleTelemetryLogger = new RuleTelemetryLogger(this.Telemetry.TelemetryClient);
                 ruleTelemetryLogger.AnalysisStarted();
 
+                aggregatingLogger.Loggers.Add(context.Logger);
                 aggregatingLogger.Loggers.Add(ruleTelemetryLogger);
                 context.Logger = aggregatingLogger;
             }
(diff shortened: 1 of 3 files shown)


AllDwarf:
## Testing
Issue was tested with multiple setups
### Setup
Running `analyze` command against empty folder wihout targets inside like this:

```pwsh
    analyze --config default --ignorePdbLoadError --hashes --statistics --output C:\Users\marekaldorf\source\repos\software-development-tools-samples\src\MyRun.sarif C:\Users\marekaldorf\source\repos\AllDwarf\BlobStorageTrigger\bin\*.so --log ForceOverwrite
```

You can change the path to point to any empty folder and the sarif you want and reproduce it the same way I did.
```pwsh
    analyze --config default --ignorePdbLoadError --hashes --statistics --output {PATH_TO_SARIF}\{NAME}.sarif {PATH_TO_EMPTY_FOLDER}\*.so --log ForceOverwrite
```

1) `--disable-telemetry` set to `false` when env variable `BinskimAppInsightsConnectionString` set to point to AppInsights with connection string set.
    1) BEFORE
    image
    1) AFTER
    image

michaelcfanning:
>             base.InitializeGlobalContextFromOptions(options, ref context);

try moving this line to the beginning of the method, perhaps the default console logging init will work then.

---
Refers to: src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs:68 in c1af956. [](commit_id = c1af9564a7840f2316e28f74a8172d2f0ae0b5e4, deletion_comment = True)

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -48,13 +48,16 @@ private bool IsValidScanTarget(string file)
...
+            base.InitializeGlobalContextFromOptions(options, ref context);
+
             if (this.Telemetry?.TelemetryClient != null)
             {
                 var aggregatingLogger = new AggregatingLogger();
 
                 var ruleTelemetryLogger = new RuleTelemetryLogger(this.Telemetry.TelemetryClient);
                 ruleTelemetryLogger.AnalysisStarted();
 
+                aggregatingLogger.Loggers.Add(context.Logger);
```
michaelcfanning, line 60: >aggregatingLogger [](http://example.com/codeflow?start=16&length=17)

Run this test: configure your analysis to specify a SARIF output location. And then debug into this code. What I *think* will happen is that your context object will come back with an aggregating logger that has two things, a console logger and a SARIF logger.

And your change will create this

-> AggregatingLogger
  --> Telemetry Logger
  --> AggregatingLogger
    --> ConsoleLogger
    --> SARIF Logger

Now, I suppose there's no reason *not* to have this nesting, your approach ends up being pretty clean because of it.

Just wanted to make sure we're on the same page with what is happening at runtime and that this is a conscious decision.

AllDwarf: Yep seems to be more clean than the first attempt. I will investigate the behavior you describing and let's see if I can come up with something elegant.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
```diff
@@ -48,13 +48,16 @@ private bool IsValidScanTarget(string file)
 
         public override BinaryAnalyzerContext InitializeGlobalContextFromOptions(AnalyzeOptions options, ref BinaryAnalyzerContext context)
         {
+            base.InitializeGlobalContextFromOptions(options, ref context);
```
michaelcfanning, line 51: >InitializeGlobalContextFromOptions [](http://example.com/codeflow?start=17&length=34)

btw! now seems a good time to talk a bit about testing. one thing I observed in the previous change is that you might have broken support for the `--quiet` flag. So you might run a test: revert to your previous change and run all tests: hopefully we see that some test protecting the functionality of the quiet switch fired.

I was only peripherally involved with the telemetry client work, unfortunately, so I don't know the state of testing.

This method is very testable, however, you can create this object and call this public method on it. So I would suggest creating some tests that hit this code path.

Verification looks straightforward, populate the object with a non-null Telemetry object, then call this init method and ensure the `context.Loggers` property contains what you think that it should.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
michaelcfanning: Oh, right. The SARIF log itself is actually initialized in `InitializeOutputs(globalContext)` not in this core global init.

Makes sense! I like the way you approached this, looks perfect.

src/BinSkim.Driver/MultithreadedAnalyzeCommand.cs
AllDwarf: Yep good point. I tried it with `--quiet true` and I think seems to be working as expected no ConsoleLogger set, just telemetry is sent.
```pwsh
C:\Users\marekaldorf\source\repos\binskim\bld\bin\x64_Debug\netcoreapp3.1\BinSkim.exe (process 40088) exited with code 1.
To automatically close the console when debugging stops, enable Tools->Options->Debugging->Automatically close the console when debugging stops.
Press any key to close this window . . .
```
`context.Logger` in this case it is without ConsoleLogger. It has empty IAnalysisLogger. So in quiet mode this specific code will add one empty logger and one RuleTelemetryLogger together.

I tried to redo it a bit with something like this:
```csharp
            if (this.Telemetry?.TelemetryClient != null)
            {
                // Create an aggregating logger that will combine all loggers into a single logger.
                var aggregatingLogger = new AggregatingLogger();
                if (context.Logger is AggregatingLogger)
                {
                    aggregatingLogger = context.Logger as AggregatingLogger;
                }
                else
                {
                    aggregatingLogger.Loggers.Add(context.Logger);
                }

                var ruleTelemetryLogger = new RuleTelemetryLogger(this.Telemetry.TelemetryClient);
                ruleTelemetryLogger.AnalysisStarted();

                // Combine rule telemetry with any other loggers that may be present.
                aggregatingLogger.Loggers.Add(ruleTelemetryLogger);
                context.Logger = aggregatingLogger;
            }
```
In this case we'll achieve non empty IAnalyseLogger in the first place but it's a bit hacky I would say with the type checks and such WDYT?

AllDwarf: just FYI, I'm working now on the test hopefully tomorrow I'll make them working :)

AllDwarf: @michaelcfanning just added the tests

src/Test.UnitTests.BinSkim.Driver/MultithreadedAnalyzeCommandTests.cs
```diff
@@ -150,5 +155,89 @@ private static string[] Shuffle(string[] data)
 
             return data;
         }
+
+        [Fact]
+        public void MultithreadedAnalyzeCommand_InitializeGlobalContextFromOptions_TelemetryNotSpecifiedHasOneConsoleLogger()
```
michaelcfanning, line 160: nice tests!

michaelcfanning (approved):
:shipit:

PR #1004, opened 2024-08-07
Add conditional check for `TryGetPortablePdbMetadataReader` when PDB file is missing

lukaskohl-msft:
`File.OpenRead` can throw if `pdb.PdbLocation` is a directory.

This can happen if (for some reason) the pdb file is not found -- In which case `PdbLocation` will return the working directory, causing `File.OpenRead` to throw.


AllDwarf (approved):
lgtm

lukaskohl-msft:
@microsoft-github-policy-service agree

PR #1014, opened 2024-10-08
Wrap XML tags in Markdown code blocks

nil4:
Make the values visible in the rendered Markdown, not just in source code.

e.g. going from this:

> or populate the '<ChecksumAlgorithm>' project property with 'SHA256' to enable secure source code hashing

to this:

> or populate the '`<ChecksumAlgorithm>`' project property with 'SHA256' to enable secure source code hashing


mkacmar:
Thanks for your contribution, unfortunately I don't think we can merge it for a few reasons - the Markdown file with documentation is [generated ](https://github.com/microsoft/sarif-sdk/blob/main/src/Sarif.Driver/Sdk/ExportRulesMetadataCommandBase.cs) (see here for [source data](https://github.com/microsoft/binskim/blob/main/src/BinSkim.Rules/RuleResources.resx)), additionally these messages are also used when logging to stdout where Markdown styling doesn't fit very well.

PR #1015, opened 2024-10-09
Update CODEOWNERS

AllDwarf:
Update CODEOWNERS to include new team 1es-sdl-engineering as a default code owner

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,4 +2,4 @@
 # This file controls who is tagged for review for any given pull request.
 
 # For anything not explicitly taken by someone else:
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -2,4 +2,4 @@
 # This file controls who is tagged for review for any given pull request.
 
 # For anything not explicitly taken by someone else:
-* @eddynaka @michaelcfanning @marmegh
+*       @microsoft/1es-sdl-engineering @AllDwarf @martin-reznik @mkacmar
```
mkacmar, line 5: Would you mind referencing just the team instead of specific usernames? I think our future selves will thank us if we ever need to deal with offboarding.

mkacmar (approved):

PR #1016, opened 2024-10-10
Cross-Platform .rsp File Support

lukaskohl-msft:
Unwrapping/expanding .rsp files used to be supported on Windows only due to FFI calling to `shell32.dll`. This PR removes FFI in favor of a code native solution.

the `sarif-sdk` submodule contains a reimplementation of `CommandLineToArgvW`. Rather than updating `sarif-sdk`, we've moved the relevant code into the BinSkim codebase, and pointed to the reimplementation instead to slightly lessen our reliance on `sarif-sdk`, while simultaneously increasing our supported platforms.

Alongside this change, tests were ported and altered over into the BinSkim codebase.

### DIFF
--- a/src/BinSkim.Driver/ExpandArguments.cs
+++ b/src/BinSkim.Driver/ExpandArguments.cs
@@ -20,13 +20,14 @@ namespace Microsoft.CodeAnalysis.IL
 
             foreach (string argument in args)
             {
-                if (!IsResponseFileArgument(argument))
+                string trimArgument = argument.Trim('"');
+                if (!IsResponseFileArgument(trimArgument))
                 {
-                    expandedArguments.Add(argument);
+                    expandedArguments.Add(trimArgument);
                     continue;
                 }
 
-                string responseFile = argument.Trim('"').Substring(1);
+                string responseFile = trimArgument.Substring(1);
 
                 responseFile = environmentVariables.ExpandEnvironmentVariables(responseFile);
                 responseFile = fileSystem.PathGetFullPath(responseFile);
--- a/src/BinSkim.Driver/ExpandArguments.cs
+++ b/src/BinSkim.Driver/ExpandArguments.cs
@@ -48,7 +48,16 @@ namespace Microsoft.CodeAnalysis.IL
         {
             foreach (string responseFileLine in responseFileLines)
             {
-                List<string> fileList = ArgumentSplitter.CommandLineToArgvW(responseFileLine.Trim()) ??
+                string responseFilePath = responseFileLine;
+
+                // Ignore comments from response file lines
+                int commentIndex = responseFileLine.IndexOf('#');
+                if (commentIndex >= 0)
+                {
+                    responseFilePath = responseFileLine.Substring(0, commentIndex);
+                }
+
+                List<string> fileList = ArgumentSplitter.CommandLineToArgvW(responseFilePath.Trim()) ??
                     throw new InvalidOperationException("Could not parse response file line:" + responseFileLine);
 
                 expandedArguments.AddRange(fileList);
--- a/src/BinSkim.Driver/ExpandArguments.cs
+++ b/src/BinSkim.Driver/ExpandArguments.cs
@@ -1,0 +1,57 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+
+using Microsoft.CodeAnalysis.Sarif;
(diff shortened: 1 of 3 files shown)


src/BinSkim.Driver/ExpandArguments.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            return expandedArguments.ToArray();
+        }
+
+        private static bool IsResponseFileArgument(string argument)
+        {
+            return argument.Length > 1 && argument[0] == '@';
+        }
+
+        private static void ExpandResponseFile(string[] responseFileLines, List<string> expandedArguments)
```
mkacmar, line 46: Does this handle `#` for comments ([ref](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-response-files?view=vs-2022))? - I don't know if it should or whether previous implementation did, but it's something to think about considering we'll also use this functionality for Windows.

src/BinSkim.Driver/ExpandArguments.cs
```diff
@@ -0,0 +1,57 @@
...
+                string[] responseFileLines = fileSystem.FileReadAllLines(responseFile);
+                ExpandResponseFile(responseFileLines, expandedArguments);
+            }
+
+            return expandedArguments.ToArray();
+        }
+
+        private static bool IsResponseFileArgument(string argument)
+        {
+            return argument.Length > 1 && argument[0] == '@';
```
mkacmar, line 43: Nitpick: You may want to extract `@` into some reasonably named constant.

mkacmar (commented):
I've left a few comments / questions.

src/BinSkim.Driver/ExpandArguments.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            foreach (string argument in args)
+            {
+                if (!IsResponseFileArgument(argument))
+                {
+                    expandedArguments.Add(argument);
+                    continue;
+                }
+
+                string responseFile = argument.Trim('"').Substring(1);
```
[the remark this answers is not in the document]
lukaskohl-msft: I see what you're saying, I took this from the original implementation. I think you're correct here, though this concerns only a situation when you're escaping characters (Who knows what the original reason was).

in windows CLI: `BinSkim.exe analyze "@example.rsp"` would treat the double quotes as a single parameter `@example.rsp` (quotations are needed because of PowerShell)
in Linux CLI: `./BinSkim analyze "@example.rsp"` or `./BinSkim analyze @example.rsp` are both valid.

It would be an issue if we escape the quotation marks: `./BinSkim analyze \"@example.rsp\"`. At that point it would break this implementation but `./BinSkim analyze @example.rsp\"\"\"\"` wouldn't.

I'll make this a tiny bit more robust, so that one can input an escaped quote.

src/BinSkim.Driver/ExpandArguments.cs
lukaskohl-msft: Sure!

src/BinSkim.Driver/ExpandArguments.cs
lukaskohl-msft: Great question! I don't believe the original implementation takes that into consideration, but I can see that being an issue. 

I've played with this a bit and yes, It would treat the comments as like a list of items to scan: `"analyze" "/some/path/" "#" "Some" "Random" "Comment"`.
But won't crash, or be valid towards the analysis. But this is something that should be addressed.

AllDwarf (approved):
Seems to be working on my end, Good Job! We should proceed once it got merged with E2E testing with Guardian on Linux.

src/BinSkim.Driver/ExpandArguments.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        private static void ExpandResponseFile(string[] responseFileLines, List<string> expandedArguments)
+        {
+            foreach (string responseFileLine in responseFileLines)
+            {
+                string responseFilePath = responseFileLine;
+
+                // Ignore comments from response file lines
+                int commentIndex = responseFileLine.IndexOf('#');
+                if (commentIndex >= 0)
```
mkacmar, line 55: Would you mind also adding a test for processing comments?

mkacmar (approved):

mkacmar:
Can you please rebase @lukaskohl-msft so that CodeQL passes before merge?

shoobu:
Hi @lukaskohl-msft, @mkacmar, @martin-reznik, 

Just trying to understand, is there anything blocking this from being deployed?

-Shubha

PR #1024, opened 2024-12-19
BinSkim .NET Updates to version 9

marekaldorf:
# [WIP] BinSkim .NET Updates to version 9
## Overview
This pull request focuses on updating the BinSkim project to incorporate the latest .NET updates. The goal is to ensure compatibility with the newest .NET features and improve overall performance and security.

## Changes
Updated the project to target the latest .NET version.

## Benefits
Enhanced Compatibility: Ensures that the BinSkim project is compatible with the latest .NET version, taking advantage of new features and improvements.

## Testing
Verified that the project builds successfully with the latest .NET version.
Ran all existing tests to ensure no regressions were introduced.
Conducted manual testing to confirm that the functionality remains intact.
Additional Notes
If you encounter any issues or have questions about these updates, please feel free to reach out.

PR #1037, opened 2025-01-22
Update BinSkim to use Net8

Penguinwizzard:
DotNet 6 left support back in November; we need to roll forward to continue getting security updates.


AllDwarf:
Thanks @Penguinwizzard for the contributions, we discussed it internally via DMs and there is an open PR from me which is focusing on .Net9 updates https://github.com/microsoft/binskim/pull/1024. Seems to me like the changes are the same, but also the mentioned PR includes cross platform changes. The thing is Project itself can be build, but in the testing phase you'll get stuck with cross platform changes.

Being that said, let me followup in the dotnet 9 PR, which should at the end serve same purpose.

PR #1041, opened 2025-02-11
Implement new ignore setting when PE file cannot be analyzed

AllDwarf:
### Issue Description: 
The error ERR998.ExceptionInCanAnalyze occurs when BinSkim attempts to determine if 'netfile.exe' is a valid analysis target for the `'EnableControlFlowGuard'` check. The exception is caused by an `InvalidOperationException` related to RVA not belonging to any section. We want to not break the build but proceed instead and control breaking based on the Sarif logs.

### Steps to Reproduce: 
The error can be reproduced by running BinSkim analysis for `'netfile.exe'` stored in TestData or similar files. Which cannot be easily read by PEReader.

### Proposed Fix: 
Update BinSkim to handle exceptions during the analysis process without breaking the build. This includes proceeding with the analysis even if a file cannot be read or analyzed properly. 

### Additional Context: 
The fix should apply to all occurrences of ERR998.ExceptionInCanAnalyze. The release of the fix is planned for Wednesday/Thursday next week.

### DIFF
--- a/src/BinaryParsers/BinaryParsersProperties.cs
+++ b/src/BinaryParsers/BinaryParsersProperties.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 
@@ -20,7 +19,8 @@ namespace Microsoft.CodeAnalysis.BinaryParsers
                 DisableTelemetry,
                 IncludeWixBinaries,
                 LocalSymbolDirectories,
-                SymbolPath
+                SymbolPath,
+                IgnorePELoadError
             }.ToImmutableArray();
         }
 
@@ -60,5 +60,10 @@ namespace Microsoft.CodeAnalysis.BinaryParsers
                 "environment variables at runtime. Use this argument instead for specifying the symbol path." +
                 "WARNING: Be sure to specify a local file cache in the symbol path if at all possible, in order " +
                 "to avoid the possibility of significance time-to-analyze performance degradataion.");
+
+        public static PerLanguageOption<bool> IgnorePELoadError { get; } =
+            new PerLanguageOption<bool>(
+                "BinaryParsers", nameof(IgnorePELoadError), defaultValue: () => false,
+                "Set this value to 'true' to ignore exceptions thrown in CanAnalyze method.");
     }
 }
--- a/src/BinaryParsers/BinaryParsersProperties.cs
+++ b/src/BinaryParsers/BinaryParsersProperties.cs
@@ -64,6 +64,6 @@ namespace Microsoft.CodeAnalysis.BinaryParsers
         public static PerLanguageOption<bool> IgnorePELoadError { get; } =
             new PerLanguageOption<bool>(
                 "BinaryParsers", nameof(IgnorePELoadError), defaultValue: () => false,
-                "Set this value to 'true' to ignore exceptions thrown in CanAnalyze method.");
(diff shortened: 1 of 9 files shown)


Sasinkas (approved):

mkacmar (commented):
Do we know why does the code actually throw an exception for the referenced files? Is our parser misbehaving or is it the binary itself? It feels to me we're exposing an implementation detail as a configurable option to the user here.

mkacmar (changes requested):
See my comment above.

src/BinaryParsers/BinaryParsersProperties.cs
```diff
@@ -60,5 +60,10 @@ public IEnumerable<IOption> GetOptions()
                 "environment variables at runtime. Use this argument instead for specifying the symbol path." +
                 "WARNING: Be sure to specify a local file cache in the symbol path if at all possible, in order " +
                 "to avoid the possibility of significance time-to-analyze performance degradataion.");
+
+        public static PerLanguageOption<bool> IgnorePELoadError { get; } =
+            new PerLanguageOption<bool>(
+                "BinaryParsers", nameof(IgnorePELoadError), defaultValue: () => false,
+                "Set this value to 'true' to ignore exceptions thrown in CanAnalyze method.");
```
mkacmar, line 67: Do you want to make this message more specific to errors while loading PEs?

src/Test.FunctionalTests.BinSkim.Driver/AnalyzeCommandTests.cs
```diff
@@ -135,6 +135,58 @@ public void AnalyzeCommand_DeterminismTest()
             log.Runs[0].Invocations[0].ToolConfigurationNotifications.Count(t => t.Message.Text.Contains("E_PDB_NOT_FOUND")).Should().Be(1);
         }
 
+        [Fact]
+        public void AnalyzeCommand_IgnoreExceptionInCanAnalyzeErrorTest()
```
mkacmar, line 139: Do you think reflecting this test is specific for PEs makes sense in test name here?

src/Test.FunctionalTests.BinSkim.Driver/AnalyzeCommandTests.cs
```diff
@@ -135,6 +135,58 @@ public void AnalyzeCommand_DeterminismTest()
...
+        {
+            if (!PlatformSpecificHelpers.RunningOnWindows()) { return; }
```
mkacmar, line 141: Is this Windows check required here?

src/Test.FunctionalTests.BinSkim.Driver/AnalyzeCommandTests.cs
```diff
@@ -135,6 +135,58 @@ public void AnalyzeCommand_DeterminismTest()
...
+            var command = new MultithreadedAnalyzeCommand();
+            command.Run(options);
+            var log = SarifLog.Load(fileName);
+            log.Runs[0].Invocations[0].ToolConfigurationNotifications.Count(t => t.Message.Text.Contains("System.InvalidOperationException: RVA does not belong to any section")).Should().Be(4);
+        }
+
+        [Fact]
+        public void AnalyzeCommand_NotIgnoreExceptionInCanAnalyzeErrorTest()
+        {
+            if (!PlatformSpecificHelpers.RunningOnWindows()) { return; }
```
mkacmar, line 167: Same as above, is this required?

src/Test.FunctionalTests.BinSkim.Driver/AnalyzeCommandTests.cs
AllDwarf: Hmm good point maybe, it is not, I'll try it out.

mkacmar (approved):

PR #1049, opened 2025-02-26
Dynamic Lookup of Compiler Name for Managed PE for BA4001

lukaskohl-msft:
Adds the compiler name for managed PE's by looking into compiland details.

Additionally slightly restructures the rule while ensuring comparable data.


mkacmar:
@lukaskohl-msft is it feasible to have any tests for this change?

mkacmar (approved):

AllDwarf (approved):
LGTM, just please since we're in the release process update the branch afterwards to the lates.

PR #1056, opened 2025-03-17
Fixing release 4.4.0 with the revision possibility removal

AllDwarf:
This is fixing revision versioning back to patch only, in the logs BinSkim version is showing 0.0.0.0 instead of actual one.

Removal of BA2026 test because of the issues with the testing file. It is flagged as a malware.


src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
```diff
@@ -1252,14 +1252,14 @@ public void BA2025_EnableShadowStack_NotApplicable()
...
 
-        [Fact]
-        public void BA2026_EnableMicrosoftCompilerSdlSwitch_Pass()
-        {
-            if (BinaryParsers.PlatformSpecificHelpers.RunningOnWindows())
-            {
-                this.VerifyPass(new EnableMicrosoftCompilerSdlSwitch(), useDefaultPolicy: true);
-            }
-        }
+        // [Fact]
```
mkacmar, line 1255: Can we just remove this just if we think it's not a valuable test? If we think the test has value, but there are $REASONS why we can't run it now, perhaps we can mark it as skipped?

mkacmar (changes requested):

src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
AllDwarf: it is valuable, but we lost binary for it so we will need to generate new one which will not have any security implications.

mkacmar (approved):

PR #1070, opened 2025-04-30
Fix XML properties being interpreted as tags

jbzdarkid:
Github's markdown preview will treat text in angle brackets (like `<ChecksumAlgorithm>`) as an HTML tag, even if it's not a valid tag. In these cases, the text will be invisible in the preview.

By surrounding the text in backticks (``), it will be escaped, and actually render to the user.


jbzdarkid:
@microsoft/1es-sdl-engineering I'm not sure why 2 of the checks are showing as pending -- I am a microsoft employee (and can sign the CLA) but don't know how to trigger that explicitly.

plade:
This is fixed now.

PR #1097, opened 2025-08-31
catch ERR998/997

Sasinkas:


src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -527,7 +527,7 @@ public bool IsManagedResourceOnly
                     return this.isManagedResourceOnly.Value;
                 }
 
-                this.isManagedResourceOnly = this.metadataReader.MethodDefinitions.Count == 0;
+                this.isManagedResourceOnly = this.metadataReader?.MethodDefinitions.Count == 0;
```
michaelcfanning, line 530: maybe you can just have a check for IsManaged at the top of this helper? would be clearer, I think.

if (!this.IsManaged) { return false;}

if you do that, afterwards you could Debug.Assert on this.metadataReader being non-null.

src/BinSkim.Rules/PERules/WindowsBinaryAndPdbSkimmerBase.cs
```diff
@@ -133,13 +133,31 @@ public static void LogExceptionLoadingPdb(IAnalysisContext context, PdbException
...
+            string key = "";
+            if (pdbException == null)
             {
+                context.Logger.LogConfigurationNotification(
+                   Errors.CreateNotification(
+                       context.CurrentTarget.Uri,
+                       "ERR997.ExceptionLoadingPdb",
+                       string.Empty,
+                       FailureLevel.Error,
+                       new Exception(),
```
michaelcfanning, line 145: >new Exception() [](http://example.com/codeflow?start=23&length=15)

We are creating an empty Exception()? This code looks very suspicious and probably indicates we may not have the right pattern here.

Do you have a description somewhere of the problem you're trying to fix?

michaelcfanning:
>         public static void LogExceptionLoadingPdb(IAnalysisContext context, PdbException pdbException, string pdbLoadTrace)

Are you going to add unit tests for this change? They might help me understand what's going on here.

---
Refers to: src/BinSkim.Rules/PERules/WindowsBinaryAndPdbSkimmerBase.cs:130 in 680584c. [](commit_id = 680584c834bd981ef2bb3dbda751b2917d47ccae, deletion_comment = False)

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -593,7 +593,12 @@ public bool IsDotNetFramework
 
         internal static ManagedPlatform ComputeIsDotNetCore(MetadataReader metadataReader)
         {
-            if (metadataReader.AssemblyReferences.Count == 0)
+            if (metadataReader?.AssemblyReferences == null)
```
michaelcfanning, line 596: >metadataReader [](http://example.com/codeflow?start=16&length=14)

why is this value null? it looks like this method is protected universally by an IsManaged check which makes me think this should be non-null. but you are seeing a condition when this isn't so?

src/BinaryParsers/PEBinary/PEBinary.cs
```diff
@@ -80,7 +80,24 @@
...
+            {
+                try
+                {
+                    return this.pdb?.Value;
+                }
+                catch (DllNotFoundException)
+                {
+                    return null;
+                }
+                catch (Exception)
```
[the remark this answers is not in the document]
Sasinkas: Yes, I will revert this whole part

src/BinaryParsers/PEBinary/PEBinary.cs
```diff
@@ -80,7 +80,20 @@ public static void ClearLocalSymbolDirectoriesCache()
...
         public PE PE { get; private set; }
 
-        public Pdb Pdb => this.pdb?.Value;
+        public Pdb Pdb 
+        {
+            get
+            {
+                try
+                {
+                    return this.pdb?.Value;
```
michaelcfanning, line 89: >pdb [](http://example.com/codeflow?start=32&length=3)

this doesn't look right to me. first, the code is hard to read, why would a field retrieval raise an exception?

the answer is that the field is itself lazily initialized. ok. :)

so my question is why you don't update LoadPdb instead to return null for this case.

src/BinSkim.Rules/PERules/WindowsBinaryAndPdbSkimmerBase.cs
Sasinkas: I added tests, and I make the exception more specific.

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -516,6 +516,12 @@ public bool IsManagedResourceOnly
         {
             get
             {
+                //todo
```
lukaskohl-msft, line 519: To do... what here? 👀

Sasinkas: deleted, this was there for debugging purposes, but I forgot to delete it! thank you :)

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
michaelcfanning: The question is, if we have a managed binary that raises the exception you have provoked, is it impossible for us to have a non-null metadata reader.

Is that the case? If it is, then your changes are reasonable. But if what's happening here is that we're simply not processing the exceptions properly to finish instantiating a valid metadata reader, that's the change that we want.

PR #1100, opened 2025-09-02
Azure pipelines/202509

Sasinkas:


mkacmar:
Would you mind squashing those `Update 1ES.binskim.BuildAndTest.EXTERNAL.yml for Azure Pipelines` commits before merging?

PR #1106, opened 2025-09-10
Updated build and baselining

michaelcfanning:
This change is a monster. We had an unfortunate set of entangled issues conspiring to prevent us from rebaselining and running tests successfully. Rather than try to decompose and properly order the work, a plan for which was not at all obvious and which likely would have increased costs, I am proposing this large and complex change. The reality is that detailed baseline comparisons will always be costly and arguably what we've done here requires a single pass only. I myself have gone through every baseline change in detail. Here's what we have:

- Lots of build clean-up from @nguerrera [NOTE: NOW MERGED]. See the description on this [commit](https://github.com/microsoft/binskim/commit/561faa8fb2aeec8fdc0c7aa49cc2a6ac74ee1de7) for details.
- Add a hidden option to specify the `--enlistment-root`. When present, this indicates a general need to normalize test output, starting with converting hard-coded paths into a relative path from the enlistment root.
- When invoked, the test normalization code path looks for any `SarifLogger` instances that have been created. This logger is replaced with a `NormalizingSarifLogger` instance that wraps the SARIF logger and provides normalization.
- The normalizer does something unpleasant: it retrieves the underlying SARIF logger `Run` instance via reflection, so that we can clobber various of its properties that are inconsistent across builds and runtime environments. In the future, normalization should be pushed into the SARIF SDK. A very minimal change there would be to simply make the logger Run instance publicly available. I've filed a work item [here](https://github.com/microsoft/sarif-sdk/issues/2870).
- After normalizing the logger Run property, the `NormalizingSarifLogger` initializes a rewriter instance that trims the enlistment root from any observed URI in a Sarif `ArtifactLocation`. We create a specialized private `SarifRewritingVisitor` for this (which knows how to visit all URLs in SARIF results, notifications, etc.
- One outcome of changing to a relative path is that the logger won't have an absolute path to the file with the result that it can't hash it. If we wanted to insist on hashes in the output, we'd need to reconsider our approach. I think it's fine to dispense with this data. If anyone disagrees, speak up!
- Critically, all the above allows us to delete duplicated file normalization logic that previously existing in `BaselineTests.RunRules` and the `UpdateBaselines.ps1` powershell script. Both code paths now simply pass the enlistment root command to invoke the converged normalization logic.
- We now set `System.Globalization.UseNls` to true which has the effect of forcing Windows collation when sorting strings. The presence of an underscore in an object file name would otherwise provoke different ordering in SARIF output.
- Properly resolving the above, we observe some baseline changes that had accrued in the development process. SARIF differences that aren't meaningful (in the sense of literally altering scan results) won't fail the baselining test. And so, changes to rule string in the SARIF were not properly reflected in the baseline.
- Nick's helpful build changes also reveal that we'd checked in baselines for scans that raised exceptions (for PDB loading). It is too easy to allow this sort of thing to get checked in. We should make another change that hardens the tests to insist on an exception free run (unless explicitly configured otherwise for a negative test. [Work item filed](https://github.com/microsoft/binskim/issues/1108)
- I have optimistically updated the SH file for rebaselining on Linux but it is untested. [Work item filed](https://github.com/microsoft/binskim/issues/1107)

@Sasinkas @nguerrera


nguerrera:
See also my commit comment in https://github.com/microsoft/binskim/pull/1106/commits/561faa8fb2aeec8fdc0c7aa49cc2a6ac74ee1de7 for the details of what I did.

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/Binskim.linux-x64.dll.sarif
```diff
@@ -95,8 +95,8 @@
       ],
       "tool": {
         "driver": {
-          "name": "testhost",
-          "version": "15.0.0.0",
+          "name": "BinSkim",
```
michaelcfanning, line 98: Intrusive change #2: we normalize the name and other driver details.

src/Test.FunctionalTests.BinSkim.Driver/BaselineTestData/Expected/Binskim.linux-x64.dll.sarif
```diff
@@ -210,48 +209,19 @@
...
-                "text": "'Binskim.linux-x64.dll' was not evaluated because its PDB could not be loaded (E_PDB_FORMAT)."
-              },
-              "level": "error",
-              "descriptor": {
-                "id": "ERR997.ExceptionLoadingPdb"
-              }
-            }
-          ],
-          "executionSuccessful": false
+          "executionSuccessful": true
```
michaelcfanning, line 217: >executionSuccessful [](http://example.com/codeflow?start=11&length=19)

For this file, our broken build system did not copy supporting PDB libraries properly and so this scan session raised an exception. And we checked in errors that indicate the scan literally failed. We should try to prevent this from happening again, I filed a work item noted in the description. It's hard to read JSON diffs and understand what changes are good or not...

src/BinSkim.Driver/BinSkim.Driver.csproj
```diff
@@ -7,7 +7,8 @@
     <TargetFramework>$(NetCoreVersion)</TargetFramework>
     <TargetLatestRuntimePatch>True</TargetLatestRuntimePatch>
     <OutputType>Exe</OutputType>
-    <RuntimeIdentifiers>win-x64;linux-x64;osx-x64</RuntimeIdentifiers>
+    <Platforms>x64</Platforms>
+    <SelfContained>true</SelfContained>
```
nguerrera, line 11: Was this to debug the issue you saw with restore failing? The RuntimeIdentifiers need to be set and SelfContained should not need to be set always as that will slow down incremental IDE builds unnecessarily.

michaelcfanning:
@Sasinkas  or @martin-reznik , please merge when you like and I will work on the remaining output non-determinism issue.

martin-reznik (approved):

src/Directory.Build.targets
```diff
@@ -0,0 +1,31 @@
...
+    <DeployDia Condition="$(RuntimeIdentifier) == 'win-x64'">true</DeployDia>
+    <!-- Deploy dia dll to unofficial, unpublished build, but only on Windows -->
+    <DeployDia Condition="$(RuntimeIdentifier) == '' and $([MSBuild]::IsOSPlatform('Windows')) AND '$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'x64'">true</DeployDia>
+    <!-- But don't deploy to non-exes and non-test projects-->
+    <DeployDia Condition="'$(OutputType)' != 'Exe' and !$(MSBuildProjectName.StartsWith('Test'))">false</DeployDia>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(DeployDia)' == 'true'">
+    <MsDiaLibs Include="$(MSBuildThisFileDirectory)..\refs\x64\*" />
+    <MsDiaLibs Include="$(MSBuildThisFileDirectory)..\refs\x64\*" />
```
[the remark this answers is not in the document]
martin-reznik: Keeping it for now - will fix in future

PR #1112, opened 2025-09-12
Build cleanup

nguerrera:
1. Stop Using Platforms/Platform and rely on RuntimeIdentifier only.
2. Use ArtifactsPath to move output directories. Projects were building to the same directory and this is risky and not really supported.
3. Replace hack to overwrite files in microsoft.diagnostics.tracing.tracevent with a lesser hack to exclude the conflicting files in the build.
4. Stop using local package directory that looks to have been just for (3).
5. Disable CandidateAssemblyFile resolution, which was a long part of debugging the conflict with traceevent. This should be really have been off by default and it reminds to go disable it in our other repos.
6. Don't set RuntimeIdentifiers in every project, it is only needed in the root exe project and putting it everywhere is triggering more project builds than necessary.
7. Use TargetFramework instead of TargetFrameworks (plural) since there's only one TFM. Make build output folders tidier and speeds up the build.
8. Remove unused x86 FCIBs and unused CommandLine.dll FCIB.
9. Don't set SelfContained on exe project. Let the publish do this. Speeds up iteration time in IDE.
10. Remove workaround in BuildAndTest.sh for bug fixed in 2017.
11. Don't pass configuration to dotnet restore, it is a cross-configuration operation.
12. Consolidate test API to locate files into one place. A `Shared\TestHelpers.cs` is injected into each test project for this. Also, remove P2P references between tests that were just to get this helper API from PEBinaryTests. That was not done consistently and many other times the helpers were manually copied around.
13. Remove xunit.runner.console references. xunit.runner.visualstudio is sufficient for `dotnet test` on the command line.
14. Use `dotnet format` built into the SDK and fix a few formatting issues it flagged. This is done because one of my changes ran into bug in old dotnet-format.
15. Delete unused Test.ConcurrencyTests project that was not in solution and no longer builds successfully.
16. Delete "nightly" test option from BuildAndTest.cmd that only would have run tests from (15) that don't exist.
17. Make BuildAndTest.sh and BuildAndTest.cmd run all tests in sln. One project, Test.UnitTests.BinSkim.Driver was missing.
18. Make BuildAndTest.shm more like BuildAndTest.cmd: take Configuration argument and don't try to run tests if build fails.
19. Fix test failures on Linux.


nguerrera:
(Force push since @michaelcfanning's 2nd approval was just to squash the commits since first approval down to 1 more.)

michaelcfanning:
If someone will merge this PR I can get the next wave of work submitted. :)

martin-reznik (approved):

PR #1118, opened 2025-10-03
Fix escaping of XML-like tags in documentation

rcastano:
When looking at the BA2004 documentation (https://github.com/microsoft/binskim/blob/main/docs/BinSkimRules.md#rule-ba2004enablesecuresourcecodehashing) in the preview tab, I noticed what seemed to be an empty property name:
```
     on the csc.exe command-line or populate the '' project property with 'SHA256'
```
 When I checked the Code tab, the property was mentioned:
```
    on the csc.exe command-line or populate the '<ChecksumAlgorithm>' project property with 'SHA256'
```

### DIFF
--- a/src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs
+++ b/src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs
@@ -81,7 +81,7 @@ namespace Microsoft.CodeAnalysis.IL.Rules
                 // '{0}' is a managed binary compiled with an insecure ({1}) source code hashing algorithm.
                 // {1} is subject to collision attacks and its use can compromise supply chain integrity.
                 // Pass '-checksumalgorithm:SHA256' on the csc.exe command-line or populate the project
-                // <ChecksumAlgorithm> property with 'SHA256' to enable secure source code hashing.
+                // \<ChecksumAlgorithm\> property with 'SHA256' to enable secure source code hashing.
                 context.Logger.Log(this,
                     RuleUtilities.BuildResult(ResultKind.Fail, context, null,
                     nameof(RuleResources.BA2004_Error_Managed),


martin-reznik (approved):

mkacmar:
One needs to update these doc comments e.g. https://github.com/microsoft/binskim/blob/d5572eb3e68e21d0c36df2c5d5074dc488469978/src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs#L81-L84 in the rule itself for Markdown documentation change to stick. See 

https://github.com/microsoft/binskim/blob/d5572eb3e68e21d0c36df2c5d5074dc488469978/BuildAndTest.cmd#L64

otherwise the change would be overwritten with next script run. The same message appears on CLI too so any escaping also appears there.

rcastano:
> One needs to update these doc comments e.g.
> 
> https://github.com/microsoft/binskim/blob/d5572eb3e68e21d0c36df2c5d5074dc488469978/src/BinSkim.Rules/PERules/BA2004.EnableSecureSourceCodeHashing.cs#L81-L84
> 
> in the rule itself for Markdown documentation change to stick. See
> https://github.com/microsoft/binskim/blob/d5572eb3e68e21d0c36df2c5d5074dc488469978/BuildAndTest.cmd#L64
> 
> otherwise the change would be overwritten with next script run. The same message appears on CLI too so any escaping also appears there.

For the Description part, do I need to update any other file? I found:
[src/BinSkim.Rules/RuleResources.resx](https://github.com/microsoft/binskim/blob/d5572eb3e68e21d0c36df2c5d5074dc488469978/src/BinSkim.Rules/RuleResources.resx#L435)
[src/BinSkim.Rules/RuleResources.Designer.cs](https://github.com/microsoft/binskim/blob/d5572eb3e68e21d0c36df2c5d5074dc488469978/src/BinSkim.Rules/RuleResources.Designer.cs#L903)

They don't look like they'd impact the generated file, but not sure.

PR #1126, opened 2025-11-06
Update 1ES.binskim.BuildAndTest.EXTERNAL.yml for Azure Pipelines

Sasinkas:


.pipelines/1ES.binskim.BuildAndTest.EXTERNAL.yml
```diff
@@ -24,7 +23,7 @@ extends:
       jobs:
         - job: linux
           pool:
-            name: pool-1espt-linux-mseng-dev
+            name: guardian-build-infra-linux-arm64-dev
```
martin-reznik: I think this should be ```guardian-build-infra-linux-x64-dev```

PR #1160, opened 2026-03-27
BA2025: Skip CETCOMPAT check on 32-bit binaries

danmoseley:
Windows does not support CET Shadow Stack (`/CETCOMPAT`) on 32-bit binaries. The [MSVC documentation](https://learn.microsoft.com/en-us/cpp/build/reference/cetcompat?view=msvc-170) has been [updated](https://github.com/MicrosoftDocs/cpp-docs/pull/5911) to reflect this.

This PR updates BA2025 to return NotApplicable for non-64-bit PE binaries, by checking `PEHeader.Magic != PEMagic.PE32Plus` (the same approach used by BA2001).

### Changes
- Add PE32Plus check in `EnableShadowStack.CanAnalyzePE()` after existing ARM/ARM64EC exclusions
- Move x86 test fixture from `Pass/` to `NotApplicable/`
- Update 7 x86 SARIF baselines (remove BA2025 warning + adjust ruleIndex values)

Fixes #1149

> [!NOTE]
> This PR description was AI/Copilot-generated.


martin-reznik:
merged to dev branch so we can run required pipelines on it

PR #1161, opened 2026-03-27
Fix DWARF parser crash on Rust binaries: handle abbreviation table terminator

martin-reznik:
## Problem

BinSkim's DWARF parser crashes with `ArgumentNullException` when analyzing Rust-compiled ELF binaries containing DWARF debug info. This prevents BA2030.RustBuildWithCompliantToolchainDWARF from running on Linux.

**Stack trace:**
```
ArgumentNullException: Value cannot be null. (Parameter 'source')
   at System.Linq.Enumerable.Any(...)
   at DwarfCompilationUnit.ReadData(...)
   at DwarfCompilationUnit..ctor(...)
   at DwarfSymbolProvider.ParseAllCompilationUnits(...)
   at ElfBinary.get_DebugFileType()
   at DwarfSkimmerBase.CanAnalyze(...)
```

## Root Cause

`GetDebugDataDescription()` in `DwarfCompilationUnit.cs` did not recognize the abbreviation table null terminator (code == 0) per [DWARF5 spec section 7.5.3](https://dwarfstd.org/doc/DWARF5.pdf#section.7.5.3):

> *"The abbreviations for a given compilation unit end with an entry consisting of a 0 byte for the abbreviation code."*

When the parser encountered code 0, it attempted to parse it as a valid entry, corrupting the reader position. Subsequent lookups failed, hit EOF, and returned a default `DataDescription` struct with `Attributes == null`. The caller `ReadData` then crashed on `description.Attributes.Any(...)`.

## Fix

Three changes in `src/BinaryParsers/ElfBinary/Dwarf/DwarfCompilationUnit.cs`:

1. **Handle abbreviation table terminator** - check for `code == 0` after reading the abbreviation code, per DWARF5 spec 7.5.3
2. **Replace `throw NotImplementedException`** with graceful return of empty `DataDescription` when a requested code is not found
3. **Bail out of ReadData** when `Attributes` is null/empty - without attribute descriptions, the parser cannot determine the DIE data size to skip, so it stops and preserves what was parsed so far

## Tests

Five new unit tests in `DwarfCompilationUnitFormTests`:

- `AbbrevTable_WithNullTerminator_ParsesCorrectly` - verifies code 0 terminates the table
- `AbbrevTable_WithTrailingDataAfterTerminator_DoesNotCorruptParsing` - shared .debug_abbrev with multiple CUs
- `AbbrevTable_RequestingNonExistentCode_DoesNotThrow` - graceful handling of missing codes
- `AbbrevTable_NullTerminator_StopsBeforeGarbageData` - proves parser stops at code 0, doesn't read garbage
- `ReadData_StopsWhenAbbrevCodeNotFound` - proves ReadData stops at unknown code and preserves prior DIEs


Sasinkas (approved):
LGTM

PR #1162, opened 2026-03-27
Add internal validation stage to external pipeline, switch to production pools

martin-reznik:
### Internal validation stage
Added \InternalValidation\ stage that runs **in parallel** with \BuildAndTest\. It triggers an integration pipeline with the current commit hash and waits for it to complete. If the internal build fails, the PR is blocked from merging.

### Production pools
Switched all agent pools from \-dev\ to production (\guardian-build-infra-windows-x64\, \guardian-build-infra-linux-x64\).

### DIFF
--- a/.pipelines/1ES.binskim.BuildAndTest.EXTERNAL.yml
+++ b/.pipelines/1ES.binskim.BuildAndTest.EXTERNAL.yml
@@ -81,11 +81,7 @@ extends:
           - task: ComponentGovernanceComponentDetection@0
             condition: eq(variables['runComponentGovernance'], 'true')
 
-    # ── Stage 2: Trigger internal build with the current commit ──────────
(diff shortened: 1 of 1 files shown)


.pipelines/1ES.binskim.BuildAndTest.EXTERNAL.yml
```diff
@@ -80,3 +80,81 @@ extends:
 
           - task: ComponentGovernanceComponentDetection@0
             condition: eq(variables['runComponentGovernance'], 'true')
+
+    # ── Stage 2: Trigger internal build with the current commit ──────────
+    # Runs in parallel with BuildAndTest. Queues the internal BinSkimInternal
```
mkacmar: Nitpick - can we not split this comment into so many lines?

mkacmar (approved):

mkacmar (approved):

PR #1163, opened 2026-03-27
Add --disable-archive-extraction feature flag

martin-reznik:
## Summary

SARIF SDK 4.6.0 introduced automatic extraction and scanning of archive formats (ZIP, OPC packages, etc.) via \OpcFileExtensions\. This new behavior may be undesirable for existing users who only want to scan top-level binaries.

This PR adds a \--disable-archive-extraction\ CLI flag that, when enabled, clears the \OpcFileExtensions\ set on the analysis context, preventing the SARIF driver from identifying and recursively extracting archive files.

## Changes

| File | Change |
|------|--------|
| \AnalyzeOptions.cs\ | New \ool? DisableArchiveExtraction\ option with \[Option]\ attribute |
| \MultithreadedAnalyzeCommand.cs\ | Wire option in \InitializeGlobalContextFromOptions\ to clear \OpcFileExtensions\ |
| \BinSkim.cs\ | Bare-switch rewrite support (\--disable-archive-extraction\ without \	rue\/\alse\) |
| \CommandLineTests.cs\ | Command-line parsing test following \DisableTelemetryCommandlineTest\ pattern |

## Usage

\\\ash
# Disable archive extraction (archives will be skipped, not expanded)
binskim analyze *.exe --disable-archive-extraction

# Explicit true/false
binskim analyze *.exe --disable-archive-extraction true
binskim analyze *.exe --disable-archive-extraction false
\\\

## How it works

The SARIF SDK's \MultithreadedAnalyzeCommandBase.EnumerateArtifact\ checks \OpcFileExtensions\ to detect archive files (\.zip\, \.jar\, \.msix\, etc.). Setting this to an empty set prevents any file from matching the OPC check, effectively disabling archive extraction while preserving all other analysis behavior.


Sasinkas (approved):
LGTM

PR #1167, opened 2026-04-01
Dwarf memory reader exception handeling

Sasinkas:
[Enhance DWARF memory reader with bounds checking and exception handling for read operations]
This branch implements defensive programming improvements to the ELF/DWARF binary parser by introducing proper exception handling for buffer boundary violations. Instead of allowing buffer overreads to silently fail or crash, the code now:

Detects when reads would exceed buffer bounds
Throws a meaningful DwarfBufferOverreadException
Includes comprehensive unit tests covering both success and exception paths
Updates calling code (DwarfSymbolProvider) to handle the new exception


src/Test.UnitTests.BinaryParsers/Dwarf/DwarfMemoryReaderTests.cs
```diff
@@ -225,5 +225,73 @@ public void Peek_DoesNotAdvancePosition()
             reader.Peek().Should().Be(0x42);
             reader.Position.Should().Be(0);
         }
+        }
```
martin-reznik: Does this file actually compile? :)

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
```diff
@@ -60,6 +60,14 @@ public bool IsEnd
             }
         }
 
+        private void EnsureAvailable(uint bytesToRead)
+        {
+            if (bytesToRead > (uint)Data.Length || Position > (uint)Data.Length - bytesToRead)
```
martin-reznik, line 67: First part in front of || seems redundant (it's equivalent of second part, with Position = 0)

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
```diff
@@ -60,6 +60,14 @@ public bool IsEnd
...
+            {
+                throw new InvalidOperationException("Attempted to read past end of DWARF data buffer.");
```
martin-reznik: Thinking if we should not set position to the end of the data as well. To make sure no loop tries to re-read that broken sequence (if they would not handle the exception correctly)

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
```diff
@@ -205,15 +219,42 @@ public ulong ReadUlong(uint size)
         /// </summary>
         public ulong ULEB128()
         {
+            uint startPosition = Position;
+
+            if (Position >= Data.Length)
+            {
+                throw new InvalidOperationException("Attempted to read past end of DWARF data buffer.");
+            }
```
martin-reznik, line 227: This seems to be redundant with the check in first iteration of while

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
```diff
@@ -205,15 +219,42 @@ public ulong ReadUlong(uint size)
...
+                x |= (uint)((b & 0x7f) << shift);
                 shift += 7;
                 Position++;
             }
+
+            if (Position >= Data.Length)
+            {
+                Position = startPosition;
+                throw new InvalidOperationException("Attempted to read past end of DWARF data buffer.");
+            }
```
martin-reznik, line 254: This is also redundant - it does the same check as last iteration of the while above

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
```diff
@@ -224,17 +265,45 @@ public ulong ULEB128()
         /// </summary>
         public uint SLEB128()
```
martin-reznik, line 264: Same questions as for ULEB128 above

martin-reznik, line 264: Same questions as for ULEB128 above

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
martin-reznik: Also, would it be worth having our custom exception for this scenario?

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
```diff
@@ -83,6 +92,7 @@ public byte Peek()
         /// <typeparam name="T">Type of the structure to be read</typeparam>
         public T ReadStructure<T>()
```
martin-reznik, line 95: This one is missing test

martin-reznik (changes requested):

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: Even if Position is 0, there might be a scenario where we want to read more bytes than length (Data.Length = 4 and bytesToRead = 8 -> that is the kind of scenario where first part is necessary

src/Test.UnitTests.BinaryParsers/Dwarf/DwarfMemoryReaderTests.cs
Sasinkas: it works now, not sure how I missed that :)

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: deleted

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: deleted

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: deleted

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: deleted

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: deleted

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: exception changed

Sasinkas: Well I put Position into exception so it would not make sense to change that

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: updated (deleted first part)

Sasinkas: After some configuration and adding tests I'll keep it there

src/Test.UnitTests.BinaryParsers/Dwarf/DwarfMemoryReaderTests.cs
```diff
@@ -225,5 +225,107 @@ public void Peek_DoesNotAdvancePosition()
...
+        public void ReadUshort_WhenNotEnoughBytesRemain_ThrowsDwarfBufferOverreadException()
+        {
+            using var reader = new DwarfMemoryReader(new byte[] { 0x01 }); // only one byte available
+
+            Assert.Throws<DwarfBufferOverreadException>(() => reader.ReadUshort());
+            reader.Position.Should().Be(0);
+        }
+
+        [Fact]
+        public void ReadUint_WhenNotEnoughBytesRemain_ThrowsInvalidOperationException()
```
martin-reznik: Name of the test does not reflect expected exception

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
```diff
@@ -6,6 +6,8 @@
 using System.Runtime.ExceptionServices;
 using System.Runtime.InteropServices;
 
+using Microsoft.CodeAnalysis.BinaryParsers.Dwarf;
+
 namespace Microsoft.CodeAnalysis.BinaryParsers.Dwarf
```
martin-reznik: Do we need to have using, when we are defining that namespace?

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
martin-reznik, line 174: Shouldn't we add EnsureAvailable(3) here, just for the sake of consistency?

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
```diff
@@ -129,9 +141,9 @@ public string ReadString()
                 Position += (uint)result.Length + 1;
                 return result;
             }
-            catch (Exception ex)
+            catch (Exception ex) when (ex is ArgumentException || ex is AccessViolationException)
             {
-                throw new InvalidOperationException("Failed to read string from memory.", ex);
+                throw new DwarfBufferOverreadException(Position, 1, Data.Length);
             }
         }
```
martin-reznik, line 150: The exception filter was narrowed from catch (Exception ex) to catch (Exception ex) when (ex is ArgumentException || ex is AccessViolationException). This is an improvement. However, ReadString has no pre-validation of Position before calling Marshal.PtrToStringAnsi((nint)(pointer + Position)).

If Position is past the buffer end, the marshal call reads from whatever memory pointer + Position points to. In .NET 6+, AccessViolationException is a corrupted-state exception and will not be caught — it will crash the process. The method should add EnsureAvailable(1) at the top to validate Position before the unsafe marshal call.

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
martin-reznik: It still seems present

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
martin-reznik: Again, still there :)

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
martin-reznik: does not seem so :)

martin-reznik (changes requested):

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: we don't

src/Test.UnitTests.BinaryParsers/Dwarf/DwarfMemoryReaderTests.cs
Sasinkas: changed

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: yes, changed

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: probably forgot to commit

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: same as above :/, deleted

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: added check in the beggining of method            
if(Position >= Data.Length)
            {
                throw new DwarfBufferOverreadException(Position, 1, Data.Length);
            }

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: deleted for good.

src/BinaryParsers/ElfBinary/Dwarf/DwarfMemoryReader.cs
Sasinkas: updated

martin-reznik (approved):

varujans (approved):

PR #1174, opened 2026-04-16
Suppress ERR997 for IL-only managed assemblies

### LINKED ISSUE #1173: ERR997 should not fire on managed resource assemblies (lots of noise)
# Suppress ERR997 (PDB not found) for IL-only managed assemblies

## Summary

`ERR997.ExceptionLoadingPdb` fires for IL-only managed assemblies (including satellite resource assemblies) even though no current BinSkim rule requires PDB data for IL-only binaries. This creates significant noise in scan results — for example, scanning dotnet/dotnet macOS artifacts produces thousands of ERR997 entries, all on IL-only assemblies, none actionable.

Resource assemblies are produced by AL (the managed linker) and have zero code and no PDB. So this is pure noise for anyone that ships localized resources.

What should happen: ERR997 should not happen for IL-only assemblies UNLESS a rule actually needs the PDB for an IL-only assembly, and none of the rules in this repo do.

## Root cause

In `WindowsBinaryAndPdbSkimmerBase.Analyze()` (line 50-59), the condition that gates ERR997 is:

```csharp
if (target.Pdb == null &&
    (!target.PE.IsManaged ||
      target.PE.IsMixedMode ||
      EnforcePdbLoadForManagedAssemblies))
```

For an IL-only managed assembly (`IsManaged=true`, `IsMixedMode=false`), this evaluates to `(!true || false || true)` = `true`, so ERR997 fires and the rule is skipped entirely (`return` at line 58).

`EnforcePdbLoadForManagedAssemblies` defaults to `true` and **no rule in the repo overrides it**. The only two PDB-using rules that apply to IL-only assemblies (BA2004 and BA2027) both use a different mechanism — they override `LogPdbLoadException => false` and handle null PDB gracefully in their `AnalyzePortableExecutableAndPdb()`.

## Proposed fix

**1. Remove `EnforcePdbLoadForManagedAssemblies` from the ERR997 condition:**

```csharp
// Before:
if (target.Pdb == null &&
    (!target.PE.IsManaged ||
      target.PE.IsMixedMode ||
      EnforcePdbLoadForManagedAssemblies))

// After:
if (target.Pdb == null &&
    (!target.PE.IsManaged ||
      target.PE.IsMixedMode))
```

This means:
- **Native binaries**: ERR997 still fires (PDB is critical for native-code rules)
- **Mixed-mode**: ERR997 still fires (has native code that needs PDB analysis)
- **IL-only managed**: ERR997 suppressed, rule proceeds with `Pdb=null`. Rules that need PDB already handle null gracefully in their `AnalyzePortableExecutableAndPdb()`.

**2. Mark `EnforcePdbLoadForManagedAssemblies` as obsolete** (it's `public virtual`, so removing it would be a breaking change):

```csharp
[Obsolete("This property is no longer used. IL-only managed assemblies no longer trigger "
        + "ERR997. Rules that require PDB for managed assemblies should check for null PDB "
        + "in their AnalyzePortableExecutableAndPdb implementation.")]
public virtual bool EnforcePdbLoadForManagedAssemblies => true;
```

## Why this is safe

- **No rule overrides `EnforcePdbLoadForManagedAssemblies`** — grep confirms zero overrides in the repo.
- **The only two PDB rules applicable to IL-only (BA2004, BA2027)** already override `LogPdbLoadException => false` and check `if (pdb == null) { return; }` in their analysis methods.
- **All other PDB rules** (BA2002, BA2006, BA2007, BA2011, BA2013, BA2014, BA2024, BA2026) return `NotApplicableToSpecifiedTarget` for IL-only assemblies in `CanAnalyzePE`, so they never reach `Analyze()`.
- **Future rules** that need PDB for IL-only assemblies should handle null PDB in `AnalyzePortableExecutableAndPdb()` — this is the correct pattern already established by BA2004/BA2027.

danmoseley:
Fixes https://github.com/microsoft/binskim/issues/1173

> [!NOTE]
> This is an AI-assisted draft.

> [!IMPORTANT]
> **Breaking change:** `EnforcePdbLoadForManagedAssemblies` is replaced by `RequiresPdbForManagedAssemblies` with inverted default (`false` instead of `true`). Any rule or external subclass that relied on the old property name or its `true` default will need to be updated.

## Problem

`ERR997.ExceptionLoadingPdb` fires for IL-only managed assemblies even though
no current BinSkim rule requires PDB data for IL-only binaries. This creates
significant noise — for example, scanning dotnet/dotnet macOS artifacts produces
thousands of ERR997 entries on IL-only assemblies (including satellite resource
assemblies), none actionable.

Note — this makes the raw logs cleaner. ERR997 is not going to Guardian or S360
so it's just a tidiness thing.

## Root cause

In `WindowsBinaryAndPdbSkimmerBase.Analyze()`, the old
`EnforcePdbLoadForManagedAssemblies` defaulted to `true` and no rule overrode
it, so the ERR997 condition always evaluated to true for IL-only assemblies
with no PDB.

## Fix

Replace `EnforcePdbLoadForManagedAssemblies` (default `true`) with
`RequiresPdbForManagedAssemblies` (default `false`). This is a **public API
breaking change** — the old property is removed, not obsoleted.

IL-only managed assemblies with no PDB now proceed to
`AnalyzePortableExecutableAndPdb()` instead of emitting ERR997 and returning
early. Resource-only assemblies (`IsManagedResourceOnly`) are also excluded
from the PDB requirement.

Rules that genuinely need PDB data for IL-only managed assemblies can override
`RequiresPdbForManagedAssemblies` to return `true`.

## Why this is safe

Verified all 17 `WindowsBinaryAndPdbSkimmerBase` subclasses:

- **14 rules** reject IL-only in `CanAnalyzePE` (directly or via
  `StackProtectionUtilities`) — they never reach `Analyze()` for IL-only
  assemblies.
- **3 rules** (BA2004, BA2027, BA4001) do accept IL-only assemblies. All three
  already null-check `Pdb` at the top of `AnalyzePortableExecutableAndPdb()`
  and return early when it is null.

Native and mixed-mode binaries still trigger ERR997 as before.


danmoseley:
@martin-reznik

danmoseley:
For our dotnet/dotnet repo, while this doesn't produce noise in Guardian/S360, it is super noisy in local logs, and makes SARIF much bigger — 63,702 notification entries bloat the SARIF to 80MB. With this fix it is 50MB

danmoseley:
Changed to the alternative approach suggested by @martin-reznik  -- note, this version is a public API breaking change, which I had avoided in the first change. I do not know whether that is a problem. Description updated above.

Sasinkas (commented):
I would like to test this one on our Internal repository, we have some additional rules that are using this.
- buildAndTest is working

danmoseley:
@Sasinkas

Sasinkas (approved):
LGTM

danmoseley:
@Sasinkas how can I get this merged? I've pinged Martin offline but he's quite busy. Can we merge?

danmoseley:
Oh - I see you just signed of today! thanks. Should I click merge?

Sasinkas (commented):
LGTM

Sasinkas (approved):
LGTM

PR #1177, opened 2026-04-18
Reduce redundant file I/O and memory copies in PE loading to give ~20% speedup

danmoseley:
# Title
Reduce redundant file I/O and memory copies in PE loading. Fundamentally binskim runs are very IO bound, especially run on a cold worker. No dramatic improvements are possible without some kind of heavy caching (which apparently Sarif SDK has attempted). Instead I was just looking for some easy wins that produce measurable improvement. I don't plan to investigate further after this set.

# Body
When BinSkim processes large numbers of PE files (e.g., 69K+ in VMR SDL builds), redundant file operations add up. This PR eliminates several sources of unnecessary overhead:

### 1. Zero-copy image bytes in PE constructor

Replace `GetContent().ToBuilder().ToArray()` (which copies the entire image) with `ImmutableCollectionsMarshal.AsArray()` to get the backing array directly — zero allocation, zero copy.

### 2. Compute hashes from in-memory bytes

`SHA1Hash` and `SHA256Hash` properties previously re-opened the file and streamed it through a manual `TransformBlock` loop. Now they use `SHA1.HashData()` / `SHA256.HashData()` on the already-loaded `pImage.array`, falling back to `File.ReadAllBytes` only if the image isn't loaded.

### 3. Use `RandomAccess` for PE magic byte check

`PEBinary.CanLoadBinary` previously allocated a `FileStream` just to read 2 bytes. Now uses `File.OpenHandle` + `RandomAccess.Read` with a stack-allocated span — no managed stream object, no heap allocation.

### 4. Skip redundant `CanLoadBinary` for PE files

Add `PEBinary.TryLoadBinary` which constructs the `PEBinary` directly and checks `.Valid` afterward, returning `null` for non-PE files. This eliminates the separate `CanLoadBinary` file open entirely for PE files — the dominant case. Non-PE files (ELF, Mach-O) fall through to the existing cascade.

### Cold-cache benchmark

Measured on Windows scanning a full .NET installation (55,083 files, 37,434 PE, 10.5 GB) with filesystem caches purged between every run. Five rounds, alternating order:

| Variant | Avg | Min | Max |
|---------|-----|-----|-----|
| main (stock) | 28.7s | 27.6s | 30.5s |
| This PR without change 4 | 25.1s | 23.8s | 25.9s |
| This PR | 22.3s | 21.3s | 25.4s |

The target set is predominantly IL-only managed DLLs, representative of the mix in dotnet/dotnet VMR outputs (~5,000 binaries). At that scale, the absolute savings are proportionally smaller (~0.6s), but the per-file overhead reduction applies to any workload.

**Net effect per file:** eliminates 1 large memory copy + up to 2 redundant file opens (hash + magic check), and for PE files avoids a third open by combining the magic check with construction.

All existing tests pass, plus 11 new tests covering image bytes fidelity, hash correctness, and `CanLoadBinary` edge cases (empty files, 1-byte files, non-PE content).

> [!NOTE]
> This PR was generated with the assistance of GitHub Copilot.


danmoseley:
Other things tried and rejected, for the record --

1. **Unified magic-byte sniffing** — read 4 bytes once in `BinaryTargetManager` to route directly to the correct binary type, avoiding the PE → ELF → MachO `CanLoadBinary` cascade. This showed negligible improvement since (in my .NET repo context) ~99% of scanned files are PE (which wins on the first try anyway). Dropped this as at least in my cases, relatively few files are ELF/MachO and duplicates identification logic from ELFSharp. If we wanted improvement here the right path would be to add to ELFSharp CanLoadBinary overloads that take a handle or stream.

2. **`FileOptions.SequentialScan`** — hint to the OS for prefetch. No measurable difference on NVMe; could help on slower storage but we lacked evidence. Not included as no evidence.

3. **`PEStreamOptions.PrefetchEntireImage`** — forces `PEReader` to read the entire file synchronously in the constructor. **Caused a 2× regression** (~28s → ~56s) due to massive concurrent I/O + memory pressure with 20 threads. Reverted.

danmoseley:
cc @martin-reznik

Sasinkas (approved):
looks good to me, but I would like @martin-reznik to look as well.

danmoseley:
@Sasinkas how can I get this merged? Can you help? It's been a little while now.

Sasinkas (commented):
LGTM

PR #1179, opened 2026-04-20
Handle compressed DWARF sections in ELF binaries

### LINKED ISSUE #1175: BA3004 reports DWARF version 0 for Go 1.26 Linux ELF binaries that contain DWARF v5
### Summary

BinSkim 4.4.9.7 reports **BA3004** with DWARF version **`0`** for Go 1.26 Linux ELF binaries, even though the same binaries contain **DWARF v5** according to `readelf`.

I reproduced this on simple Go-produced ELF binaries for `linux/amd64`. The minimal repro below uses a plain default `go build`, but I also saw the same BA3004 result on `-buildmode=pie`, `-buildmode=pie -linkmode external`, and `-buildmode=pie -linkmode external -extldflags ''-static-pie''`.

### Environment

- BinSkim: **4.4.9.7**
- Go: **go1.26.0**
- Repro path used here: build Linux ELF in WSL2 / Linux, then analyze that ELF with BinSkim on Windows
- `readelf` from GNU binutils

### Steps to reproduce

1. In WSL/Linux, create a minimal Go program in a path visible from Windows (for example under `/mnt/c/temp/binskim-go-dwarf5/hello.go`):

```go
package main

func main() {
    println("Hello, World!")
}
```

2. Build a Linux ELF with Go 1.26:

```bash
cd /mnt/c/temp/binskim-go-dwarf5
go version
GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o hello hello.go
```

3. Confirm the binary contains DWARF v5:

```bash
readelf --debug-dump=info hello | grep -m 1 "Version:"
```

Observed output:

```text
Version:       5
```

4. Analyze the same binary with BinSkim on Windows:

```powershell
C:\path\to\BinSkim.exe analyze --verbose --output C:\temp\binskim-go-dwarf5\hello.sarif C:\temp\binskim-go-dwarf5\hello
```

5. Inspect the BA3004 result in the generated SARIF:

```powershell
Select-String '"ruleId": "BA3004"' -Path C:\temp\binskim-go-dwarf5\hello.sarif -Context 0,10
```

Observed BA3004 excerpt:

```json
{
  "ruleId": "BA3004",
  "level": "error",
  "message": {
    "id": "Error",
    "arguments": [
      "hello",
      "0"
    ]
  }
}
```

### Expected behavior

If the ELF actually contains DWARF v5, BA3004 should either:

- pass, or
- at minimum report the detected version as **5**, not **0**.

### Actual behavior

BinSkim reports BA3004 as an error and says the file is using DWARF version **`0`**, even though `readelf` shows **DWARF version 5** in `.debug_info`.

### Notes

I also reproduced the same BA3004 / version `0` result on other Go 1.26 Linux build modes, including PIE and external-linker builds. So this does not appear to be specific to one linker mode; it looks like a BinSkim interpretation issue for these Go-produced ELF binaries.

qmuntal:
## Summary
- detect ELF `SHF_COMPRESSED` debug sections via raw section flags
- decompress zlib-compressed DWARF section contents before parsing
- add regression coverage for ELF64 compressed-section payloads

Fixes #1175


danmoseley:
Interesting that this hasn't come up before -- apparently no other toolchains produce this? or at least with flags typically used?

danmoseley:
cc @martin-reznik

src/BinaryParsers/ElfBinary/ElfBinary.cs
```diff
@@ -383,13 +390,76 @@ private byte[] LoadSection(string sectionName)
             {
                 if (section.Name == sectionName + ".dwo" || section.Name == sectionName)
                 {
-                    return section.GetContents();
+                    return GetSectionContents(section, Is64bit);
```
Sasinkas, line 393: Is this going to work even if Is64bit is false? Maybe there is something I am missing. :)

qmuntal:
Friendly ping 😸

src/BinaryParsers/ElfBinary/ElfBinary.cs
qmuntal: `GetSectionContents` should also handle 32-bit. I've added a test case, just in case.

danmoseley:
@martin-reznik ?

qmuntal:
Friendly ping 😸

danmoseley:
@Sasinkas can maybe help. @Sasinkas this is part of bringing up fuller support for running binskim over Microsoft golang binaries.

danmoseley:
@martin-reznik

src/BinaryParsers/ElfBinary/ElfBinary.cs
```diff
@@ -383,13 +390,76 @@ private byte[] LoadSection(string sectionName)
...
+            {
+                throw new NotSupportedException("ELF zstd-compressed sections are not supported on this runtime target because there is no built-in Zstandard stream to decode them.");
+            }
+
+            if (compressionType != ZlibCompressionType)
+            {
+                throw new NotSupportedException($"Unsupported ELF compression type: {compressionType}.");
+            }
+
+            ulong uncompressedSize = is64bit
```
mkacmar: Have you considered rejecting sizes greater than `Array.MaxLength` (e.g. with `InvalidDataException`) since this method ends up returning `byte[]`?

src/BinaryParsers/ElfBinary/ElfBinary.cs
```diff
@@ -383,13 +390,76 @@ private byte[] LoadSection(string sectionName)
...
+            }
+
+            int headerSize = is64bit ? 24 : 12;
+
+            if (contents.Length < headerSize)
+            {
+                throw new InvalidOperationException("Compressed ELF section header is truncated.");
+            }
+
+            uint compressionType = BinaryPrimitives.ReadUInt32LittleEndian(contents.AsSpan(0, sizeof(uint)));
```
mkacmar: I believe we already support both little and big endian when parsing ELF, can you extract reading these values to a helper and support both? (same for reading the size below)

src/BinaryParsers/ElfBinary/ElfBinary.cs
```diff
@@ -383,13 +390,76 @@ private byte[] LoadSection(string sectionName)
...
+
+            if (compressionType == ZstdCompressionType)
+            {
+                throw new NotSupportedException("ELF zstd-compressed sections are not supported on this runtime target because there is no built-in Zstandard stream to decode them.");
+            }
+
+            if (compressionType != ZlibCompressionType)
```
mkacmar: Do I understand correctly that this is an acceptable tradeoff for your use-case because Go emits zlib-compressed DWARF?

mkacmar (changes requested):
Can you add an end to end test using a real binary with compressed DWARF (there are existing tests that use checked-in ELF fixtures already)?

mkacmar:
Also, can you please rebase?

src/BinaryParsers/ElfBinary/ElfBinary.cs
qmuntal: Done.

src/BinaryParsers/ElfBinary/ElfBinary.cs
qmuntal: Yes, I'm only interested on zlib for now.

src/BinaryParsers/ElfBinary/ElfBinary.cs
qmuntal: Done.

mkacmar:
This looks good, can you please add some actual binaries like I mentioned in my previous comment? https://github.com/microsoft/binskim/blob/main/src/Test.UnitTests.BinaryParsers/Elf/ElfBinaryTests.cs

qmuntal:
> This looks good, can you please add some actual binaries like I mentioned in my previous comment?

Done!

mkacmar (approved):

PR #1182, opened 2026-04-28
Added extraction of sourcelink for compiler information events

varujans:
Extracts raw SourceLink JSON from PDBs during BinSkim analysis and includes it in the CompilerInformation AppInsights telemetry event


varujans:

src/BinSkim.Rules/PERules/BA4001.ReportPortableExecutableCompilerData.cs
```diff
@@ -82,6 +84,8 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
 
             if (target.PE.IsManaged)
             {
+                string sourceLinkJson = GetSourceLinkJson(target, pdb);
```
martin-reznik: this won't cover c/c++ I am afraid 🤔

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -235,7 +235,8 @@ public void Write(BinaryAnalyzerContext context, CompilerData compilerData)
                 { "moduleName", compilerData.ModuleName ?? string.Empty },
                 { "moduleLibrary", (compilerData.ModuleName == compilerData.ModuleLibrary ? string.Empty : compilerData.ModuleLibrary ?? string.Empty) },
                 { "hash", fileHash },
-                { "error", string.Empty }
+                { "error", string.Empty },
+                { "sourceLinkJson", compilerData.SourceLinkJson ?? string.Empty },
```
martin-reznik: appinsights has a limit of 8kb (should be per property), this may result in truncated content. There is SendChunkedContent for this purpose (see line 244), perhaps we should use that for sourcelink as well.

src/BinSkim.Rules/PERules/BA4001.ReportPortableExecutableCompilerData.cs
```diff
@@ -139,5 +144,30 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
...
+                else
+                {
+                    IEnumerable<string> docs = pdb.WindowsPdbGetSourceLinkDocuments();
+                    return docs?.FirstOrDefault();
+                }
+            }
+            catch (Exception ex)
+            {
+                // SourceLink extraction is best-effort — never fail the analysis.
+                Trace.WriteLine($"SourceLink extraction failed: {ex.Message}");
```
martin-reznik: We are not using Trace.WriteLine - I am afraid that would get lost (there is no trace listener configured), use Context.Logger for this

src/BinSkim.Rules/PERules/BA4001.ReportPortableExecutableCompilerData.cs
varujans: Good catch ! Addressed, so we can read C/C++. Thanks !

src/BinSkim.Sdk/CompilerDataLogger.cs
varujans: Addressed and added de-duplication per `CompilerInformation `events. Now they all will have same foreign key to chunked data

src/BinSkim.Rules/PERules/BA4001.ReportPortableExecutableCompilerData.cs
varujans: Good catch ! Replaced with` Context.Logger`. Thanks !

src/BinSkim.Rules/PERules/BA4001.ReportPortableExecutableCompilerData.cs
```diff
@@ -139,5 +147,43 @@ public override void AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext conte
...
+            try
+            {
+                if (pdb.FileType == PdbFileType.Portable)
+                {
+                    return target.PE.ManagedPdbGetSourceLinkDocument(pdb);
+                }
+                else
+                {
+                    IEnumerable<string> docs = pdb.WindowsPdbGetSourceLinkDocuments();
+                    return docs?.FirstOrDefault();
```
martin-reznik, line 168: Just for confirmation: are we okay with losing any subsequent data there?

src/BinSkim.Sdk/CompilerDataLogger.cs
```diff
@@ -58,6 +58,10 @@ public class CompilerDataLogger : IDisposable
         internal const string AssemblyReferencesId = "assemblyReferencesId";
         internal const string NumberOfBinaryAnalyzed = "numberOfBinaryAnalyzed";
         internal const string ChunkedAssemblyReferences = "chunkedassemblyReferences";
+        internal const string SourceLinkJson = "sourceLinkJson";
+        internal const string SourceLinkJsonId = "sourceLinkJsonId";
+        internal const string SourceLinkJsonEventName = "SourceLinkJsonInformation";
+        internal const string ChunkedSourceLinkJson = "chunkedsourceLinkJson";
```
martin-reznik: This seems to be unused?

src/Test.UnitTests.BinSkim.Driver/TelemetryTests.cs
```diff
@@ -22,7 +22,10 @@ public class TelemetryTests
         [Fact]
         public void Telemetry_TelemetryClientShouldBeNullIfEnvironmentVariablesNotSet()
         {
-            using var telemetry = new IL.Sdk.Telemetry();
+            // Use the internal constructor with null to verify that
+            // a null configuration produces a null TelemetryClient,
+            // bypassing the default connection string in EnvironmentResources.
+            using var telemetry = new IL.Sdk.Telemetry((TelemetryConfiguration)null);
```
martin-reznik: Any reason for this change? Shouldn't we test for the default constructor as well?

src/BinSkim.Rules/PERules/BA4001.ReportPortableExecutableCompilerData.cs
varujans: I think yes, as in practice PDBs contain a single SourceLink document in majority of cases. This matches how BA2027.EnableSourceLink consumes the same API. (.Any?) So I'd assume its sufficient for us to extract only 1

src/BinSkim.Sdk/CompilerDataLogger.cs
varujans: Good catch ! Clean up

src/Test.UnitTests.BinSkim.Driver/TelemetryTests.cs
varujans: Riight, it was degraded test after internal build. Reverted changes

src/BinSkim.Rules/PERules/BA4001.ReportPortableExecutableCompilerData.cs
martin-reznik: Well, BA2027 checks, if it has any source link present. Here we care about actual data. But I agree that we can keep it for now and fix later - if needed.

martin-reznik (approved):

PR #1183, opened 2026-04-29
Add Linux agent disk capacity pre-flight gate

martin-reznik:
Adds a Linux-only pre-flight step to the BuildAndTest job that checks for at least 6 GB of free space on the agent work folder before running the build.

If insufficient, the leg is marked `SucceededWithIssues` and remaining steps are skipped via a `skipLinux` variable, so that sticky agent-side disk-pressure failures on `guardian-build-infra-linux-x64` do not surface as real test failures and don't block the rest of the pipeline (Windows leg, ValidateReleasePRs, InternalValidation).

No cleanup is performed — shared paths (`/tmp`, `/var/tmp`, Docker) are not safe to wipe mid-job. Persistent pressure should be addressed at the agent-image / pool level.

### DIFF
--- a/.pipelines/1ES.binskim.BuildAndTest.EXTERNAL.yml
+++ b/.pipelines/1ES.binskim.BuildAndTest.EXTERNAL.yml
@@ -38,16 +38,56 @@ extends:
             name: $(poolName)
             os: $(poolOS)
           steps:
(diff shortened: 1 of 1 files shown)


Sasinkas (approved):
LGTM

PR #1188, opened 2026-05-20
 Fix portable PDB lifetime management to prevent AccessViolationException

Sasinkas:
IcM 798776046 — BinSkim crashes with an AV in rule BA2027 (EnableSourceLink) when scanning large codebases with --threads 10: System.AccessViolationException: Attempted to read or write protected memory.
   at System.Text.Ascii.WidenAsciiToUtf16(Byte*, Char*, UIntPtr)
   at Microsoft.CodeAnalysis.BinaryParsers.PortableExecutable.PE.ManagedPdbGetSourceLinkDocument(Pdb)
   
   RootCause: 
TryGetPortablePdbMetadataReader opens a MetadataReaderProvider (which owns the unmanaged memory backing MetadataReader, / BlobReader but never roots it past the method return. Under .NET 9's Dynamic PGO with aggressive GC, the provider is collected while BlobReader.ReadUTF8 still points at its memory. Additionally, the FileStream used to open the PDB is never disposed, leaking file handles for the process lifetime.
   
   Fix: 
The fix addresses a GC lifetime bug where TryGetPortablePdbMetadataReader opened a portable PDB's MetadataReaderProvider and FileStream as local variables that were never rooted on the PE instance—allowing the GC (especially under .NET 9 Dynamic PGO) to collect the provider while BlobReader still pointed into its memory, causing AccessViolationException, and simultaneously leaking the file handle. The solution promotes the provider, stream, and metadata reader to PE instance fields with lazy-init semantics (guarded by a portablePdbInitialized flag), so the PDB data remains valid for the entire analysis lifetime of the PE object, and PE.Dispose() deterministically releases both the provider and the stream.


src/Test.FunctionalTests.BinSkim.Rules/BA2027_PortablePdbLifetimeTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        /// </summary>
+        [Fact]
+        public void ManagedPdbGetSourceLinkDocument_ReleasesFileHandle_AfterDispose()
+        {
+            if (!PlatformSpecificHelpers.RunningOnWindows()) { return; }
+
+            string sourceDll = Path.Combine(TestDataDirectory, "CSharp_PortablePdb_SourceLink.dll");
+            string sourcePdb = Path.Combine(TestDataDirectory, "CSharp_PortablePdb_SourceLink.pdb");
+
+            if (!File.Exists(sourceDll) || !File.Exists(sourcePdb)) { return; }
```
martin-reznik: Silent `return` here turns a missing test fixture into a green test. The `.csproj` wildcard-deploys `FunctionalTestData\**\*` so absence indicates a broken deployment, not a known-skip scenario. `FluentAssertions` is already imported in this file — please use an assert:

```suggestion
            File.Exists(sourceDll).Should().BeTrue(
                because: "BA2027 PortablePdb test fixture must be deployed to FunctionalTestData.");
            File.Exists(sourcePdb).Should().BeTrue(
                because: "BA2027 PortablePdb test fixture must be deployed to FunctionalTestData.");
```

src/Test.FunctionalTests.BinSkim.Rules/BA2027_PortablePdbLifetimeTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        /// </summary>
+        [Fact]
+        public void ManagedPdbSourceFileChecksumAlgorithm_ReleasesFileHandle_AfterDispose()
+        {
+            if (!PlatformSpecificHelpers.RunningOnWindows()) { return; }
+
+            string sourceDll = Path.Combine(TestDataDirectory, "CSharp_PortablePdb_SourceLink.dll");
+            string sourcePdb = Path.Combine(TestDataDirectory, "CSharp_PortablePdb_SourceLink.pdb");
+
+            if (!File.Exists(sourceDll) || !File.Exists(sourcePdb)) { return; }
```
martin-reznik: Same as the comment on line 53 — replace the silent skip with an assert so a missing fixture surfaces clearly:

```suggestion
            File.Exists(sourceDll).Should().BeTrue(
                because: "BA2027 PortablePdb test fixture must be deployed to FunctionalTestData.");
            File.Exists(sourcePdb).Should().BeTrue(
                because: "BA2027 PortablePdb test fixture must be deployed to FunctionalTestData.");
```

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -886,42 +909,67 @@ public IEnumerable<string> GetAssemblyReferenceStrings()
...
+                    this.FileName,
+                    filePath =>
+                    {
+                        if (!File.Exists(filePath))
+                        {
+                            return null;
+                        }
+
+                        this.portablePdbStream = File.OpenRead(filePath);
+                        return this.portablePdbStream;
```
martin-reznik, line 939: This lambda writes to `this.portablePdbStream` as a side effect. PEReader may invoke it multiple times with candidate PDB paths — it disposes any returned stream it rejects, and takes ownership (via the out provider) of the accepted one. The fact that this doesn't leak relies on that PEReader behavior, and the fact that the eventual double-dispose on the accepted stream (once via the provider, once via the field in `PE.Dispose()`) is safe relies on `FileStream.Dispose` being idempotent. Worth making the contract explicit so a future maintainer doesn't "simplify" this into a leak:

```suggestion
                    // PEReader may invoke this callback multiple times with candidate paths.
                    // It disposes any returned stream it rejects, and takes ownership of the
                    // accepted stream via the out provider. We track the latest returned stream
                    // in a field so PE.Dispose() releases it deterministically; double-dispose
                    // is safe because FileStream.Dispose is idempotent.
                    filePath =>
                    {
                        if (!File.Exists(filePath))
                        {
                            return null;
                        }

                        this.portablePdbStream = File.OpenRead(filePath);
                        return this.portablePdbStream;
                    },
```

martin-reznik (changes requested):
Thanks for the fix — the lifetime model is correct and the IcM 798776046 root cause is well captured in code comments. Two small change requests before this lands; both are self-contained and won't change the fix itself.

**1. Silent `return` on missing test fixture (×2 occurrences).** The `.csproj` deploys `FunctionalTestData\**\*` via wildcard, so a missing DLL/PDB means a broken deployment. Silently passing the test in that case green-washes a real CI problem. Replace with an explicit assert so the failure mode is visible. Inline suggestions on the two lines below.

**2. Document the lambda's stream-ownership contract in `PE.cs`.** The lambda writes to `this.portablePdbStream` as a side effect, and the fact that this doesn't leak across multiple PEReader invocations relies on PEReader internally disposing rejected streams. Future maintainers will look at this and wonder. A short comment makes the contract explicit. Inline suggestion below.

**Non-blocking nit (description):** The PR body is just a truncated repeat of the title. Please add a paragraph covering: root cause (missing root + dispose, latent since #657 in 2022), the .NET 9 + Dynamic PGO trigger that surfaced the latent bug, the IcM reference, and what's covered by the new tests (file-handle leak) vs. what isn't (the original AVE — Dynamic PGO + GC race, impractical to reproduce deterministically). Helps anyone landing on this PR without IcM access.

Once these are in I'm happy to approve.

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
```diff
@@ -869,42 +892,72 @@ public IEnumerable<string> GetAssemblyReferenceStrings()
...
+        /// preventing the GC lifetime bug (IcM 798776046) and the FileStream leak.
+        /// </summary>
+        private MetadataReader GetPortablePdbMetadataReader(Pdb pdb)
         {
+            if (this.portablePdbInitialized)
+            {
+                return this.portablePdbMetadataReader;
+            }
+
+            this.portablePdbInitialized = true;
```
martin-reznik: Non-blocking, but worth a one-line fix: setting `portablePdbInitialized = true` before the I/O completes caches partial state if anything after this line throws.

`File.OpenRead` (L949), `MetadataReaderProvider.FromPortablePdbStream` (L950), and `GetMetadataReader()` (L959) can all throw — `BadImageFormatException` on a corrupt portable PDB is the realistic case. There's no `try`/`finally` here, and C# doesn't roll back field writes when an exception propagates out.

The SARIF SDK's `MultithreadedAnalyzeCommandBase.AnalyzeTargetHelper` wraps each `skimmer.Analyze(context)` call in a per-rule `catch (Exception ex)` and continues to the next skimmer using the same `context` (and therefore the same `target.PE`). So if BA2027 throws here on a corrupt PDB:

1. SDK logs the exception, moves to BA4001 / BA2004.
2. They call into the same `PE` instance, hit the early-return on L919-922 because `portablePdbInitialized == true`.
3. Get the still-null `portablePdbMetadataReader` and silently treat the binary as having no portable PDB available — wrong answer, no error surfaced.

The same shape also caches `null` if a future caller exercises the `pdb: null` + `TryOpenAssociatedPortablePdb` miss path (production callers don't today, but the new tests do).

Suggested fix: move `this.portablePdbInitialized = true;` to immediately after `this.portablePdbMetadataReader = this.portablePdbProvider.GetMetadataReader();` (L959) so the flag only commits on a clean init.

martin-reznik (approved):

martin-reznik (approved):

src/BinaryParsers/PEBinary/PortableExecutable/PE.cs
Sasinkas: fixed

PR #1190, opened 2026-05-27
Update building application logic + setting version

Sasinkas:
This PR consolidates and modernizes the BinSkim versioning strategy by removing the legacy SetCurrentVersion.cmd script and the manually-generated src/BinaryParsers/VersionConstants.cs file, eliminating the version-stamping shell logic from BuildAndTest.cmd, and instead declaring the version once in Directory.Build.props using standard MSBuild properties (AssemblyVersion, FileVersion, Version). It enables the SDK's GenerateAssemblyInfo (while suppressing a few unneeded attributes) in build.common.props, removes the now-redundant assembly-level version attributes from the Driver's AssemblyInfo.cs, and updates BuildPackages.cmd to query the version from MSBuild at pack time—making Directory.Build.props the single source of truth for the 4.4.9.9 version and allowing a parent BinSkimInternal repo to override it when this project is consumed as a submodule.


martin-reznik (changes requested):
Requesting changes — the in-repo refactor is clean, but the cleanup is incomplete.

Three sibling release/publish scripts in this repo still `call SetCurrentVersion.cmd` and will break after merge:
- `CreatePackagesFromLayoutDirectory.cmd:21`
- `DelistCurrentPackages.cmd:2` (also loses `%MAJOR_PREVIOUS%`/`%PATCH_PREVIOUS%` source)
- `PublishSignedPackages.cmd:2` (also loses the `-developer` safety guard)

CI green ≠ those release scripts work — they aren't invoked by any `.pipelines/*.yml` or `.github/workflows/*.yml`, so green CI does not prove they still work.

The paired internal PR (BinSkimInternal!922477) covers most of the BinSkimInternal-side restructure, but does NOT modify `BinSkimInternal/BuildPackage.cmd` (still calls `SetCurrentVersion.cmd`) and leaves the `Get-Content "SetCurrentVersion.cmd"` block in `IntegrationTests.yml` unchanged (gated by `publishAndTestPackage=false`, so latent). Flagging those on the internal PR — they shouldn't block this one, but need to land before the submodule pointer is bumped.

Suggested resolution: either update the three scripts above to use `dotnet msbuild --getProperty:Version` like `BuildPackages.cmd` now does, or keep a tiny `SetCurrentVersion.cmd` stub that emits the same `set MAJOR=...` lines derived from `Directory.Build.props`. The stub option keeps every existing caller working without touching them.

Per-file details below — leaving them in the body because the three affected scripts aren't part of this PR's diff and don't accept inline review comments.

---

### `CreatePackagesFromLayoutDirectory.cmd:21`

`call SetCurrentVersion.cmd` will fail after merge — the file is deleted in this PR. Once the call fails, `%MAJOR%`/`%MINOR%`/`%PATCH%`/`%PRERELEASE%` stay unset and the script will pack with an empty/garbage `Version`.

Easiest fix: do the same `dotnet msbuild --getProperty:Version` trick that `BuildPackages.cmd` now uses, and parse out `MAJOR`/`MINOR`/`PATCH` from the result. Or restore a `SetCurrentVersion.cmd` stub.

### `DelistCurrentPackages.cmd:2`

Two problems here after merge:

1. `SetCurrentVersion.cmd` is deleted, so `%MAJOR_PREVIOUS%`/`%MINOR_PREVIOUS%`/`%PATCH_PREVIOUS%`/`%PRERELEASE_PREVIOUS%` won't get set.
2. Even with the script restored, `Directory.Build.props` has no concept of "previous" version — that data only lived in `SetCurrentVersion.cmd`.

You either need a separate previous-version source (env var? script arg? `PreviousVersion.props`?), or you need to keep the old `SetCurrentVersion.cmd` alive specifically for delist.

### `PublishSignedPackages.cmd:2`

Same `SetCurrentVersion.cmd` dependency as the other two release scripts — but worth flagging extra: this script's `if "%PRERELEASE%" EQU "-developer"` early-return is the safety guard that prevents accidentally publishing developer bits. After this PR `%PRERELEASE%` will be empty, the check will pass, and the script will happily try to publish whatever's in the layout dir. Please don't lose that guard while wiring up the new version source.

martin-reznik (approved):
Thanks Sara. Confirmed offline:
- The release pipelines (392, 150) that consume the residual `CreatePackagesFromLayoutDirectory.cmd` -> `SetCurrentVersion.cmd` path are manually triggered, so there's no automated-build risk.
- No tracking work item; the leftover dependency will be cleaned up the next time a release is cut.

Dropping my CHANGES_REQUESTED -- approving.

martin-reznik (approved):
Re-approving: only change since my prior approval is the additive 'Add unit tests for MultithreadedAnalyzeCommand rule specifier parsing' commit. No regressions to the previously resolved concerns. LGTM.

PR #1199, opened 2026-07-02
fix: populate fileHash in CompilerDataLogger telemetry

martin-reznik:
## Summary

The hash field in CompilerDataLogger telemetry (AppInsights CompilerInformation events + CSV) has always been empty because context.Hashes is not populated by the SARIF framework during rule execution.

## Root Cause

CompilerDataLogger.Write() reads context.Hashes?.Sha256, but the SARIF SDK only computes hashes for SARIF artifact entries after analysis - not before dispatching rules.

## Fix

Compute the hash directly in the rules via new CompilerData.FileHash property:
- BA4001 (PE): uses PE.SHA256Hash (zero extra I/O)
- BA4002 (ELF/Mach-O): SHA256.HashData(File.ReadAllBytes(...))

## Tests

Two tests verify hash is populated in CSV output with and without DataToInsert=Hashes.

### DIFF
--- a/.pipelines/1ES.binskim.BuildAndTest.EXTERNAL.yml
+++ b/.pipelines/1ES.binskim.BuildAndTest.EXTERNAL.yml
@@ -17,6 +17,7 @@ extends:
       os: windows
     featureFlags:
       autoEnableRoslynWithNewRuleset: false
+      sdlSdpEnabled: false
     stages:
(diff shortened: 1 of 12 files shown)


martin-reznik:
Closing - hash fix will be handled in Guardian instead. Pipeline fixes moved to #1201.

PR #1207, opened 2026-07-31
Handle DWARF string sections larger than 2 GB

v-davidquiceno-ascendion:
## Summary

- Add a string-reader abstraction for DWARF string sections.
- Add the opt-in `--dwarf-string-read-threshold` CLI option.
- Use file-backed reads only when that option is supplied and the DWARF string section reaches the configured limit.
- Preserve the existing in-memory behavior by default and for sections below the configured limit.
- Add regression tests for reads beyond 2 GB and for the opt-in threshold behavior.

## Root cause

ELFSharp materializes section contents in a byte array. Sections larger than `int.MaxValue` overflow when their size is converted to a 32-bit integer, preventing BinSkim from analyzing otherwise valid ELF binaries.

## Validation

- All 154 `Test.UnitTests.BinaryParsers` tests passed.
- All 85 `Test.UnitTests.BinSkim.Driver` tests passed.
- The new CLI option is present in `BinSkim analyze --help`.
- The earlier custom-package validation successfully analyzed the affected Minecraft Switch artifact.

### DIFF
--- a/src/BinSkim.Driver/AnalyzeOptions.cs
+++ b/src/BinSkim.Driver/AnalyzeOptions.cs
@@ -49,7 +49,7 @@ namespace Microsoft.CodeAnalysis.IL
         public string LocalSymbolDirectories { get; internal set; }
 
         [Option(
-            "dwarf-string-section-file-read-threshold-bytes",
+            "dwarf-string-read-threshold",
             HelpText = "Use file-backed reads for DWARF string sections whose size is at least this many bytes. " +
                        "By default, DWARF string sections are loaded into memory.")]
         public ulong? DwarfStringSectionFileReadThreshold { get; set; }
--- a/src/BinSkim.Driver/AnalyzeOptions.cs
+++ b/src/BinSkim.Driver/AnalyzeOptions.cs
@@ -49,6 +49,12 @@ namespace Microsoft.CodeAnalysis.IL
         public string LocalSymbolDirectories { get; internal set; }
 
         [Option(
+            "dwarf-string-section-file-read-threshold-bytes",
+            HelpText = "Use file-backed reads for DWARF string sections whose size is at least this many bytes. " +
+                       "By default, DWARF string sections are loaded into memory.")]
+        public ulong? DwarfStringSectionFileReadThreshold { get; set; }
+
+        [Option(
             "ignorePdbLoadError",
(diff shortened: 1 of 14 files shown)


jamoor-moj:

v-davidquiceno-ascendion:

martin-reznik (approved):
Thanks @v-davidquiceno-ascendion, and sorry for the slow turnaround. I reviewed this carefully and it's in good shape — **approving.**

**What I verified**
- **The behavior is genuinely opt-in, as requested.** `ShouldUseFileBackedDwarfStringReader` returns true only when `--dwarf-string-read-threshold` is supplied *and* the `.debug_str` section reaches it. With no argument the value is `null`, so every existing user keeps the exact in-memory path. Nice touch adding `..._DwarfStringFileReadIsOptIn` to assert the default stays null.
- **The 2 GB root cause is actually bypassed, not just guarded.** In the file-backed branch the large `byte[]` (`DebugDataStrings`) is never materialized — you seek by `section.Offset`/`section.Size` directly — so the ELFSharp `int` overflow can't occur.
- **No thread-safety regression.** Per-binary DWARF parsing is sequential (`ParseAllCompilationUnits` / `ParseLineNumberPrograms` are plain `while` loops; BinSkim parallelizes at the file level), and each `ElfBinary` builds and disposes its own reader per `Lazy`, so the stateful file-backed reader is never shared across threads. Hoisting the reader from per-compilation-unit to per-binary is behavior-preserving because `DwarfMemoryReader.ReadString(uint)` is positional (saves/restores `Position`).
- **Mach-O is untouched functionally** — only signature adaptation to `IDwarfStringReader`, no threshold wired in. Correct for the ELF scenario here.

**Non-blocking notes (fine to defer)**
1. `ReadString(uint position)` bounds the addressable offset at 4 GB, so this covers the 2–4 GB DWARF32 case but not a >4 GB (DWARF64) `.debug_str`. Reasonable to leave for later.
2. The opt-in read is byte-by-byte (`Seek` + `ReadByte` into a `List<byte>` per string), which is slow on very large sections. Acceptable as an escape hatch; could be buffered later if it matters.
3. Mach-O still can't handle a >2 GB `.debug_str`. Worth noting as a known limitation so it isn't a surprise for a future caller.

I'm going to retarget this onto a `fix/large-debug-str` integration branch so we can land a couple of related BinSkim changes together and promote to `main` as one step — nothing you need to do, the diff is unchanged.

PR #1214, opened 2026-08-24
Fix BA2022 verification for long paths

### LINKED ISSUE #977: Unclear Error message when the path of the file to long
User report that in Windows when the path is too long, Binskim report the error is unclear:

```
Analyzing... 
D:\a\_work\verylong path\aaa.dll: error BA2022: 'aaa.dll' signing was flagged as insecure by WinTrustVerify with error code 'CRYPT_E_FILE_ERROR' (An error occurred while reading or writing to a file.) 
```
MAX_PATH, which is defined as 260 characters.

This issue is created to investigate if improvements can be made.

akoeplinger:
Enable long-path awareness for BinSkim and its test host, and add a regression test for signed binaries beyond MAX_PATH.

Fixes #977

### DIFF
--- a/src/Directory.Packages.props
+++ b/src/Directory.Packages.props
@@ -9,7 +9,7 @@
     <PackageVersion Include="ELFSharp" Version="2.17.3" />
     <PackageVersion Include="FluentAssertions" Version="7.0.0" />
     <PackageVersion Include="Microsoft.ApplicationInsights" Version="2.22.0" />
-    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.12.0" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.14.1" />
     <PackageVersion Include="Moq" Version="4.20.72" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="Sarif.Driver" Version="4.6.0" />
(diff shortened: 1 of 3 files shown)


src/Test.FunctionalTests.BinSkim.Rules/RuleTests.cs
```diff
@@ -94,17 +96,23 @@ private void Verify(
...
+                        continue;
+                    }
 
-                if (skimmer.CanAnalyze(context, out string reasonForNotAnalyzing) != AnalysisApplicability.ApplicableToSpecifiedTarget)
+                    skimmer.Analyze(context);
+                }
+                finally
                 {
-                    continue;
+                    context.Dispose();
```
akoeplinger, line 114: this was needed since otherwise the context was leaking and keeping file handles open

src/Directory.Packages.props
```diff
@@ -9,7 +9,7 @@
     <PackageVersion Include="ELFSharp" Version="2.17.3" />
     <PackageVersion Include="FluentAssertions" Version="7.0.0" />
     <PackageVersion Include="Microsoft.ApplicationInsights" Version="2.22.0" />
-    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.12.0" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.14.1" />
```
akoeplinger, line 12: this was needed since the testhost binary itself was made longPathAware in that version

